PR #2, opened 2018-08-13
Logging in millisecs

igormcoelho:
After change to Akka consensus in NEO 3.0, it's so fast that seconds do not mean much anymore.. now milliseconds is important to actually know of what is going on the consensus algorithm.


igormcoelho:
reference to millisecs need : https://github.com/neo-project/neo/pull/288

PR #3, opened 2018-08-24
remove break statement from FilterFree

hal0x2328:
In `FilterFree`, once the free transaction limit has been reached, `yield break` is called, preventing further iteration on the the list of transactions. Hence, blocks become limited to contain only the number of transactions that were returned before that point, instead of iterating all the way to `MaxTransactionsPerBlock`. This change removes that statement so that any additional transactions with fees attached may be processed as the `foreach` statement continues.

### DIFF
--- a/SimplePolicy/SimplePolicyPlugin.cs
+++ b/SimplePolicy/SimplePolicyPlugin.cs
@@ -44,8 +44,6 @@ namespace Neo.Plugins
                     yield return tx;
                 else if (count++ < Settings.Default.MaxFreeTransactionsPerBlock)
                     yield return tx;
-                else
-                    yield break;
         }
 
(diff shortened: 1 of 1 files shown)


vncoelho:
Eagle eyes! I am happy to see such improvements today, really.

Edgegasm:
@vncoelho Now we just need wakka akka 😄

vncoelho:
Hey,  @Edgegasm.
I am happy about how everything happened today.

Different contributions and @erikzhang, as always, being very careful, open and fast when it is needed.
The Akka refactoring is also something impressive. It will be great when it goes live.
I am sure that learning those concepts took @erikzhang months, plus his experience and expertise.

PR #4, opened 2018-08-24
Fix Priority Fee Transactions being treated as Free

f27d:
Currently there is an issue with Integer rounding making **small fee tx behave like free tx.**

Believe this is the cause of priority fee not getting observed.

Simply removing p.Size in the ranking constraint will resolve this. Discussed with jseagrave21.
Based on Belane's investigation.

@erikzhang


erikzhang:
Transaction fees should be based on the size of the transaction. When sorting, we have to consider the size of the transaction and the amount of the fee.

f27d:
@Belane and @Jseagrave21 improved on the previous iteration @erikzhang.

vncoelho:
Eagle eyes! I am happy to see such improvements today, really.

shargon:
Nice work!

PR #5, opened 2018-08-31
OnImportMaxHeight Plugin in the scope of simple policy

vncoelho:
This would be the impact, @erikzhang.
Everyone would need to update their `config` file because of the parameters load.
What do you think, master?

Should we move it to that isolated plugin, such as that `ILoadingPlugin/IOnImportPlugin`?


vncoelho:
Perfect, @erikzhang.
Great generalization you have done. Thanks.

PR #6, opened 2018-08-31
Initial README with TODO

vncoelho:
@erikzhang, there is still a missing explanation in the guidelines, like the use of the object `Policies`.
I did not get how many objects are expected inside these objects https://github.com/neo-project/neo/blob/6f9208d2e96dda7dae3b5bca2e9ecccc69c03531/neo/Plugins/Plugin.cs#L10-L12

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,22 @@
+# What is it
+
+A set of plugins that can be used inside the [Neo core library](https://github.com/neo-project/neo).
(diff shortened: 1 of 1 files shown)


vncoelho:
Makes sense, @erikzhang. :dagger: 
The plugin is almost like editing the core library itself.

PR #7, opened 2018-09-01
add ImportBlocks

erikzhang:
@vncoelho You can add `MaxImportHeight` logic to this plugin.

After that, we can remove the import block function from neo-cli and neo-gui.

### DIFF
--- a/ImportBlocks/ImportBlocks.cs
+++ b/ImportBlocks/ImportBlocks.cs
@@ -1,0 +1,83 @@
+﻿using Akka.Actor;
+using Neo.IO;
+using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
+using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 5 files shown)


vncoelho:
Nice idea, @erikzhang. Great insight. It will proportionate more use cases and better portability.
I learned several small things on this plugin journey.

How should be done? After being merged we edit on top?
I think that it would be:
* the `settings.cs` file with:  
    * `public uint MaxOnImportHeight { get; }`
    * and `this.MaxOnImportHeight = (uint)GetValueOrDefault(section.GetSection("MaxOnImportHeight"), 0, p => int.Parse(p));`
    * The function below   inside the `ImportBlocks.cs` and, probably, being called at `GetBlocks`:
```cs
         public bool CheckMaxOnImportHeight(uint currentImportBlockIndex)
        {
	    if(Settings.Default.MaxOnImportHeight != 0 && Settings.Default.MaxOnImportHeight == currentImportBlockIndex)
		return true;
	    return false;
        }
```

* an additional `config.json` file with: "MaxOnImportHeight": 0

PR #8, opened 2018-09-02
Trying to include onmaxheight import functionality

vncoelho:


vncoelho:
Hi, Erik,
It took my some time to understand this "reverse" logic...aehuaheauea
Thanks for the fix,

vncoelho:
Maybe an `else{ break;}`, in order to avoid the rest of the loop in this `if` https://github.com/neo-project/neo-plugins/pull/8/files#diff-36010cba45ac43855397981d68cd892dR75, or the compiler can do its job?

erikzhang:
You are right. Changed: https://github.com/neo-project/neo-plugins/commit/bd3db2ccd56706cb8496e0afdacd9bd002d72476

erikzhang:
The code for importing blocks is in the constructor of the plugin, so it will run when the plugin is loaded and does not need to be called again.

PR #17, opened 2018-10-20
Including number of dumped states

vncoelho:


StatesDumper/StatesDumper.cs
```diff
@@ -40,7 +40,7 @@ protected override bool OnMessage(object message)
                 return state;
             }));
             File.WriteAllText(path, array.ToString());
-            Console.WriteLine($"States have been dumped into file {path}");
+            Console.WriteLine($"States ({states.Count().ToString()}) have been dumped into file {path}");
```
erikzhang: `states.Count()` causes `states` to be re-enumerated. I recommend using `array.Count`:

```suggestion
            Console.WriteLine($"States ({array.Count}) have been dumped into file {path}");
```

PR #19, opened 2018-10-24
Blockstorage command - DumpInBlock

vncoelho:


vncoelho:
Complex implementation, off course..

diskooooo:
> 50 commits. Complex implementation, off course..

You do know you can squash commits and force push a WIP branch/PR, right? 😆

erikzhang:
If a data is written in block 100, and is modified in block 200. If you export the states of height 150, the data will be missed.

vncoelho:
Yes, @erik, exactly this.

We use this for dumping the state of every block when re-syncing the chain, at Persist.

Then, we can create files only with the states transitions.

https://github.com/NeoResearch/neo-storage-audit

erikzhang:
> We use this for dumping the state of every block when re-syncing the chain, at Persist.

If so, we do not need `StorageItem.Height`, right?

StatesDumper/StatesDumper.cs
```diff
@@ -42,5 +48,34 @@ protected override bool OnMessage(object message)
...
+            JArray array = new JArray(states.Where(p => 
+            {
+                byte[] src = p.Value.ToArray();
+                StorageItem si = new StorageItem();
+                using (MemoryStream stream = new MemoryStream(src)) {
+                  using (BinaryReader reader = new BinaryReader(stream)) {
+                    si.Deserialize(reader);
+                  }
+                }
+                uint h = si.Height;
```
vncoelho: We need it here, Erik.

vncoelho:
si.Height after serialization.

vncoelho:
Thanks, @erikzhang, great idea and will probably work much faster than the current Filter.
It will probably solve our goal of exporting the storage quick and precise!

The only tradeoff will be the RPC call `getstorageheight` (PR https://github.com/neo-project/neo/pull/424), which is a tool that is good for transparency, exchanges would probably like.
However, we could create a parallel service for that.

vncoelho:
A draft of the dump is being done here, @erikzhang: https://github.com/vncoelho/neo/commithttps://github.com/vncoelho/neo/commit/e4c86ba486e29d93a6e0f1f4e625ed0fbf967430

PR #21, opened 2018-10-27
Persitence Plugin for exporting storage changes

vncoelho:
First draft for Persistence plugin.

- [X] First draft
- [x] Fix compiling errors
- [x] Create a bool variable that says the task to be done on Persist, in our cases, storage dump on every block.
- [x] Sync mainnet
- [x] Tested


vncoelho:
@erikzhang, @igormcoelho, we are not being able to get the object class `Trackable`.

```cs
Persistence.cs(24,22): error CS0246: The type or namespace name 'Trackable' could not be found (are you missing a using directive or an assembly reference?) [/opt/neo-plugins/Persistence/Persistence.csproj]
```
Can you give a hand?

erikzhang:
By the way, is it really necessary to put this plugin into neo-project repos?

vncoelho:
image

Almost....aehauheauheaea

igormcoelho:
Erik, if you think this is useful as example we can keep it, otherwise we just use it in our storage verification project.

vncoelho:
```cs
[{"state":"Added","key":"System.Byte[]","value":"System.Byte[]"},{"state":"Added","key":"System.Byte[]","value":"System.Byte[]"},{"state":"Added","key":"System.Byte[]","value":"System.Byte[]"},{"state":"Added","key":"System.Byte[]","value":"System.Byte[]"},{"state":"Added","key":"System.Byte[]","value":"System.Byte[]"}]},
```
Again... =/

vncoelho:
New export
```js
{"block":1444843,"size":5,"storage":
[{"state":"Added","key":"85104d0b1bc285289b17717a6facaa2cbd1712b3","value":"0060b7c6c5fe11"},{"state":"Added","key":"130b891dc5341bcef93c077fc7ec5624ee8776f8","value":"00f87f24795120"},{"state":"Added","key":"45bcb590e3e0fb0010d7bfde6f7cd39382fd86e9","value":"00c0b6403b6f0c"},{"state":"Added","key":"746f74616c537570706c79","value":"0018ee2b7abf3e"},{"state":"Added","key":"73616c65734e656f","value":"0041ad29410b"}]},
```
The previous one:
```js
{"block":1444843,"size":5,"storage":
[{"key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec130b891dc5341bcef93c077fc7ec562400ee8776f80000000000000000000000000c","value":"000700f87f2479512000eb0b1600"},{"key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec45bcb590e3e0fb0010d7bfde6f7cd3930082fd86e90000000000000000000000000c","value":"000700c0b6403b6f0c00eb0b1600"},{"key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec73616c65734e656f000000000000000008","value":"00060041ad29410b00eb0b1600"},{"key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec746f74616c537570706c79000000000005","value":"00070018ee2b7abf3e00eb0b1600"},{"key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec85104d0b1bc285289b17717a6facaa2c00bd1712b30000000000000000000000000c","value":"00070060b7c6c5fe1100eb0b1600"}]},
```

I think it is still wrong. Values not match

vncoelho:
Done:  :+1: 

```js
{"block":1444843,"size":5,"storage":
[{"state":"Added","key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec85104d0b1bc285289b17717a6facaa2c00bd1712b30000000000000000000000000c","value":"00070060b7c6c5fe1100"},{"state":"Added","key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec130b891dc5341bcef93c077fc7ec562400ee8776f80000000000000000000000000c","value":"000700f87f2479512000"},{"state":"Added","key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec45bcb590e3e0fb0010d7bfde6f7cd3930082fd86e90000000000000000000000000c","value":"000700c0b6403b6f0c00"},{"state":"Added","key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec746f74616c537570706c79000000000005","value":"00070018ee2b7abf3e00"},{"state":"Added","key":"f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec73616c65734e656f000000000000000008","value":"00060041ad29410b00"}]},
```

Thanks @erikzhang, without your assistance and attention it would be impossible to reach these magic functions!

PS: @igormcoelho, there is some minnor different between the storage reported here and the previous one we were getting before.
Even the Key is different.
The values is different (as expected) because we added that new field of Height. But I do know, what do you think?

igormcoelho:
Congratulations brother @vncoelho for your Persistance xD The values seem correct, it's just the order that is not the same. 
https://raw.githubusercontent.com/NeoResearch/neo-storage-audit/master/BlockStorage_1500000/dump-block-1445000.json
But order is not important for modified PatriciaMerkle tree, so we can use this valuable information to finally build that tree ;)

Persistence/Persistence.cs
```diff
@@ -31,47 +31,50 @@ public void OnPersist(Snapshot snapshot)
...
+                Settings.Default.BlockStorageCache = Settings.Default.BlockStorageCache + "{\"block\":" + blockIndex.ToString() + ",\"size\":" + array.Count.ToString() + ",\"storage\":\n";
+                Settings.Default.BlockStorageCache = Settings.Default.BlockStorageCache + array.ToString() + "},\n";
 
-            if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (blockIndex > Settings.Default.HeightToStartRealTimeSyncing))
-            {
-                Settings.Default.BlockStorageCache += "]";
-                File.WriteAllText(path, Settings.Default.BlockStorageCache);
-                Settings.Default.BlockStorageCache = "[";
-            }   
+                if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (blockIndex > Settings.Default.HeightToStartRealTimeSyncing))
```
igormcoelho: Good idea brother... a config variable.

vncoelho: @igormcoelho, maybe put all these parameters in an array is a better idea, yep? What do you think?

Because there is a switch case at OnPersist in the way that other information can be extract if someone thing about something.

erikzhang:
I think we can merge this plugin with [StatesDumper](https://github.com/neo-project/neo-plugins/tree/master/StatesDumper).

vncoelho:
Storage dump on top, @erikzhang. 

image


We gonna migrate to StatesDumper and test again.

vncoelho:
On top and syncing the blockchain. 
image

All tests are completed.
Soon we gonna move to StateDumper and separate parameters.

vncoelho:
https://github.com/NeoResearch/neo-storage-audit

StatesDumper/StatesDumper.cs
```diff
@@ -63,5 +65,77 @@ private bool OnHelp(string[] args)
...
+                            state["key"] = trackable.Key.ToArray().ToHexString();
+                            break;
+                    }
+                    array.Add(state);
+                }
+
+                Settings.Default.BlockStorageCache = Settings.Default.BlockStorageCache + "{\"block\":" + blockIndex.ToString() + ",\"size\":" + array.Count.ToString() + ",\"storage\":\n";
+                Settings.Default.BlockStorageCache = Settings.Default.BlockStorageCache + array.ToString() + "},\n";
+
+                if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (blockIndex > Settings.Default.HeightToStartRealTimeSyncing))
```
erikzhang: Does this prevent data loss?

erikzhang: I think the only way to prevent data loss is to save the data every time the block is persisted.

vncoelho: It does not prevent, @erikzhang , you are right.

What we did is the option of choosing the `BlockCacheSize`.
If 'BlockCacheSize==1` there will be no data loss.

In our case, we chose it `BlockCacheSize==1000` (it is a little faster instead of persisting every block when syncing for the first time) and dumped with `ImportPlugin chain.acc`.

The chance of data loss is lower when Importing.
That is why there is an additional variable that is `HeightToStartRealTimeSyncing` which, implicitly, forces BlockCacheSize to 1.

There could be a tool that prints it every block and later deletes the files and merge then.
We currently do it in `bash` scripts: https://github.com/NeoResearch/neo-tests/blob/master/simple-rpc-node/merge_Storage.sh

erikzhang: What's the difference between `BlockCacheSize=1` and `BlockCacheSize=1000`? Is it related to performance?

vncoelho: The performance was not good when persisting file on every block during import, then we created this possible cache.
What do you think, @igormcoelho?

igormcoelho: Yeah, reducing disk writes is fundamental for performance... we experimented with BlockCacheSize=1 and it was very slow during block import... the bigger the Cache the better. We found that 1000 was a good compromise between risking losing block loading and offloading it to disk at every few seconds (during block import only, after that behavior should go to 1 again, which is enough for every 15 seconds).

vncoelho:
@erikzhang, you organized everything. Thanks!
Sorry for the previous design.

It looks all correct, if there is any error it would be minor.
When possible we gonna set a node to sync the whole Mainnet again.

erikzhang:
If you have no problem testing, you can merge it by yourself.

vncoelho:
Thank you for the opportunity and trust.

PR #24, opened 2018-11-03
Basic unit test structure for SimplePolicyPlugin

igormcoelho:
Guys, I did some progress in creating basic unit testing for SimplePolicyPlugin. 
Using this, we will be able to easily test if transaction policies are being respected, even on compile time. I tried to not change anything on the original classes, but I couldn't... sorry again for that. I hope my changes are not too big.
With this proposed change, it's possible to put configuration files on other directories... this was a problem in my system here, but perhaps we can do that in another way, without changing original class.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,39 @@
+## What is it
+A set of plugins that can be used inside the Neo core library. Check [here](http://docs.neo.org/en-us/node/plugin.html) for the official documentation.
+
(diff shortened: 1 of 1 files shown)


igormcoelho:
I don't know what happened to my branch here... I'll start another clean one: https://github.com/neo-project/neo-plugins/pull/25

PR #25, opened 2018-11-03
Basic unit test structure for SimplePolicyPlugin (2)

igormcoelho:
Guys, I did some progress in creating basic unit testing for SimplePolicyPlugin. Using this, we will be able to easily test if transaction policies are being respected, even on compile time.
I created a simple configuration example with maximum 10 transactions and 3 free.


erikzhang:
Waiting for #10

PR #32, opened 2018-12-06
SmartContract plugin for neo-cli

txhsl:
First, this is just a test version for SmartContract plugin, so there are many conflicts. And the `neo` dependency is from source code instead of nuget.

All related projects are forked here [neo-system](https://github.com/neo-system) and organized here [txhsl/neo-system](https://github.com/txhsl/neo-system). A new plugin interface is added in [neo/Plugins](https://github.com/neo-system/neo/blob/master/neo/Plugins/ISmartContractPlugin.cs), an init call is added in [neo-cli/Shell](https://github.com/neo-system/neo-cli/blob/master/neo-cli/Shell/MainService.cs) and a new plugin is added here.

This plugin, provides `test invoke <hash>`, `invoke <hash>`, `compile <path>`, `test deploy <path>` and `deploy <path>`, 5 new command for neo-cli to add the smart contract function in the format of plugin.

The source code of neon 2.9.2 is attached in the project so it looks huge, but the main part is in the SmartContract.cs.

compile:
image

test deploy:
image

deploy:
image

test invoke:
image

invoke:
image

It's a sudden idea which came out when I was free, why not allow smart contract development in neo-cli?


erikzhang:
I don't think we need to add `ISmartContractPlugin` to neo. You can use `OnMessage` method for receiving messages from cli.

shargon:
If we don't reuse the logic from the original repository, we can lose a lot of fixes

erikzhang:
There are two options for reuse neo-compiler code:
1. Use `git submodule`
2. Use nuget

shargon:
@erikzhang Both of them are fine for me, but i need at least one to approve this PR

txhsl:
@shargon I've tried `git submodule` and get a problem. The project neo-compiler has 3 sub folders, and they are neoa, neoj and neon. They can be built separately but there will be naming conflicts if I use them as lib codes, e.g. `NeoModule`. So I change to the nuget solution, which is just temp, as I wrote in the description.

shargon:
@erikzhang please, make a official nuget for `Neo.Compiler`. Maybe should be easier than submodules

erikzhang:
@txhsl You can transfer the ownership of the package to dev@neo.org.

txhsl:
@erikzhang Made it, plz remove me later

erikzhang:
Done. Now all neo packages are under [neo-project](https://www.nuget.org/profiles/neo-project).

vncoelho:
@txhsl, now it was cleaned I could check.
This is a great plugin, congratulations. It has plenty of potential good applications!
In particular, for Unit Testing.

1. Could you merge `Deploy` with `TestDeploy` and `Invoke` with `TestInvoke`? From my perspective, they have quite similar codes and that would be great for future maintenance and simplicity. You can use a flag `testMode` as parameter of them.

1. Unit testing (This can be in another PR if you want).
    * Since are mastering this NuGets maybe you could add a Nuget for the [Examples](https://github.com/neo-project/examples-csharp).
    * In this sense, you could cache the example file and load it as Unit Test and call `OnCompiler`. That is a thing with several potential good applications!
   * Consequently, this will test `neo-compiler`, `neo-devpack`. Furthermore, it will open doors for several other UT that can check even `Neo-Vm` expected results with `Invoke` and `Deploy`
   * That could be also possible to cross-add this UT_SmartContract in that aforementioned projects!

txhsl:
@vncoelho ,nice suggestion!

erikzhang:
I have a little bit worried about the plugin getting the wallet. Maybe we should think of a security mechanism?

erikzhang:
Before we figure out the security mechanism, maybe we should put the contract deployment function into the neo-cli?

reference: https://github.com/neo-project/neo-cli/pull/180

vncoelho:
@txhsl, what about the Unit Tests? :dagger: :dancer: 
Would you like to do it for this PR or in another one? Did you get the idea?

I think that Compilation UT (using neo-projects/csharpexamples) have great potential for the whole NEO projects, as I highlighted above.

jsolman:
I may revive https://github.com/neo-project/neo-cli/pull/180 soon. It was made before the Akka rewrite, and needs to be adjusted now for minimum fees based on transaction size. Also, I have one change on it that allows a param for setting the contract payable flag.

txhsl:
@vncoelho That would be in another PR, and my idea is using docker and expect, based on the UTs in neo-cli, for 
 now. I was working on other issues recently, I will manage to start as soon as possible.

jsolman:
Updated neo-project/neo-cli#180 by creating neo-project/neo-cli#273

txhsl:
@vncoelho Some UTs are added based on [neo-project/examples-csharp](https://github.com/neo-project/examples-csharp). I can just compile from the `.dll` files instead of '.cs' files, so I built these examples first.
Here is a screenshot: 
image
They work well on Windows 7, I've tested cloning the repository, and building from the source without warning on VS2017 Pro. But there seems to be some problems opening this UT project on Windows 10.
I may fix it later. :sun_with_face:

txhsl:
Add a new command `debug <path>` for developers to test some basic functions of their contract based on a local avm file.
image

txhsl:
After having the `debug <path>` function, I notice that there are much more tests can be done based on [neo-project/examples-csharp](https://github.com/neo-project/examples-csharp).
But I think the plugin is already my v1.0 now. I need some review. @erikzhang @shargon @vncoelho

vncoelho:
Hi @txhsl,

Congratulations for the great job!
You already have my approval on this.

One minor thing are these `.dll`, I did not like this because we would need to update the files here. I think it is not a good solution.
I think that we should try to get the raw `.cs` because you are using the .dll and converting it to string `(new string[] { "compile", "HelloWorld.dll", "--compatible" }`
What do you recommend?

Wish you all the best, a merry Christmas and happy new year.

txhsl:
Merry Christmas! I will have a try about that.

igormcoelho:
This is nice, Merry Christmas everyone!

vncoelho:
Hi @txhsl,

let's go for this PR, it is a quite important feature :dagger: that is a great step for Unit Testing.

vncoelho:
I also do not like, @shargon,

We could use nuget of git submodule for getting the source and putting it into a variable that is later compiled.
This can be the base of an Unit Tests for both the `C#_Examples`, `neo`, `neo-cli` and `neo-vm`.

We can use a set of compile, deploy and invokes that should be always the same.
It gives us some extra robustness toward modification.

What do you think?

shargon:
Submodules is fine for me

vncoelho:
@txhsl, do you think you can roll it back to use `.cs` files instead of `.dll`, which were the limitations?

txhsl:
> @txhsl, do you think you can roll it back to use `.cs` files instead of `.dll`, which were the limitations?

Well, I still insist that using `.cs` files will result in different hashes on different platforms. And I'm working on other business in the school.
I'll roll it back at least.

igormcoelho:
@txhsl the workflow for C# is: 
Step 1. Compile C# to MSIL (Microsoft Common Language)
Step 2. Convert MSIL to AVM (each AVM is represented by a 20-byte ScriptHash)
Step 3. Optimize AVM (currently not done automatically by neon compiler)

Step 1 may vary from platform (linux/windows), dotnet version, sdk version, so a slight variation will also impact Step 2 (and generate a different AVM). But in my opinion, if  we have a strong Step 3 (optimizer), which is something we've been working on github.com/neoresearch/neo-avm-optimizer project, in many scenarios we are now able to generate the same optimized ScriptHash, even when the Step 1 AVM is different.

But there's no way to guarantee that the same C# code will ever generate the same scripthash, unless you also record: what compiler you used? in which platform? which dependencies?

vncoelho:
It is kind of a hard job, but maybe we could create those if cases for key plataforms that the Unit Tests might be running.
At least for Travis CI we might have no problem. The main issue will be for running the tests locally.
These tests are promising for the consistency of our different reps.

shargon:
@igormcoelho we should join step 3 and 2 with `--optimize` or something like that

vncoelho:
Considering commit https://github.com/neo-project/neo-cli/commit/b5cb60e1916b692317fcc02c4bfc9725ab93418e, what should be kept in this PR?

jsolman:
Some parts of this should be useful. https://github.com/neo-project/neo-cli/commit/b5cb60e1916b692317fcc02c4bfc9725ab93418e Added accepting quoted options on the command line which is nice to be able to not prompt for each parameter; so a deploy command can just be copied and pasted. Some refactoring is needed on `neo-cli` still to make the code more modular and maintainable.

The code for deploy support could be moved back into a plugin like this, but it is probably fine for it to be in the main code for neo-cli, while the compile support I think is best to stay in a plugin like this. My other commit still needed better support for all the invoke parameter types and would probably be nice to have test ops as this introduced.

I’m working on a plugin PR at the moment, but when I’m done with it I had expected to improve the invoke support. However, @txhsl if you want to make some of the changes for it, let me know.

shargon:
https://github.com/neo-project/neo-cli/commit/b5cb60e1916b692317fcc02c4bfc9725ab93418e i prefer to move the logic to neo-cli

vncoelho:
Yes, I agree.

The compiling part could be here, yes? Or move everything?

If you prefer everything I also agree. This tool does not have negative impacts on the client performance, just more features.

vncoelho:
So, let's close this for now, @shargon, right?

The deploy and invoke are already implemented now on the neo-cli, correct?
Thus, what would need to be migrated is the Compiling part.

@txhsl, fell free to reopen and thanks for the insights and contributions.

PR #37, opened 2019-01-08
Expose the max TX per block and max low priority TX per block.

jsolman:
The Neo core needs to know the maximum transactions per block and maximum low priority transactions per block.

See how this is used in the core here: support:neo-project/neo/pull/500

### DIFF
--- a/SimplePolicy/SimplePolicyPlugin.cs
+++ b/SimplePolicy/SimplePolicyPlugin.cs
@@ -44,6 +44,9 @@ namespace Neo.Plugins
             return FilterForBlock_Policy2(transactions);
         }
 
+        public int MaxTxPerBlock { get; } = Settings.Default.MaxTransactionsPerBlock;
+        public int MaxLowPriorityTxPerBlock { get; } = Settings.Default.MaxFreeTransactionsPerBlock;
+
         private static IEnumerable<Transaction> FilterForBlock_Policy1(IEnumerable<Transaction> transactions)
         {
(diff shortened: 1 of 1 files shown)


SimplePolicy/SimplePolicyPlugin.cs
```diff
@@ -63,7 +66,7 @@ private static IEnumerable<Transaction> FilterForBlock_Policy2(IEnumerable<Trans
             Transaction[] free = tx_list.Where(p => p.IsLowPriority)
                 .OrderByDescending(p => p.NetworkFee / p.Size)
                 .ThenByDescending(p => p.NetworkFee)
-                .ThenByDescending(p => InHighPriorityList(p))
+                .ThenByDescending(InHighPriorityList)
```
shargon: I think that was more readable before this line change

jsolman: Seems like just extra characters to me, but I can put it back if others think it is more readable.

erikzhang:
Wait for neo-project/neo#500 to merge or submit a separate pull request, then merge this one.

jsolman:
neo-project/neo#500 is merged. Can I get an approval on this now?

SimplePolicy/SimplePolicyPlugin.cs
jsolman: No one else commented, but I put it back for you @shargon :)

PR #38, opened 2019-01-18
Ensure PoW correct order for TXs tiebreakers

vncoelho:
With this we should ensure the following order (if fee and fee/byte are the same):
[000000,000001, ..., 00000f, ..., ffffff]


igormcoelho:
Lets add tests for this part. Its important to be absolutely correct..

erikzhang:
Next time we should add tests before merging.

vncoelho:
You are right, Erik.
We gonna prepare more tests for the Hash (some were made for the mempool) and also the travis CI.for this repo.

PR #39, opened 2019-01-18
Expose HighPriorityTxType

vncoelho:


SimplePolicy/SimplePolicyPlugin.cs
```diff
@@ -46,6 +46,7 @@ public IEnumerable<Transaction> FilterForBlock(IEnumerable<Transaction> transact
 
         public int MaxTxPerBlock => Settings.Default.MaxTransactionsPerBlock;
         public int MaxLowPriorityTxPerBlock => Settings.Default.MaxFreeTransactionsPerBlock;
+        public EnumSet<TransactionType> HighPriorityTxType => Settings.Default.HighPriorityTxType;
```
jsolman: Lets name the method name here something that doesn't indicate high priority since it is actually going to be a low priority transaction, just sorted higher than hashes. How about name it: `HigherLowPriorityTxTypes`

vncoelho: Funny but true.

jsolman:
We can probably just sort claim transaction as higher priority in `CompareTo` in PoolItem and not really need to change this to get it from here.

vncoelho:
All rights, if that is the case we remove the expose but, at least, change the variable name for better readability.

jsolman:
It also doesn't hurt to leave the method; the method doesn't hurt anything being there even if it isn't in the interface.

igormcoelho:
This PR is nice! it would help maintaining a Mempool which is very similar to the plugin tx selection.

SimplePolicy/SimplePolicyPlugin.cs
```diff
@@ -112,6 +113,6 @@ private bool VerifySizeLimits(Transaction tx)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private static bool InHighPriorityList(Transaction tx) => Settings.Default.HighPriorityTxType.Contains(tx.Type);
+        private static bool InHigherLowPriorityList(Transaction tx) => Settings.Default.HighPriorityTxType.Contains(tx.Type);
```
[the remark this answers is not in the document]
vncoelho: ajajaja, thanks, Shargon, much simpler.

PR #40, opened 2019-01-21
Commit as the final step of persistence in StatesDumper

jsolman:
Use the new `Commit` method for performing writes. See neo-project/neo#568

### DIFF
--- a/StatesDumper/StatesDumper.cs
+++ b/StatesDumper/StatesDumper.cs
@@ -122,6 +122,12 @@ namespace Neo.Plugins
 
         public void OnCommit(Snapshot snapshot)
         {
+            if (Settings.Default.PersistAction.HasFlag(PersistActions.StorageChanges))
+                OnCommitStorage(snapshot);
+        }
+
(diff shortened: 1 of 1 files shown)


jsolman:
It probably goes without saying, but the next Neo release must be made and referenced in the csproj before this can be merged.

jsolman:
@vncoelho You should have waited to merge this probably until it is referencing the next release of Neo. This isn't compatible with version 2.9.4

vncoelho:
Your commit on Neo core already destroyed the plugin, it was needed for fixing StateDumper.

jsolman:
This project doesn't reference the commit though, it references the 2.9.4 release.

vncoelho:
Anhhhh, you are right.
In this sense, yes, since I do not use visual studio I missed that. We use a script that get all the masters and keep it compiling.
I personally think it is good, because it keeps like a Travis CI on (neo-cli,neo-vm,neo-plugins and neo).

erikzhang:
You can use VS Code. https://code.visualstudio.com

PR #43, opened 2019-03-01
Add plugin for NEP5 balances and transfer history (Persistence + RPC)

jsolman:
This plugin adds initial support for obtaining NEP5 balances and transfer history for arbitrary addresses. For those that don't need transfer history, but only care about accurate balances, the transfer history can be turned off in the config.

Some wallets have been obtaining NEP5 balances by making calls to `balanceOf` periodically for every asset. This is expensive since many of the times the balances haven't changed and are still zero, leading to unnecessary load on seed nodes. This provides a better solution that allows obtaining only the balances of assets that changed.

Two new RPC Calls:

1. `getnep5balances` : 
   * Params:
     * param[0] = account script hash
   * Response example:
     ```
     {
         "jsonrpc": "2.0",
         "id": 1,
         "result": {
             "balance": [
                 {
                     "asset_hash": "a2ca38cb66e2e572ceb3bc621e1f27cee24cad31",
                     "amount": "10000000000000",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "591eedcd379a8981edeefe04ef26207e1391904a",
                     "amount": "16986435118916704",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "05b07b48049b9474681c37cde8b7205a2d710a77",
                     "amount": "35000000000",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "025c91e6f6792e087feebb30fd4761f4fbcb4e82",
                     "amount": "100000000000",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "155153854ed377549a72cc1643e481bf25b48390",
                     "amount": "23800884731190000",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "08b9f4c2e0110b8eb650f89bbcaaaaa379497a94",
                     "amount": "100000000000",
                     "last_updated_block": 0
                 },
                 {
                     "asset_hash": "9aff1e08aea2048a26a3d2ddbb3df495b932b1e7",
                     "amount": "4920295790109",
                     "last_updated_block": 0
                 }
             ],
             "address": "AJ9uXVEth1vbSdBxQiCprPBQuyvitdB3FT"
         }
     }
     ```
2. `getnep5transfers` : 
   * Params:
     * param[0] = account script hash
     * param[1] = start time in epoch seconds (optional - defaults to 1 week in the past)
     * param[2] = end time in epoch seconds (optional - defaults to the current time)
   * Response example:
     ```
     {
         "jsonrpc": "2.0",
          "id": 1,
          "result": {
              "sent": [
                  {
                      "timestamp": 1551144880,
                      "asset_hash": "155153854ed377549a72cc1643e481bf25b48390",
                      "transfer_address": "AcbUu9M5R4zFdCWeJBxTuggk5N6NgwFdb7",
                      "amount": "1000000000000",
                      "block_index": 2368329,
                      "transfer_notify_index": 0,
                      "tx_hash": "8a73450186454f3d392a32f9ae3f075cd38ea0df57316b6469abd06b9efbb591"
                  }
              ],
              "received": [
                  {
                      "timestamp": 1551426241,
                      "asset_hash": "591eedcd379a8981edeefe04ef26207e1391904a",
                      "transfer_address": "AcbUu9M5R4zFdCWeJBxTuggk5N6NgwFdb7",
                      "amount": "80000000000",
                      "block_index": 2381735,
                      "transfer_notify_index": 1,
                      "tx_hash": "328245a6e5caf1b328fa1e3f76fffb233d2cdc18a9c56508b241cf8ce3f4aa78"
                  }
              ],
              "address": "AJ9uXVEth1vbSdBxQiCprPBQuyvitdB3FT"
          }
      }
      ```


RpcNep5Balances/RpcNep5Balances.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <None Update="RpcNep5Balances\config.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Neo" Version="2.9.5" />
```
jsolman: This depends on some recent changes to Neo expected to be in the 2.9.5 release (Specifically https://github.com/neo-project/neo/pull/613).

jsolman:
Should be working now. I've tested the balances API. Re-syncing to test transfers again after fixing byte order issue with the transfer key, but I expect it to work now.

RpcNep5Balances/RpcNep5Balances.cs
```diff
@@ -0,0 +1,259 @@
...
+using Neo.Persistence.LevelDB;
+using Snapshot = Neo.Persistence.Snapshot;
+using Neo.SmartContract;
+using Neo.Network.RPC;
+using Neo.VM;
+using Neo.Wallets;
+
+namespace Neo.Plugins
+{
+    public class RpcNep5Balances : Plugin, IPersistencePlugin, IRpcPlugin
```
jsolman: Originally when I started writing the plugin it only handles balances, since it can handle both balance and transfer history, I’ll rename the plugin to `RpcNep5Plugin`

jsolman: Decided to rename to `RpcNep5Tracker`.

jsolman:
Should be ready to go now. To use it prior to 2.9.5 being released, you need to run using the commit form https://github.com/neo-project/neo/pull/613 or later.

RpcNep5Tracker/RpcNep5Tracker/config.json
```diff
@@ -0,0 +1,6 @@
+﻿{
+  "PluginConfiguration": {
+    "DBPath": "Nep5BalanceData",
+    "TrackHistory" : true
```
erikzhang: Should this be `false` by default?

PeterLinX:
We'll test it ASAP. I'm not sure if this plugin can be released as 2.9.5 for neo-cli 2.9.5 is quite urgent.

jsolman:
That’s fine if it can’t be released initially when 2.9.5 is released. It can be pushed any time after 2.9.5 is released as it is a plugin compatible with 2.9.5 of Neo core.

RpcNep5Tracker/RpcNep5Tracker/config.json
jsolman: I guess it depends whether people want less by default or more by default. My thought was people would want more by default.

jsolman: One other thought is that you could turn it off later and remove the keys from leveldb more easily than turning it on later and having to resync the chain.

jsolman:
Currently using this plugin requires resyncing the chain if it were installed after blocks had already been persisted. It is possible to make a change so that this plugin can be turned on later and would index the data from the past blocks so that it could be installed on a node without requiring having to resync the chain.

jsolman:
I believe the plugin to be in a properly working state from my testing. It should have an independent test performed though, and I believe @PeterLinX will do that once they finish testing 2.9.5 and it gets released. 

@vncoelho if you are looking for rich list functionality for tokens, that functionality could be build without that much trouble while only requiring to index the top `X` holders balances, since balances are known each time they change. If I get time I'll see about adding the feature and an API and a setting toggle for whether to turn it on. It could always be added in a separate PR though if I don't get around to it in time.

vncoelho:
Perfect, Jeff.
Let's consider version 2.10.0 instead of 2.9.5.

jsolman:
Once Merkle Patricia tree is in, it would be cool if there was a new token standard that mandates storage keys that are used for balances. This would also allow getting the hashes in the merkle patricia tree to the root to prove balance without trusting the node providing it.
Note: It would be awesome if that standard also could allow for a contract to manage multiple tokens.

I may create an issue to track this.

vncoelho:
MPT opens this nice possibilities and safety, as emphasized by @igormcoelho and @rodoufu.

superboyiii:
@jsolman If param[0] is a standard address might be more convenient for using, I think.

jsolman:
> @jsolman If param[0] is a standard address might be more convenient for using, I think.

I can make it detect whether it is an address or script hash so that it supports both.

jsolman:
I still need to add the new IRpcPlugin interface methods. I will do this Sunday, and test again with the latest code, then it will be ready.

erikzhang:
@jsolman They have already been added.

jsolman:
@erikzhang
Ah, I missed that you added them. Ok. I’ll merge it.

RpcNep5Tracker/Nep5Balance.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Numerics;
+using Neo.IO;
+using Neo.Ledger;
+
+namespace Neo.Plugins
+{
+    public class Nep5Balance : StateBase, ICloneable<Nep5Balance>
+    {
+        public BigInteger Balance;
+        public uint LastUpdatedBlock;
```
shargon: Is not serialized?

RpcNep5Tracker/Nep5Transfer.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Neo.Plugins
+{
+    public class Nep5Transfer : StateBase, ICloneable<Nep5Transfer>
+    {
+        public UInt160 UserScriptHash;
+        public uint BlockIndex;
+        public UInt256 TxHash;
+        public BigInteger Amount;
+
+        public override int Size => base.Size + 20 + Amount.ToByteArray().GetVarSize();
```
shargon: `20 + 4 + 32+ Amount.ToByteArray().GetVarSize();` ?

RpcNep5Tracker/Nep5Balance.cs
jsolman: Nice catch and fix in #67 
I added this field after the initial implementation and did not adjust the sizes. I should have written a unit test.

RpcNep5Tracker/Nep5Transfer.cs
jsolman: Your fix in #67 has been merged. Nice job.

PR #44, opened 2019-03-05
Add RpcWallet

erikzhang:


jsolman:
Should all the code that deals with maintaining wallet db indices be moved into the plugin also? Did you plan to do this still?

jsolman:
Actually I guess it is more work if an attempt were made to move wallet tracking into the plugin since the cli will still interact with the wallet directly. It would need other interfaces for the cli to interact with the plugin. Still it is perhaps something to consider for the future.

shargon:
Good move :)

erikzhang:
Waiting for the test from NGD.

PR #47, opened 2019-03-06
v2.10.0

vncoelho:


vncoelho:
should we remove `net47` and just leave `netstandard2.0`.

vncoelho:
I think I will close this, because the others PRs are updating each projects separately and version 2.10.1 is already on the way.

PR #48, opened 2019-03-06
Fix missing public declaration on pre/post process functions

vncoelho:


erikzhang:
Not good. 😕

erikzhang:
@shargon You need to test before approving.
@vncoelho You need to test before merging.

jsolman:
FYI: The csproj file still references 2.9.4 where the interface for RpcPlugin doesn’t even contain these methods.

erikzhang:
> The csproj file still references 2.9.4 where the interface for RpcPlugin doesn’t even contain these methods.

Yes. So why do we create one pr for adding `PreProcess()` and `PostProcess()`, one pr for adding `public`, and another pr for adding reference to 2.10.0?

jsolman:
Vitor was in a hurry for some reason. You will have to have him answe it. @vncoelho

shargon:
Maybe should be better with `protect` ?

vncoelho:
The masters should build and that inclusion in NEO core had broken 2 plugins.

The public was a minor problem, it should had come with the function.

vncoelho:
The public was copied from the RPC wallet but it probably could be protected.

PR #53, opened 2019-03-11
Import blocks after plugins are loaded.

jsolman:
All persist plugins need to be loaded before blocks start importing.

Depends on neo-project/neo/pull/633

### DIFF
--- a/ImportBlocks/ImportBlocks.csproj
+++ b/ImportBlocks/ImportBlocks.csproj
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="2.10.1" />
+    <PackageReference Include="Neo" Version="2.10.1-CI00002" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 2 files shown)


ImportBlocks/ImportBlocks.csproj
```diff
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="2.9.4" />
+    <PackageReference Include="Neo" Version="2.10.1" />
```
jsolman: Can't be pushed anyway until this reference is available.

jsolman: I can update this to a NuGet build, once https://github.com/neo-project/neo/pull/633 is pushed.

vncoelho: That is it, Jeff.

jsolman:
This also still references NuGet CI build, thus waiting to merge.

PR #54, opened 2019-03-12
Check authentication during pre-processing of the rpc request.

jsolman:
Order plugins are run could allow another plugin to process something before the `RpcSecurity` plugin throws an exception. This fixes the issue by having `RpcSecurity` throw from `PreProcess`.


jsolman:
One thing to be careful with is not to load this plugin after this change in an older version before 2.10.0, or else it would potentially have no security, since the `PreProcess` method would not be called in that case, but the plugin might actually appear to load successfully.

PR #58, opened 2019-03-14
Adding settings to RPCWallet for MaxGasInvoke limit from Settings

vncoelho:
Closes https://github.com/neo-project/neo-cli/issues/318

### DIFF
--- a/RpcWallet/RpcWallet.cs
+++ b/RpcWallet/RpcWallet.cs
@@ -128,7 +128,7 @@ namespace Neo.Plugins
                     Script = result["script"].AsString().HexToBytes(),
                     Gas = Fixed8.Parse(result["gas_consumed"].AsString())
                 };
-                tx.Gas -= Settings.Default.MaxGasInvoke;
+                tx.Gas -= Fixed8.FromDecimal(10);
                 if (tx.Gas < Fixed8.Zero) tx.Gas = Fixed8.Zero;
                 tx.Gas = tx.Gas.Ceiling();
                 tx = Wallet.MakeTransaction(tx);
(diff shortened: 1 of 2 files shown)


RpcWallet/RpcWallet.cs
```diff
@@ -127,7 +128,7 @@ private void ProcessInvoke(JObject result)
                     Script = result["script"].AsString().HexToBytes(),
                     Gas = Fixed8.Parse(result["gas_consumed"].AsString())
                 };
-                tx.Gas -= Fixed8.FromDecimal(10);
+                tx.Gas -= Settings.Default.MaxGasInvoke;
```
erikzhang: Should be 10 gas (the free 10 gas for each transaction)

vncoelho: In fact, I did not understand this part very well.

vncoelho: What is the best way to limit the `Wallet.MakeTransaction(tx);` gas limit simulation?

erikzhang: It created an InvocationTransaction for the `sendrawtransaction`.

vncoelho: Maybe add an extra parameter on the https://github.com/neo-project/neo/blob/cc0f315854a9cf09f2efda2f83cf1199c9ece4a9/neo/Wallets/Wallet.cs#L223?

vncoelho: Or just try to add simulate that to  tx.SystemFee = Settings.Default.MaxGasInvoke;

vncoelho: I still did not understand the direction to implement this.

vncoelho: I think I will leave this closed, maybe this can be solved direction on Neo RpcServer.cs class.

PR #55, opened 2019-03-12
Fix application logs after adding EventStream usage and OnPluginsLoaded.

jsolman:
Closes #50

PR #52, opened 2019-03-11
Add Plugin for UTXO tracking by account.

jsolman:
This plugin facilitates sending NEO or GAS for arbitrary addresses by providing an RPC method `getunspents` for retrieving the unspent UTXOs. The unspent UTXOs are stored indexed by account script hash in an additional LevelDB database whose path is configurable in the plugin settings. This plugin was migrated from my initial PR here: neo-project/neo/pull/569

The format of the response data is compatible with the format returned from the neoscan v1 API get_balance` call.

PR #61, opened 2019-03-14
Aggregate - Plugin Improvements for Neo 2.10.1

jsolman:
This rolls up the following changes:
* Fix application logs after adding EventStream usage and OnPluginsLoaded. (#55)
* Import blocks after plugins are loaded. (#53)
* Add Plugin for UTXO tracking by account. (#52)


erikzhang:
I prefer that each PR submit a separate feature instead of a bunch of changes mixed together.

jsolman:
I can break them back apart, but there are individual PRs for each in this aggregate.

jsolman:
In this case people that want to use 2.10.1 plugins really want to have one branch to get them all, but I don’t necessarily have to make a single PR for it, so I can break it back out into multiple PRs against master if you want.

erikzhang:
Yes, please break it back out into multiple PRs against master.

vncoelho:
It is also my fault, @erikzhang, I was also supporting this for an easy way to test all together. I believe that after revised it becomes easier, but this case was outstanding because we needed to change 3 plugins.

If you prefer everything split let's split it back.

jsolman:
I will split it back into multiple PRs against master. I do kind of think there should be a branch for 2.10.0 and 2.10.1 though, because for example the 2.10.0 and 2.10.1 important blocks plugin need to be different. I’ll have a pr out against master for both, we can merge the 2.10.0 one first since it works with the current neo-cli release, but devs and some others may need the code that works against 2.10.1.

jsolman:
Changes for ApplicationLogs plugin, Import Blocks Plugin, and UTXO tracking plugin have been split out independently.

PR #62, opened 2019-03-14
Use IPersistencePlugin and batch Application Log writes.

### LINKED ISSUE #60: Application logs plugin should use batch write
If you kill neo-cli using control-c while using the application log plugin it can lead to inconsistency in application logs.

jsolman:
Closes #60


jsolman:
Needs testing.

vncoelho:
Nice, @jsolman, I will review asap, let me just double check 2.10.1 branch.

As you mentioned, If this is good we can even remove that applog modification there.

ApplicationLogs/LogReader.cs
```diff
@@ -39,5 +45,59 @@ public JObject OnProcess(HttpContext context, string method, JArray _params)
...
+                        }
+                        catch (InvalidOperationException)
+                        {
+                            notification["state"] = "error: recursive reference";
+                        }
+                        return notification;
+                    }).ToArray();
+                    return execution;
+                }).ToArray();
+                db.Put(WriteOptions.Default, appExec.Transaction.Hash.ToArray(), json.ToString());
```
erikzhang: I think this line should be removed?

ApplicationLogs/LogReader.cs
```diff
@@ -39,5 +45,59 @@ public JObject OnProcess(HttpContext context, string method, JArray _params)
...
+                writeBatch.Put(appExec.Transaction.Hash.ToArray(), json.ToString());
+            }
+        }
+
+        public void OnCommit(Snapshot snapshot)
+        {
+            db.Write(WriteOptions.Default, writeBatch);
```
erikzhang: Why not call `db.Write()` in `OnPersist()`?

jsolman: Because `OnCommit` is where things are supposed to be committed to a database, right?

ApplicationLogs/LogReader.cs
jsolman: Yep, not sure, how that got left in there.

ApplicationLogs/LogReader.cs
jsolman: I guess it is better to write the batch in `OnPersist()` though since the data should be idempotent if you write it twice.

jsolman: I moved it to `OnPersist()`

ApplicationLogs/LogReader.cs
jsolman: Fixed.

PR #65, opened 2019-03-16
Add Plugin for UTXO tracking by account.

jsolman:
This plugin facilitates sending NEO or GAS for arbitrary addresses by providing an RPC method `getunspents` for retrieving the unspent UTXOs. The unspent UTXOs are stored indexed by account script hash in an additional LevelDB database whose path is configurable in the plugin settings. This plugin was migrated from my initial PR here: neo-project/neo/pull/569

Additionally the plugin supports getting the claimable UTXOs as well as the total unclaimed amount of GAS available by introducing `getclaimable` and `getunclaimed` RPC methods.

The formats of the response data are compatible with the format returned from the neoscan v1 API calls: `get_balance`, `get_clamaible`, get_unclaimed`.


jsolman:
Testing after syncing with MainNet looks good. Keep in mind that to use this you need:
1. To be using Neo 2.10.1 (currently master at or above commit https://github.com/neo-project/neo/commit/432a3f6b5ca4389f7a2221fc845edf770e32b0f7)
2. The updated ImportBlocks plugin: #66

vncoelho:
@shargon, what do you think?
It is a big responsibility even as a Plugin because some projects may really start to use it.
Maybe if there are errors they will be minor.

RpcSystemAssetTracker/UserSystemAssetCoinOutputsKey.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (userAddressComparison != 0) return userAddressComparison;
+            return Comparer<UInt256>.Default.Compare(TxHash, other.TxHash);
+        }
+
+        public override int GetHashCode()
+        {
+            unchecked
+            {
+                var hashCode = IsGoverningToken.GetHashCode();
+                hashCode = (hashCode * 397) ^ (UserAddress != null ? UserAddress.GetHashCode() : 0);
```
shargon: Why 397?

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
```diff
@@ -0,0 +1,427 @@
...
+        private DataCache<UserSystemAssetCoinOutputsKey, UserSystemAssetCoinOutputs> _userSpentUnclaimedCoins;
+        private WriteBatch _writeBatch;
+        private int _rpcMaxUnspents;
+        private uint _lastPersistedBlock;
+        private bool _shouldPersistBlock;
+        private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
+
+        public override void Configure()
+        {
+            if (_db == null)
```
shargon: If is not `null` we shoud be able to change it, because we can change the path in the config file.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
```diff
@@ -0,0 +1,427 @@
...
+using Neo.Ledger;
+using Neo.Persistence;
+using Snapshot = Neo.Persistence.Snapshot;
+
+namespace Neo.Plugins
+{
+    public class RpcSystemAssetTrackerPlugin : Plugin, IPersistencePlugin, IRpcPlugin
+    {
+        private const byte SystemAssetUnspentCoinsPrefix = 0xfb;
+        private const byte SystemAssetSpentUnclaimedCoinsPrefix = 0xfc;
```
shargon: We should have a global prefix for plugins, and concat this prefix with this one, in order to prevent collisions.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
jsolman: Changing the file after syncing the chain would have to restart the plugin currently; it also isn’t very user friendly to do this right now, as it would have to reprocess all the blocks up to the current, which will take quite a long time and not indicate any progress anywhere.

jsolman: Well it can’t use the same database as the core is using, so it won’t collide with it. The method it uses takes a 1 byte prefix. We could have an additional prefix, but it doesn’t really seem needed right now.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
jsolman: Well it can’t use the same database as the core is using, so it won’t collide with it. The method it uses takes a 1 byte prefix. We could have an additional prefix, but it doesn’t really seem needed right now.

RpcSystemAssetTracker/UserSystemAssetCoinOutputsKey.cs
jsolman: It is Prime and my hash code generator plugin seemed to like it.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
shargon: If is a different storage there is no problem.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
shargon: But if we allow to change the config file and refresh the plugins, is weird that this feature works only for certain properties, isn't it?

jsolman: Well I suppose I could make it support changing the database it uses, but it will need to cause it to reload data from all blocks. Maybe I'll add support for it, but it doesn't really seem like that useful of a feature for the effort.

Supporting reloading for `TrackUnclaimed` doesn't seem very useful either. If you turn it off and turn it back on it makes the claim information inconsistent.

What is probably the most useful and should be supported is reloading the `MaxReturnedUnspents`, it should be easy to support that.

jsolman: Added support for reloading `MaxReturnedUnspents`. I don't think it is worth adding support for changing the other settings.

RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
```diff
@@ -0,0 +1,426 @@
...
+            using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
+            {
+                var storeSpentCoins = snapshot.SpentCoins;
+                byte[] prefix = new [] { (byte) 1 }.Concat(scriptHash.ToArray()).ToArray();
+                foreach (var claimableInTx in dbCache.Find(prefix))
+                    if (!AddClaims(claimable, ref totalUnclaimed, _rpcMaxUnspents, snapshot, storeSpentCoins,
+                        claimableInTx))
+                        break;
+            }
+            json["unclaimed"] = (double) (decimal) totalUnclaimed;
```
vncoelho: @jsolman, maybe this should be `json["available"]` in order to be consistent with `ProcessGetUnclaimed()`

jsolman: It currently is matching the neoscan api to use `unvlaimed`. I think it is better to match their response to make less work for those that already have code that uses their responses.

vncoelho: I think that it is good that if matches.
You are right, Jeff.

In addition, I think that there is not a standard, I was just checking the NEL explorer and they have different types of calls: http://www.xiaoyaoji.cn/share/1H0gjTDtfk/1dLQ8CDVdd

vncoelho:
Waiting for next release to be merged.

alvarodh5:
The plugin seems to work as expected:

**getunspents**
```{"jsonrpc":"2.0","id":1,"result":{"balance":[{"unspent":[{"txid":"5a736f741f456c9381759e87868b2a07c5801c4caa2c6e8e3da42ad427be800c","n":1,"value":495}],"asset_hash":"602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7","asset":"GAS","asset_symbol":"GAS","amount":495},{"unspent":[{"txid":"b5a1601408a1784fe4c64c3d366f72f7b1c1261663e8d8b1bf0cf33bfe6d72f1","n":0,"value":5}],"asset_hash":"c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b","asset":"NEO","asset_symbol":"NEO","amount":5}```

image

**getclaimable**
```{"jsonrpc":"2.0","id":1,"result":{"claimable":[{"txid":"4ba4d1f1acf7c6648ced8824aa2cd3e8f836f59e7071340e0c440d099a508cff","n":0,"value":100000000,"start_height":2007,"end_height":2445,"generated":3504,"sys_fee":0,"unclaimed":3504}],"address":"AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y","unclaimed":3504}}```

image

**getunclaimed**
```{"jsonrpc":"2.0","id":1,"result":{"available":3504,"unavailable":519,999974,"unclaimed":4023,999974}}```

image

**Comment by erikzhang** [comment]
This plugin didn't modify the .sln file. @vncoelho

**Comment by erikzhang** [comment]
After approving an pull request, please leave me at least 1 day to do the final check before merge.

**Comment by vncoelho** [comment]
Go for it. I am not used with these .sln file.
This PR has been pending for a long time.

**Comment by vncoelho** [comment]
This .sln can be included in the PR for adjusting all plugins to 2.10.1.

File: RpcSystemAssetTracker/UserSystemAssetCoinOutputs.cs [hunk 5]
```diff
@@ -0,0 +1,79 @@
...
+                {
+                    writer.Write(txIndex.Key);
+                    writer.Write(txIndex.Value);
+                }
+            }
+
+            public override void Deserialize(BinaryReader reader)
+            {
+                base.Deserialize(reader);
+                ((ISerializable)TotalAmount).Deserialize(reader);
```
shargon: If we call `AddTxIndex` we don't need to store `TotalAmount`. What do you think @jsolman ?

PR #66, opened 2019-03-16
Properly stop importing blocks on system shutdown.

### LINKED ISSUE #59: Wallet doesn’t exit properly during ImportBlocks
The wallet does not stop properly when typing exit while in the processes of syncing using the ImportBlocks plugin.

### LINKED ISSUE #77: [2.10.1] Can't run with ImportBlocks after update
Had some trouble starting neo-cli since the update to 2.10.1. This is what I got in the error log. After removing ImportBlocks, it runs as expected.

```
System.ArgumentException
Unable to resolve the target Provider
Parameter name: self
   at Akka.Actor.Futures.Ask[T](ICanTell self, Func`2 messageFactory, Nullable`1 timeout, CancellationToken cancellationToken)
   at Neo.Plugins.ImportBlocks.OnImport()
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
```

jsolman:
Closes #59 
Closes #77 

This change makes it possible to safely stop the importing blocks process such as through `exit` in `neo-cli`.

This also includes a fix so that Importing blocks does not begin until other plugins complete loading.


vncoelho:
@shargon @lightszero @erikzhang @belane, are you going to revise this one? It is a critical plugin.
By the way, congratulations for the NNL team for the great plugin for syncing!
Could it be merged here with an option for that?

I believe that some fix and adjustment is needed. I tested this PR for syncing Testnet and it went ok. Now I am syncing Mainnet.

ImportBlocks/ImportBlocks.cs
```diff
@@ -4,52 +4,32 @@
...
-            return false;
+            Settings.Load(GetConfiguration());
         }
 
-        public override void Configure()
+        protected override void OnPluginsLoaded()
         {
-            Settings.Load(GetConfiguration());
+            SuspendNodeStartup();
+            _blockImporter.Tell(new BlockImporter.StartImport { BlockchainActorRef = System.Blockchain, OnComplete = OnImportComplete});
```
vncoelho: @jsolman, why there is ImportCompleted and ImportComplete?

shargon:
I need more time for testing this sorry

ImportBlocks/ImportBlocks.cs
jsolman: `Blockchain.ImportCompleted` is what is sent back when a `Blockchain.Import` message finishes processing. The `BlockImporter` will issue many `Blockchain.Import` messages and get back many `Blockchain.ImportCompleted` messages. When all importing is complete it will call back whatever action is passed in the `OnComplete` member of the `BlockImporter.StartImport` message. In this case it is passing the `OnImportComplete` method to be called when all importing is finished.

jsolman:
This likely needs the `Blockchain` to use a `PinnedDispatcher` as discussed here: https://github.com/neo-project/neo/pull/680

erikzhang:
https://github.com/neo-project/neo/blob/5ce0e4c3192cdcce700105030ed03197961e0466/neo/Ledger/Blockchain.cs#L226-L237

I think the easiest way to stop importing blocks on system shutdown is to break this loop.

jsolman:
This plugin no works properly, with this change: https://github.com/neo-project/neo/pull/681

jsolman:
> I think the easiest way to stop importing blocks on system shutdown is to break this loop.

What's wrong with this solution? It is elegant and works well. The actors don't schedule an over abundance of work for a single message and they stop naturally. If we want it to break the loop after less blocks, we can reduce how many are sent in each batch when sending the `Import` message.

erikzhang:
I means that break the loop when shutting down.

jsolman:
> I means that break the loop when shutting down.

Why is it necessary, when this solution works well?

jsolman:
Specifically, it would have to have a check to know it was shutting down inside the `Blockchain` actor. It seems like an anti-pattern.

ImportBlocks/BlockImporter.cs
```diff
@@ -0,0 +1,125 @@
...
+
+namespace Neo.Plugins
+{
+    public class BlockImporter : UntypedActor
+    {
+        public class StartImport { public IActorRef BlockchainActorRef; public Action OnComplete; }
+        private IActorRef _blockchainActorRef;
+        private const int BlocksPerBatch = 10;
+        private bool isImporting;
+        private IEnumerator<Block> blocksBeingImported;
```
erikzhang: We need to call `blocksBeingImported.Dispose()` after import.

ImportBlocks/BlockImporter.cs
```diff
@@ -0,0 +1,125 @@
...
+using Neo.Ledger;
+using Neo.Network.P2P.Payloads;
+
+namespace Neo.Plugins
+{
+    public class BlockImporter : UntypedActor
+    {
+        public class StartImport { public IActorRef BlockchainActorRef; public Action OnComplete; }
+        private IActorRef _blockchainActorRef;
+        private const int BlocksPerBatch = 10;
```
erikzhang: Does this affect the import speed? Maybe we should turn `BlocksPerBatch` into a configurable parameter.

jsolman: It has very little impact on import speed. The time to receive the next batch is negligible compared to the time of importing a block. We could make it configurable, but it probably isn’t necessary.

ImportBlocks/BlockImporter.cs
jsolman: Nice catch. Fixed.

jsolman:
I'm testing with these changes and it looks good. I was having trouble syncing MainNet chain using either these changes or the `ImportBlocks` plugin before these changes; specifically, I was getting exceptions during `OnPersist` due to missing keys from LevelDB. My problems were resolved by rebuilding LevelDB with this patch applied: https://github.com/google/leveldb/pull/339

This may be very helpful for others that possibly encounter this issue. I believe https://github.com/google/leveldb/pull/339 will very soon be merged into the official LevelDB repository; it's amazing this bug has been known for so long (3+ years), and is only now about to get merged into the official LevelDB repository.

PR #68, opened 2019-03-19
Plugin CoreMetrics

vncoelho:
This plugin intends to support core metrics for the Blockchain.

Currently, it contains a method for getting last timestamps of blocks based on `height` and `number of requested blocks`.
This current API is quite needed because front-end services are currently relying on other db services for querying blocks timestamps (because it is impracticable to query block header and then filter).


vncoelho:
The prefix `getmetrics` sounds good, @igormcoelho.
Thanks for the project organization @erikzhang.

PR #70, opened 2019-03-21
Make calculation for extra gas on "send*" RPC

superboyiii:
Sync to SimplePolicy. But since the input UTXOs are unknown, the extra gas might make the transaction size bigger again. The subtraction of 1024 is not been added for some headroom.

### DIFF
--- a/RpcWallet/RpcWallet.cs
+++ b/RpcWallet/RpcWallet.cs
@@ -309,7 +309,7 @@ namespace Neo.Plugins
             }, from: from, change_address: change_address, fee: fee);
             if (tx.Size > 1024)
             {
-                fee += Fixed8.FromDecimal(tx.Size * 0.00001m);
+                fee += Fixed8.FromDecimal(tx.Size * 0.00001m + 0.001m);
                 tx = Wallet.MakeTransaction(null, new[]
                 {
                     new TransferOutput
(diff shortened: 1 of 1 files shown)


RpcWallet/RpcWallet.cs
```diff
@@ -334,6 +347,11 @@ private JObject SendMany(UInt160 from, JArray to, Fixed8 fee, UInt160 change_add
             if (fee < Fixed8.Zero)
                 throw new RpcException(-32602, "Invalid params");
             Transaction tx = Wallet.MakeTransaction(null, outputs, from: from, change_address: change_address, fee: fee);
+            if (tx.Size > 1024)
+            {
+                fee += Fixed8.FromDecimal(tx.Size * 0.00001m);
```
jsolman: If fee is less than `.0005` Before this line probably should add the difference so it will have some headroom in case more gas utxo’s have to be added to accommodate the additional tx size.

superboyiii: Just fixed it.

erikzhang:
Maybe we should take the `fee` parameter as `max_fee`. The automatically calculated `fee` should not exceed `max_fee`, otherwise the caller cannot estimate the cost.

PR #71, opened 2019-03-22
RpcNep5Tracker: Handle NEP5 transfers to/from null & limit max results

jsolman:
Minted NEP5 tokens are sometimes sent from null. Also some contracts destroy tokens sending to null. These changes should handle tracking their balances.

This adds support to the RpcNep5Tracker plugin for: https://github.com/neo-project/proposals/pull/44

Also, while not a standard, at least one popular NEP5 token used a `mintTokens` event to track minting of its tokens. This adds support for it in order to track balances of the known currently in use NEP5 tokens.

Warning: *Currently this code needs testing*

### DIFF
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -32,6 +32,7 @@ namespace Neo.Plugins
         private bool _shouldTrackHistory;
         private bool _recordNullAddressHistory;
         private uint _maxResults;
+        private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
 
         public override void Configure()
@@ -44,6 +45,7 @@ namespace Neo.Plugins
             _shouldTrackHistory = (GetConfiguration().GetSection("TrackHistory").Value ?? true.ToString()) != false.ToString();
             _recordNullAddressHistory = (GetConfiguration().GetSection("RecordNullAddressHistory").Value ?? false.ToString()) != false.ToString();
             _maxResults = uint.Parse(GetConfiguration().GetSection("MaxResults").Value ?? "1000");
+            _shouldTrackNonStandardMintTokensEvent = (GetConfiguration().GetSection("TrackNonStandardMintTokens").Value ?? false.ToString()) != false.ToString();
         }
 
         private void ResetBatch()
@@ -102,7 +104,7 @@ namespace Neo.Plugins
             if (!(stateItems[0] is VM.Types.ByteArray)) return;
             var eventName = Encoding.UTF8.GetString(stateItems[0].GetByteArray());
 
-            if (eventName == "mintTokens")
+            if (_shouldTrackNonStandardMintTokensEvent && eventName == "mintTokens")
             {
                 if (stateItems.Count < 4) return;
                 // This is not an official standard but at least one token uses it, and so it is needed for proper
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -60,7 +60,7 @@ namespace Neo.Plugins
 
         private void RecordTransferHistory(Snapshot snapshot, UInt160 scriptHash, UInt160 from, UInt160 to, BigInteger amount, UInt256 txHash, ref ushort transferIndex)
         {
-
+            if (!_shouldTrackHistory) return;
             _transfersSent.Add(new Nep5TransferKey(from,
                     snapshot.GetHeader(snapshot.Height).Timestamp, scriptHash, transferIndex),
                 new Nep5Transfer
@@ -137,8 +137,6 @@ namespace Neo.Plugins
                 var toKey = new Nep5BalanceKey(to, scriptHash);
                 if (!nep5BalancesChanged.ContainsKey(toKey)) nep5BalancesChanged.Add(toKey, new Nep5Balance());
             }
-
-            if (!_shouldTrackHistory) return;
             RecordTransferHistory(snapshot, scriptHash, from, to, amountItem.GetBigInteger(), transaction.Hash, ref transferIndex);
         }
 
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -97,12 +97,14 @@ namespace Neo.Plugins
             VM.Types.Array stateItems,
             Dictionary<Nep5BalanceKey, Nep5Balance> nep5BalancesChanged, ref ushort transferIndex)
         {
+            if (stateItems.Count == 0) return;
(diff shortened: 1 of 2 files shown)


jsolman:
@apisit I added the support you wanted for `mintTokens`. Please test and let me know if this is working for you.

jsolman:
I'm syncing `TestNet` with these changes now.

jsolman:
Re-syncing `TestNet` after 0e82d2e

jsolman:
Actually, maybe I should make it configurable whether to record history for the null address. Some people might want it. Since it allows to look and see all addresses that burned tokens between a particular time period or to look and see all tokens minted between a certain time period.

jsolman:
It should be robust now.

jsolman:
Syncing `MainNet` now.

erikzhang:
https://github.com/neo-project/proposals/blob/master/nep-5.mediawiki#events

In NEP-5 specification:

> A token contract which creates new tokens MUST trigger a transfer event with the from address set to null when tokens are created.
A token contract which burns tokens MUST trigger a transfer event with the to address set to null when tokens are burned.

`null` address is a part of NEP-5 standard, so I think you needn't make it configurable.

apisit:
finished syncing MainNet. 
Tested checking both ASA and LX.  triggering by event `transfer(null,...)` and `mintTokens()` work as expected.  token show up in addresses 
Thanks @jsolman

jsolman:
> `null` address is part of NEP-5 standard, so I think you needn't make it configurable.

The only thing that is configurable is whether to index the transfer history for the UInt160.Zero address. It still tracks transfers minted to addresses or destroyed from addresses with the `RecordNullAddressHistory` flag turned off. That means people can still see the history for their address when tokens were sent to them from null or when they sent tokens to null. Most people wouldn’t need the feature of seeing history of the null address itself, but turning this feature on allows querying to see all tokens minted or destroyed by anyone during a particular time range. 

So I have turned it off by default, but if someone wants to be able to query for that information they could turn it on. Most light wallet clients wouldn’t need this turned on, since they only need to check transfer history for specific wallet addresses.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -58,58 +62,96 @@ private void ResetBatch()
...
             VM.Types.Array stateItems,
             Dictionary<Nep5BalanceKey, Nep5Balance> nep5BalancesChanged, ref ushort transferIndex)
         {
+            if (stateItems.Count == 0) return;
             // Event name should be encoded as a byte array.
             if (!(stateItems[0] is VM.Types.ByteArray)) return;
             var eventName = Encoding.UTF8.GetString(stateItems[0].GetByteArray());
 
-            // Only care about transfers
+            if (eventName == "mintTokens")
```
jsolman: @erikzhang LX is the only MainNet NEP5 token using this non-standard method I think and they are going to contract migrate their NEP5 token; I could just remove handling this. What do you think? 

Also if a NEP5 token uses contract migrate it is problematic for this plugin. Maybe it could detect contract migrate of NEP5 tokens and migrate all the balances in leveldb, but it is not at all efficient.

erikzhang: I think we can remove this first. As for contract migrate, can we detect it? Maybe future token contracts should use dynamic calls to prevent changes to the hash when upgrading the contract.

jsolman: > can we detect it?

We probably could detect it; however, migrating all the balance key space in one block is definitely not a good design and also could be a potential DDOS vector.

> Maybe future token contracts should use dynamic calls to prevent changes to the hash when upgrading the contract.

I think it is a great idea to suggest the use of dynamic calls for a contract that may need upgrade so that the NEP5 contract hash would remain the same if contract logic needs to change. We should recommend this as a best practice for contract owners. However, it would be hard to enforce such a thing.

We should give guidance that NEP5 tokens should never have their hash changed; doing so requires an airdrop of their new token to existing holders, because balances will not be tracked by just migrating storage. Alternatively, we should give guidance to contract owners that if they have a NEP5 contract and think they may need to upgrade it's logic in the future using migrate, they should write dynamic invoke with logic in another contract as you suggest so that they prevent changes to the hash.

jsolman: > I think we can remove this first.

I switched it to a config option, and disabled it by default.

erikzhang: > We probably could detect it; however, migrating all the balance key space in one block is definitely not a good design and also could be a potential DDOS vector.

Yes, I also thought about this issue. Maybe we should open an issue separately to discuss the solution to this problem. What I can think of right now is not to do a storage migration when upgrading the contract, but to add a redirect record.

erikzhang:
@jsolman Are you going to work on detecting contract migrate? Or not?

jsolman:
I will definitely not add support for it in this PR. Also, I don’t have any plan to work on that feature any time soon. If you have a good idea on how to support it with a redirect record maybe you can take the time to implement it after merging this.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -30,16 +30,22 @@ public class RpcNep5Tracker : Plugin, IPersistencePlugin, IRpcPlugin
         private DataCache<Nep5TransferKey, Nep5Transfer> _transfersReceived;
         private WriteBatch _writeBatch;
         private bool _shouldTrackHistory;
+        private bool _recordNullAddressHistory;
+        private uint _maxResults;
+        private bool _shouldTrackNonStandardMintTokensEvent;
```
erikzhang: Does this need to be added to `config.json`?

jsolman: There is no necessity to add it. It will default to false.

superboyiii:
@jsolman I made a test for NEP5Tracker on mainnet yesterday. ### getnep5balances  worked well as expected, but ### getnep5transfers didn't work until I sync to the latest height. However, it works well on my privatenet. The only difference is the version of consensus. The mainnet is 2.9.4 and my privatenet is 2.10.2 preview3. So in order to prove that, we also applied it on testnet since the consensus version is 2.10.X. The result is the same as my privatenet. It worked as expected.

superboyiii:
For example, on mainnet, I choose a random address "AZsvkf5fopsUvfMbsHfx9qKmwswtgdsgZv" which has an transaction on height of "1768508". 
The txid is "0f2896032f3ad8614ce50e87b0db17a028bba976d25715f517c9201de82ee7d5". Then I use **getnep5balances** to get all nep5 transaction history. But the result is like this.
image

jsolman:
@superboyiii by default getnep5transfers only shows history for the last 7 days. You need to specify the parameter if you want further back history.

superboyiii:
Wow.....I see. Got it. Thanks for your explanation, Jeff.

superboyiii:
It works well as expected. Hahhhha.

PR #73, opened 2019-03-28
a fix suggest about extra charge for transactions over 1024 bytes

lightszero:
#neo-project/neo-cli#303

因为支付是按照交易容量计算 tx.size, 这会造成一个问题，难于处理交易的手续费

当写好了一个交易，增加手续费的输入输出时，这个动作本身就会导致txsize 变化。
而且每一个 txinput 会需要大概0.0034gas的手续费，如果一个txinput 小于这个数字，会陷入死循环

故而提出意见 txinput 不计入收费容量，可解决这个问题。
而且使用大量的txinput 构造交易，也没有很强的可攻击性。

### DIFF
--- a/RpcWallet/RpcWallet.cs
+++ b/RpcWallet/RpcWallet.cs
@@ -1,4 +1,4 @@
-﻿using Akka.Actor;
+using Akka.Actor;
 using Microsoft.AspNetCore.Http;
 using Neo.IO;
 using Neo.IO.Json;
@@ -307,6 +307,19 @@ namespace Neo.Plugins
                     ScriptHash = to
                 }
             }, from: from, change_address: change_address, fee: fee);
+            if (tx.Size > 1024)
+            {
+                fee += Fixed8.FromDecimal(tx.Size * 0.00001m + 0.001m);
+                tx = Wallet.MakeTransaction(null, new[]
+                {
+                    new TransferOutput
+                        {
+                            AssetId = assetId,
+                            Value = amount,
+                            ScriptHash = to
+                        }
+                }, from: from, change_address: change_address, fee: fee);
+            }
             if (tx == null)
                 throw new RpcException(-300, "Insufficient funds");
             return SignAndRelay(tx);
@@ -334,6 +347,11 @@ namespace Neo.Plugins
             if (fee < Fixed8.Zero)
                 throw new RpcException(-32602, "Invalid params");
             Transaction tx = Wallet.MakeTransaction(null, outputs, from: from, change_address: change_address, fee: fee);
+            if (tx.Size > 1024)
+            {
+                fee += Fixed8.FromDecimal(tx.Size * 0.00001m + 0.001m);
+                tx = Wallet.MakeTransaction(null, outputs, from: from, change_address: change_address, fee: fee);
+            }
             if (tx == null)
                 throw new RpcException(-300, "Insufficient funds");
             return SignAndRelay(tx);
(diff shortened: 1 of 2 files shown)


QingmingZi:
This will solve all cases of over 1024 transactions sent

vncoelho:
Lets be free. ;D ahauahahah

lightszero:
the main attack for translate is make many tx outputs and invokescript.
so allow inputs free is safe.

and if i want give some body some asset, the most important thing is txoutput.
whatever i use 1 input and 100 input,we should not let the 100 input is more expensive。
and if i use 100 input,i am helping to make small money join together.it is good for system.

vncoelho:
We just automatized this for https://neocompiler.io in order to automatic setting network gas deploy fee for contracts. @igormcoelho

It looks like we will node an adjustment. :D

igormcoelho:
@QingmingZi  Another solution here is to join input transactions in larger outputs... however, this assumes that the "join" transaction is below 1024 bytes. Is it the case?
Otherwise, I will have to agree with @lightszero. Removing inputs from count will likely not induce negative behaviors, as outputs will still be counted.

Will it fully resolve the calculation situation?

igormcoelho:
@jsolman I agree with you that only GAS inputs is to be ignored, but I think that perhaps it's not good to overcomplicate this, as this is already becoming complicated :)
The simpler this is now, the sooner wallets may fully adopt these changes.

erikzhang:
Gas in NEO 3.0 is no longer UTXOs.

igormcoelho:
> Gas in NEO 3.0 is no longer UTXOs.

I think this is just an emergencial change to reduce GAS costs on tx Erik, while global assets are not yet removed. Since it's an advantage to users, without exposing to any extra attacks, I think we can adopt this and pack into the newest version (as soon as this patch is finished).

QingmingZi:
@igormcoelho Yes, the original unpredictable loop calculation comes from increasing the GAS itself and causing changes in UTXO inputs, causing the tx size to change again. If UTXO inputs are not counted, it is obvious that duplicate calculations will not be triggered. Or, repeat the calculation at most once (I am not sure if it exceeds 1024 when constructing the transaction for the first time)

shargon:
I don't like it... space is space, and should have a cost

vncoelho:
@shargon, I think this solution is good, maybe it is better to duplicate the price but still remove the inputs as suggested in this PR.

igormcoelho:
> I don't like it... space is space, and should have a cost

Space will be effectively reduced @shargon (more inputs we consume, much less space system needs).

shargon:
This PR doesn't reduce the space, it discount the `Input.Size` from the size to be used to pay the fee, so you can do a TX with 62K inputs and just one output, and pay only for this one output, not the real size of the TX.

I don't like it ...

igormcoelho:
If someone does this, a lot of inputs will be effectively consumed from the network, so it's not feasible to do this many times and create an attack. But you're right in that someone could do that for free.
One possibility is to add another limit on free inputs hahhaah but gets complex and complex..

vncoelho:
Something like 20 free inputs would be enough...aheuaheuaea

But like Erik said, let's try to solve this asap for us to move forward. UTXO discussions should not really be our focus I think.

jasonliu91:
I think the problem with this strategy is that the charge for each byte is not suitable for utxo transactions, but should be a larger fee unit, and bitcoin is charged per kilobyte. If you charge per byte, the utxo of the payment will change the result, which has a loop problem and increase the difficulty of the wallet to construct the transaction. But according to the kilobyte charge, the wallet can have room to construct a legitimate transaction according to the user's utxo situation.

vncoelho:
@lightszero @igormcoelho @shargon @erikzhang @jsolman, merge this or leave NEO2 as it is?

In NEO3 this solution will not be useful.

shargon:
As i said ...
> I don't like it... space is space, and should have a cost

jasonliu91:
> As i said ...
> 
> > I don't like it... space is space, and should have a cost

I agree with your theory, but the point is how should we choose the right unit. If we are pursuing charging for all the space, isn't it better to use bit? Obviously, any unit of price should be selected according to the actual situation. And our actual situation is that byte is not suitable for UTXO transactions. If we think that the kb unit is too large, we can even consider 100bytes, but it is not needed. Because the billing starting point is in kb, we can completely unify it into a more concise strategy without causing additional security problems. For the attacker, due to the free strategy below 1kb, he can only send transactions of less than 1kb. A byte-based strategy does not solve any problems but introduces problems. Why not wait for 3.0 and then use byte as the unit?

erikzhang:
Conflict.

PR #75, opened 2019-04-03
fix size calculation

superboyiii:
fix size calculation


igormcoelho:
Perhaps this kind of logic (fee calculation) should be embedded on Wallet class itself.

superboyiii:
> Perhaps this kind of logic (fee calculation) should be embedded on Wallet class itself.

Yes, I do agree with you. I'll embedded it into wallet class when the calculation is confirmed as the final.

PR #78, opened 2019-04-07
Fixes `ImportBlocks` for Neo 2.10.1

### LINKED ISSUE #77: [2.10.1] Can't run with ImportBlocks after update

erikzhang:
Closes #77

### DIFF
--- a/ImportBlocks/ImportBlocks.cs
+++ b/ImportBlocks/ImportBlocks.cs
@@ -14,11 +14,6 @@ namespace Neo.Plugins
 {
     public class ImportBlocks : Plugin
     {
-        public ImportBlocks()
-        {
-            OnImport();
-        }
-
         private static bool CheckMaxOnImportHeight(uint currentImportBlockHeight)
         {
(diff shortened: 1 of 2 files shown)


erikzhang:
> Looks fine. I think we should still consider #66 though in order to be able to exit cleanly during import.

Yes. I created this PR not to close #66, but because v2.10.1 has been released, and now we need an import plugin available. We can continue the development of #66.

PR #79, opened 2019-04-10
Draft for ConsensusPlayground Plugin (P2PLossyConsensus)

vncoelho:
A plugin similar to this has been used during the tests for dBFT 2.0, designed by @shargon P2PPlugin and @jsolman.
This is a more configurable version.

Maybe it could be considered to be included here.
I think that NGD and other communities may use this for analyzing consensus and generating different kinds of reports.

### DIFF
--- a/ConsensusPlayground/ConsensusPlayground.cs
+++ b/ConsensusPlayground/ConsensusPlayground.cs
@@ -1,0 +1,43 @@
+﻿using Neo.Network.P2P.Payloads;
+using Neo.Consensus;
+using Neo.Network.P2P;
+using System;
+
+
+namespace Neo.Plugins
+{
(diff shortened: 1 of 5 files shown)


vncoelho:
Ok, @erikzhang, that is why I did this as a Draft. It could have been an issue, but here it would be more clear.

Let's close this. However, I believe that we need a place for sharing such plugins and ideas, otherwise, each community and team will need to look at others repo and get ideas like this.
This plugin, for example, is something that many may use.

Another day we was discussing about this Neo-Plugins repo.
I thought about this you are mentioning, about many plugins here.

However, we believe that it will probably reach a limit.
In the future, it will probably not be more than 20 plugins, for example. It will reach a point that different parts of the code are already covered and we could just improve the plugins with more configurations.

PR #81, opened 2019-04-15
Prevent DoS with wrong files

shargon:
According to this line, we can't have any block more than this constant, so i think that is a good practice to avoid any buffer bigger than this value
https://github.com/neo-project/neo/blob/3a06f9c0efcc3cc267d380bb945fe7474b1f2243/neo/Network/P2P/Message.cs#L12


igormcoelho:
Fully agreed Shargon!
Do you think it's better to launch an exception in case it's bigger? otherwise, how will we know the exact cause? Bypassing this "error" with a simple break (ignoring block) seems strange to me.

shargon:
If we throw an exception in this point, will be captured and showed?

vncoelho:
I think so, @shargon, maybe it is better to launch an exception, stop and notify.

shargon:
@vncoelho could you tested if the exception is successfully captured?

vncoelho:
Sure, my friend. I will need a couple of days. Until next Tuesday perhaps, can be? It is holiday here. :dancer:

vncoelho:
@shargon, I think that maybe we will wait for https://github.com/neo-project/neo-cli/pull/357.

Do you want to merge this for master-2.x as well?

shargon:
For both

shargon:
https://github.com/neo-project/neo-cli/pull/357 is done @vncoelho

vncoelho:
@shargon, I think that we should have neocli 3x running before merging this. Lets remember to test it when the time comes.

PR #86, opened 2019-04-22
Fix output json for multisigaddress transaction

superboyiii:
Fix output json for multisigaddress transaction

### DIFF
--- a/RpcWallet/RpcWallet.cs
+++ b/RpcWallet/RpcWallet.cs
@@ -313,29 +313,36 @@ namespace Neo.Plugins
 
             ContractParametersContext transContext = new ContractParametersContext(tx);
             Wallet.Sign(transContext);
-            tx.Witnesses = transContext.GetWitnesses();
-            if (tx.Size > 1024)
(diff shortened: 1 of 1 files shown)


RpcWallet/RpcWallet.cs
```diff
@@ -313,29 +313,36 @@ private JObject SendFrom(UIntBase assetId, UInt160 from, UInt160 to, string valu
...
             ContractParametersContext transContext = new ContractParametersContext(tx);
             Wallet.Sign(transContext);
-            tx.Witnesses = transContext.GetWitnesses();
-            if (tx.Size > 1024)
+            if (transContext.Completed)
             {
-                Fixed8 calFee = Fixed8.FromDecimal(tx.Size * 0.00001m + 0.001m);
-                if (fee < calFee)
+                tx.Witnesses = transContext.GetWitnesses();
+                if (tx.Size > 1024)
```
vncoelho: Erik, when I see this hard codded parameter I come back with that discussion about the Simple Policy as a native contract.

PR #88, opened 2019-04-23
Fix the height to end with

superboyiii:
Fix the height to end with

### DIFF
--- a/CoreMetrics/CoreMetrics.cs
+++ b/CoreMetrics/CoreMetrics.cs
@@ -75,7 +75,7 @@ namespace Neo.Plugins
 
             JArray array = new JArray();
             uint heightToBegin = lastHeight > 0 ? lastHeight - nBlocks : Blockchain.Singleton.Height - nBlocks;
-            for (uint i = heightToBegin; i <= Blockchain.Singleton.HeaderHeight; i++)
+            for (uint i = heightToBegin; i <= heightToBegin + nBlocks; i++)
             {
                 JObject json = new JObject();
                 Header header = Blockchain.Singleton.Store.GetHeader(i);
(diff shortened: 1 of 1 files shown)


CoreMetrics/CoreMetrics.cs
```diff
@@ -75,7 +75,7 @@ private JObject GetBlocksTime(uint nBlocks, uint lastHeight)
 
             JArray array = new JArray();
             uint heightToBegin = lastHeight > 0 ? lastHeight - nBlocks : Blockchain.Singleton.Height - nBlocks;
-            for (uint i = heightToBegin; i <= Blockchain.Singleton.HeaderHeight; i++)
+            for (uint i = heightToBegin; i <= heightToBegin + nBlocks; i++)
             {
                 JObject json = new JObject();
                 Header header = Blockchain.Singleton.Store.GetHeader(i);
```
shargon: We must to check if is null

vncoelho:
Nice catch, @superboyiii, thanks.
There was a minor problem when we requested the number of blocks equal to the last height, it was not returning the Genesis Block.

CoreMetrics/CoreMetrics.cs
vncoelho: You mean `Blockchain.Singleton.Height`?

superboyiii:
Awesome fix 😝. Agree with you.

superboyiii:
No problem.😀 I'll test it tommorrow.

vncoelho:
:D Tested! Thanks, @superboyiii.

CoreMetrics/CoreMetrics.cs
shargon: The header

vncoelho: That is true.
I do know when it will be `null` in practice, because we are checking the limit `Blockchain.Singleton.HeaderHeight`.
But looks like to be a protection. Let's add it in another PR if it really can happen.

PR #90, opened 2019-04-29
Feat/export import tx final application state

vncoelho:
Deals with https://github.com/neo-project/neo/issues/712 during export and import only.

Under development, this is the initial draft for export, not yet fully tested.

### DIFF
--- a/ImportBlocks/ImportBlocks.cs
+++ b/ImportBlocks/ImportBlocks.cs
@@ -3,6 +3,7 @@ using Neo.IO;
 using Neo.Ledger;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
+using Neo.IO.Json;
 using System;
 using System.IO;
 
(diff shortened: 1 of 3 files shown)


ImportBlocks/ImportBlocks.cs
```diff
@@ -94,6 +118,46 @@ private bool OnExport(string[] args)
...
+                _params.Add(txHash);
+                JObject getAppLogRequest = new JObject();
+                getAppLogRequest["id"] = -1;
+                getAppLogRequest["method"] = "getapplicationlog";
+                getAppLogRequest["params"] = _params;
+                JObject txState = System.RpcServer.ProcessRequest(null, getAppLogRequest);
+
+                Console.Write($"Result is {txState["result"]}\n");
+                if (!txState["result"])
+                    if ((int)txState["code"].AsNumber() == -100 || (int)txState["code"].AsNumber() == -32602 || (int)txState["code"].AsNumber() == -32600)
```
vncoelho: @erikzhang, @shargon, it would be nice to change this line for any kind of error.

vncoelho:
@erikzhang, what solution do you propose?

erikzhang:
My solution is to manually copy the ApplicationLogs directory, and then the `ImportBlocks` plugin directly reads the json files.

vncoelho:
We thought about that, @erikzhang, the current Application Logs folder for Mainnet at height 3687369 is 3.9G. 
What do you think, @igormcoelho?

That would be the easiest way without making changes.
Furthermore, with the MPT this might be unnecessary, right?

vncoelho:
I think there is not reason to push this for 2.x anymore.

In addition, for 3.0 we gonna have better solutions.

PR #91, opened 2019-04-29
Minor fix on storage state dumper parameter HeightToStartRealTimeSyncing

vncoelho:


StatesDumper/StatesDumper.cs
```diff
@@ -131,7 +131,7 @@ public void OnCommitStorage(Snapshot snapshot)
             uint blockIndex = snapshot.Height;
             if (bs_cache.Count > 0)
             {
-                if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (blockIndex > Settings.Default.HeightToStartRealTimeSyncing))
+                if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (Settings.Default.HeightToStartRealTimeSyncing != -1 && blockIndex > Settings.Default.HeightToStartRealTimeSyncing))
```
ixje:
```suggestion
                if ((blockIndex % Settings.Default.BlockCacheSize == 0) || (Settings.Default.HeightToStartRealTimeSyncing != -1 && blockIndex >= Settings.Default.HeightToStartRealTimeSyncing))
```
`HeightToStartRealTimeSyncing` to me sounds like it should be inclusive.  This should allow you to configure block import at a certain height + set `HeightToStartRealTimeSyncing` to the same height and get all storages up to and including that block. Right now you can't.

vncoelho: There is a command that allows you to dump the whole storage at the current height:

https://github.com/neo-project/neo-plugins/blob/fb584f5fb099fc31899d968e7e1eb66b5a5e5fee/StatesDumper/StatesDumper.cs#L52-L65

Thanks for the suggestion to be >=.

PR #92, opened 2019-05-01
Prevent exceed the height

shargon:
We think that it could be exceeded the max height using the `lastHeight` parameter


shargon:
It depend if you put a wrong value in https://github.com/neo-project/neo-plugins/blob/5da9ac5a540afbdfb956c4205bbec86666ca4828/CoreMetrics/CoreMetrics.cs#L27

vncoelho:
Let's fix the null case, @shargon. What do you think?
Fix the null and keep the logic.

PR #94, opened 2019-05-16
Minor optimization on Block Timestamps RPC call

shargon:


shargon:
Optimization, if break because null, why do you need to create this variable?

PR #97, opened 2019-05-18
Merkle Patricia Tries (2x)

rodoufu:
An experimental plugin for persisting neo blockchain states into Merkle Patricia Tries.

### DIFF
--- a/MPTStates/MPT/MerklePatriciaNode.cs
+++ b/MPTStates/MPT/MerklePatriciaNode.cs
@@ -1,0 +1,312 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using Neo.Cryptography;
+using Neo.IO;
+using Neo.Ledger;
+
+namespace Neo.Plugins.MPT
+{
+    /// <inheritdoc />
+    /// <summary>
+    /// Modified Merkel Patricia Node.
+    /// Note: It is not a thread safe implementation.
+    /// </summary>
+    public class MerklePatriciaNode : StateBase,
+        ICloneable<MerklePatriciaNode>,
+        IEquatable<MerklePatriciaNode>,
+        IEnumerable<byte[]>
+    {
+        private const int BranchSize = 18;
+        private const int ExtensionSize = 2;
+        private const int LeafSize = 3;
+
+        private byte[][] _hashes;
+
+        public MerklePatriciaNode() : this(0)
+        {
+        }
+
+        public MerklePatriciaNode(int size) => _hashes = new byte[size][];
+
+        /// <summary>
+        /// Indicates if the node is a branch.
+        /// </summary>
+        public bool IsBranch => _hashes.Length == BranchSize;
+
+        /// <summary>
+        /// Indicates if the node is an extension.
+        /// </summary>
+        public bool IsExtension => _hashes.Length == ExtensionSize;
+
+        /// <summary>
+        /// Indicates if the node is a leaf.
+        /// </summary>
+        public bool IsLeaf => _hashes.Length == LeafSize;
+
+        /// <summary>
+        /// Get and set the hashes by index.
+        /// </summary>
+        /// <param name="index">Index of the hash to get or set.</param>
+        public byte[] this[int index]
+        {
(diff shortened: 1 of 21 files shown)


igormcoelho:
Very good Rodolfo! Let's advance with this plugin. The target is still neo-project:master, could you change it for neo-project:master-2.x ?

vncoelho:
Great job, @rodoufu.
Lets create this branch here, Igor.

erikzhang:
Is this the same as https://github.com/neo-project/neo/pull/528?

igormcoelho:
Right now it's the same Erik, but it will change... this one here will be more experimental, so that we validate it and then port it to neo project.

rodoufu:
> Very good Rodolfo! Let's advance with this plugin. The target is still neo-project:master, could you change it for neo-project:master-2.x ?
@igormcoelho , it looks like there's no mater-2x branch on the neo-plugins project:
https://github.com/neo-project/neo-plugins/branches

Only on the core neo project:
https://github.com/neo-project/neo/branches

Am I missing something?

rodoufu:
Now I have changed the target.

erikzhang:
I think this plugin is for 2.x, because 2.x has no built-in MPT.

MPTStates/MPT/MerklePatriciaNode.cs
```diff
@@ -0,0 +1,312 @@
...
+        }
+
+        /// <summary>
+        /// Get and set the key of the node.
+        /// Used for leaf and branch nodes.
+        /// </summary>
+        public byte[] Key
+        {
+            get => _hashes[_hashes.Length - 2];
+            set => _hashes[_hashes.Length - 2] = value;
```
shargon: Throw an exception on empty tree

MPTStates/MPT/MerklePatriciaNode.cs
```diff
@@ -0,0 +1,312 @@
...
+                    _hashes[i] = IsBranch && _hashes[i].Length == 0 ? null : _hashes[i];
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public override void Serialize(BinaryWriter writer)
+        {
+            base.Serialize(writer);
+            writer.Write((byte) _hashes.Length);
```
shargon: is not possible to be bigger than 256?

rodoufu:
@shargon , I think the idea is to have the MPT on Neo version 3.0 but start it as a plugin in the version 2.x so we can use it and improve before going to the Neo 3.0

MPTStates/MPT/MerklePatriciaNode.cs
igormcoelho: I think node types are verified before usage.. IsNull for example, has no Key.

igormcoelho:
@shargon 
A node can be of four kinds:
- Null (size 0)
- Hash (size 1) (I don't know if Rodolfo considered this as a node... but I did on mine)
- Path (size 2) it can be Leaf or Extension. Rodolfo added the Key to Leaf, to avoid colisions, I think we can do that implicitly Rodolfo, without hardcoding the key (so size is effectly 2 and no space is wasted twice)
-Branch (size 17) I used variable sizes from 3 to 17  on my implementation (3 is Branch2 + Leaf, and 17 is classic Branch16 + Leaf). This slightly compresses the tree.

@rodoufu 
As soon as you're back, we need to implement Sparse Merkle Tree, which is a binary tree, and allows much deeper compression techniques. I believe space will be further reduced than MPT.
We can create another plugin with that other tree and compare both, so Neo 3.0 can take a clear decision.
MPT and SMT solve the same problem: validated key-value storage with inclusion and non-inclusion proofs.

MPTStates/MPT/MerklePatriciaNode.cs
igormcoelho: It's possible, just an effect of using SHA256. In fact, we should use Hash256, to avoid length attacks.

vncoelho:
@rodoufu, this has been a great effort. Thanks for all your motivation, studies and prototype on this.
NeoX has already been merged to master2x with another implementation.
In this sense, I will close this one.

However, I would to state my real appreciation for the work you have conducted.

PR #100, opened 2019-05-30
Change the way how to pay about >1024 transaction. (2x)

lightszero:
#neo-project/neo-cli#303

pay for every extrablock instread  every bytes

in default extrablock=1024 bytes.
feeperextrablock=0.0124.gas

example:
txsize  < 1024, pay 0 block,free
txsize =1025  ~  2048 bytes, pay 1 block    0.01024gas
txsize =2049  ~  3072 bytes, pay 2 block    0.02048gas

if you want pay by every extrabytes
set extrablock=1 bytes
feeperextrablock =0.0001gas

### DIFF
--- a/SimplePolicy/SimplePolicy/config.json
+++ b/SimplePolicy/SimplePolicy/config.json
@@ -2,6 +2,7 @@
   "PluginConfiguration": {
     "MaxTransactionsPerBlock": 500,
     "MaxFreeTransactionsPerBlock": 20,
+    "FeePerExtraByte": 0.00001,
     "MaxFreeTransactionSize": 1024,
     "FeePerExtraBlock": 0.01024,
     "ExtraBlockSize": 1024,
--- a/SimplePolicy/SimplePolicy/config.json
+++ b/SimplePolicy/SimplePolicy/config.json
@@ -3,7 +3,8 @@
     "MaxTransactionsPerBlock": 500,
     "MaxFreeTransactionsPerBlock": 20,
     "MaxFreeTransactionSize": 1024,
-    "FeePerExtraByte": 0.00001,
+    "FeePerExtraBlock": 0.01024,
+    "ExtraBlockSize": 1024,
     "HighPriorityTxType":
     [
         "ClaimTransaction"
--- a/SimplePolicy.UnitTests/UT_SimplePolicy.cs
+++ b/SimplePolicy.UnitTests/UT_SimplePolicy.cs
@@ -291,87 +291,87 @@ namespace SimplePolicy.UnitTests
         }
 
          [TestMethod]
-        public void TestVerifySizeLimits_FreeLessEq1024()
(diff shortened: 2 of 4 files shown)


igormcoelho:
@lightszero could we accept these tests here first? https://github.com/neo-project/neo-plugins/pull/101

Then we update the tests on this branch to see how the new prices fit.

vncoelho:
Accepted and merged, those tests are great, @igormcoelho.

vncoelho:
@lightszero, what are the advantages of this? Are you afraid of any attack?
Are you planning this for both 2x and 3.0+?

igormcoelho:
@lightszero I added your proposed option as "standard" (VerifySizeLimits2) and kept old one too (VerifySizeLimits1). Now we can discuss and create more tests for your proposal.

igormcoelho:
@shargon could you fix the indent here for us? poor editor here hahaha

jasonliu91:
This policy is for 2x ,use kb as fee unit can avoid the utxo loop calculation problem of the previous byte unit policy.

igormcoelho:
I think block calculation is a good idea. But since old formula has been around for a while , we should propose another obe which is compatible.. with lower or same prices.

igormcoelho:
@lightszero what about charging at every 512 bytes, and use FLOOR instead of CEILING? This reduces the UTXO combinations and effectively keep prices nearly the same or less (at small intervals).
`fee = floor((txsize - 1024)/512)*pricePer512Chunk + basePrice`

lightszero:
> @lightszero, what are the advantages of this? Are you afraid of any attack?
> Are you planning this for both 2x and 3.0+?

because they donot like #73 

so this is another try

SimplePolicy/SimplePolicy/config.json
```diff
@@ -2,8 +2,10 @@
   "PluginConfiguration": {
     "MaxTransactionsPerBlock": 500,
     "MaxFreeTransactionsPerBlock": 20,
-    "MaxFreeTransactionSize": 1024,
     "FeePerExtraByte": 0.00001,
```
vncoelho: It is a little bit strange to still keep `FeePerExtraByte` here.

However, the old `VerifySizeLimits1` limit was kept.

shargon:
Is for neo2, i think that we shouldn't change the economic model in neo 2

vncoelho:
I also think so, @shargon, maybe let's close this one and think about adjusting it for NEO3.
What do you think @lightszero and @erikzhang ?

lock9:
Hello @neo-project/core, what should we do with this?

vncoelho:
@lightszero, I think we can close this, right?
There is no more need of changes on NEO2x.
In my opinion, it was a good change but it was not merged in time.

However, fell free to reopen.

vncoelho:
Reopened due to https://github.com/neo-project/neo/pull/1763, if we are going to charge for tx size better to charge this way.

PR #102, opened 2019-05-31
Simplified fee tests

igormcoelho:
Simplified High Priority tx fee tests according to a simple formula: `fee = (tx_size - 1024) * 0.00001 + 0.001`. Below 1024 bytes, free are still free, and high priority require 0.001 (at least).

### DIFF
--- a/SimplePolicy.UnitTests/UT_SimplePolicy.cs
+++ b/SimplePolicy.UnitTests/UT_SimplePolicy.cs
@@ -291,9 +291,9 @@ namespace SimplePolicy.UnitTests
         }
 
          [TestMethod]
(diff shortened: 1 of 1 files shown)


igormcoelho:
Created now only five simple scenarios:
- free tx pass: below 1024 bytes
- free tx fails: after 1024 bytes
- priority tx pass: with 1024 bytes and 0.001 fee
- priority tx fails: with size 4000 and no extra fees
- priority tx pass: with size 4000 and fee = (4000-1024)*0.00001 + 0.001

vncoelho:
free tx pass: below **or equal** 1024 bytes?

igormcoelho:
I guess I'll update this to include complicated tests too... because they have been quite useful in debates.

PR #110, opened 2019-07-31
Removing SimplePolicy plugin

### LINKED ISSUE #109: "IsLowPriority"  need to be removed in SimplePolicy of NEO3
"IsLowPriority"  need to be removed in SimplePolicy since Transcation.get_IsLowPrioriy() has been removed.

image

vncoelho:
Closes #109 

@erikzhang, @shargon, @igormcoelho, I just left the `void ILogPlugin.Log(string source, LogLevel level, string message)`

I think that we should also rename this class from `SimplePolicy` -> `Loggers`, something like that.
Many people were used to have `SimplePolicy` as default and a requirement. In this sense, it will be good to change the name for avoiding confusion.


shargon:
I think that we should use our own branch

SimplePolicy/SimplePolicyPlugin.cs
```diff
@@ -1,83 +1,14 @@
...
-                .ThenBy(p => p.Hash)
-                .Take(Settings.Default.MaxTransactionsPerBlock - free.Length - 1)
-                .ToArray();
-
-            return non_free.Concat(free);
-        }
-
         void ILogPlugin.Log(string source, LogLevel level, string message)
         {
             if (source != nameof(ConsensusService)) return;
```
shargon: this should be moved to ApplicationLogs

.travis.yml
```diff
@@ -1,19 +0,0 @@
...
- - osx
-
-dist: bionic
-osx_image: xcode9.1
-
-mono: none
-dotnet: 2.1.300
-
-before_install:
-  - cd SimplePolicy.UnitTests
```
shargon: We should have a travis, just for compilation, and in the future make test for plugins too, but i think that we need it

vncoelho:
@shargon, I prefer to use neo branch, I think it is easy for contributions.

Nowadays, there are not many branches in the project. We just need to remember to clean it.
But if that becomes a requirement I do not mind in using my own repo. But nowadays I prefer using here, looks simpler to me and I do not need to give permissions for other devs to edit.

shargon:
For me is worst, i use gitkraken and i have no rights with neo branch, in neo-sharp we worked with our own branch and i think that was a good practice

erikzhang:
We can just remove the plugin and move logger to neo core.

erikzhang:
We can just remove the SimplePolicy plugin and move the logger to ApplicationLogs.

.travis.yml
vncoelho: I think it is not needed it there are no tests.
Sometime ago I did a template for each plugin and it was rejected.

vncoelho: But it looks strange to have the yellow simbol of travis forever. Maybe we can leave the file as you suggested.

erikzhang: Let's leave the file.

SimplePolicy/SimplePolicyPlugin.cs
erikzhang: Maybe we should create a new plugin to categorize and save all the logs, not just the consensus logs. I think ApplicationLogs is not used to do these logging, it is only used to record the results of the Application execution.

.travis.yml
vncoelho: Re-added without scripts

erikzhang: We can add `dotnet restore` to make it build automatically.

vncoelho: Done

erikzhang: Travis failed.

You can revert the changes to the `.travis.yml` and remove the line `dotnet test` only.

vncoelho: It is ok, we need to remove the `before_install` as well.
I think that I forgot to clean the .sln

erikzhang: You should clean `SimplePolicy.UnitTests` from .sln too. Why don't you use Visual Studio?

vncoelho: AEHuaheuahea, until nowadays using `gedit` or editing here online.
I need to configure visual studio code. I will do that as soon as possible.

PR #111, opened 2019-08-02
CI00151

erikzhang:
Close https://github.com/neo-project/neo-plugins/issues/113

### DIFF
--- a/ApplicationLogs/LogReader.cs
+++ b/ApplicationLogs/LogReader.cs
@@ -78,7 +78,7 @@ namespace Neo.Plugins
                     }
                     return notification;
                 }).ToArray();
-                writeBatch.Put((appExec.Transaction.Hash ?? snapshot.PersistingBlock.Hash).ToArray(), json.ToString());
+                writeBatch.Put((appExec.Transaction?.Hash ?? snapshot.PersistingBlock.Hash).ToArray(), json.ToString());
             }
             db.Write(WriteOptions.Default, writeBatch);
         }
--- a/ApplicationLogs/LogReader.cs
+++ b/ApplicationLogs/LogReader.cs
@@ -52,7 +52,7 @@ namespace Neo.Plugins
             foreach (var appExec in applicationExecutedList)
             {
                 JObject json = new JObject();
-                json["txid"] = appExec.Transaction.Hash.ToString();
+                json["txid"] = appExec.Transaction?.Hash.ToString();
                 json["trigger"] = appExec.Trigger;
                 json["vmstate"] = appExec.VMState;
                 json["gas_consumed"] = appExec.GasConsumed.ToString();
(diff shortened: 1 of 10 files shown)


RpcWallet/RpcWallet.cs
```diff
@@ -82,17 +77,15 @@ public JObject OnProcess(HttpContext context, string method, JArray _params)
                             from = _params[0].AsString().ToScriptHash();
                             to_start = 1;
                         }
-                        JArray to = (JArray)_params[to_start + 0];
-                        long fee = _params.Count >= to_start + 2 ? long.Parse(_params[to_start + 1].AsString()) : 0;
```
shargon: We should be able to attach ScopedSignatures, but maybe in other pull request

vncoelho: Let's do in another PR, scoped was not merged, right?

ApplicationLogs/LogReader.cs
```diff
@@ -78,7 +78,7 @@ public void OnPersist(Snapshot snapshot, IReadOnlyList<Blockchain.ApplicationExe
                     }
                     return notification;
                 }).ToArray();
-                writeBatch.Put(appExec.Transaction.Hash.ToArray(), json.ToString());
+                writeBatch.Put((appExec.Transaction.Hash ?? snapshot.PersistingBlock.Hash).ToArray(), json.ToString());
```
shargon: Is not the same expected value

erikzhang: Do you have any good ideas?

shargon: In fact, is stored, but is not readed, but one extra property in the json should be great, something like 
```csharp
json["type"]=appExec.Transaction==null?"tx":"block";
```

ApplicationLogs/LogReader.cs
```diff
@@ -52,7 +52,7 @@ public void OnPersist(Snapshot snapshot, IReadOnlyList<Blockchain.ApplicationExe
             foreach (var appExec in applicationExecutedList)
             {
                 JObject json = new JObject();
-                json["txid"] = appExec.Transaction.Hash.ToString();
+                json["txid"] = appExec.Transaction?.Hash.ToString();
```
shargon:
```suggestion
                json["type"] = appExec.Transaction == null ? "block" : "tx";
                json["txid"] = appExec.Transaction?.Hash.ToString();
```

erikzhang: We can check the `trigger` for knowing the type.

shargon: The unique without Tx is System, right?

erikzhang: Yes.

PR #112, opened 2019-08-02
Log plugin

shargon:
Related to https://github.com/neo-project/neo/pull/980/files


erikzhang:
No. We should create a new plugin. "ApplicationLog` is for logging transaction execution results.

shargon:
Maybe we should rename `ApplicationLog`, the expected behaviour is for logs 😆

igormcoelho:
Let's create a new plugin, `SystemLog` plugin, and use for this. ApplicationLog is less logging and more application data processing itself.

shargon:
if we decide the name, i can make the changes

PR #117, opened 2019-08-12
SystemLog plugin

shargon:
Improve of https://github.com/neo-project/neo-plugins/pull/112
Related to https://github.com/neo-project/neo/pull/980

### DIFF
--- a/SystemLog/ConsoleColorSet.cs
+++ b/SystemLog/ConsoleColorSet.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 
-namespace SystemLog
+namespace Neo.Plugins
 {
     internal class ConsoleColorSet
     {
--- a/SystemLog/ConsoleColorSet.cs
+++ b/SystemLog/ConsoleColorSet.cs
@@ -1,0 +1,51 @@
+﻿using System;
+
+namespace SystemLog
+{
+    internal class ConsoleColorSet
+    {
+        #region Constants
+
+        public static readonly ConsoleColorSet Debug = new ConsoleColorSet(ConsoleColor.Cyan);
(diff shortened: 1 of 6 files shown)


erikzhang:
Also need to save the logs to file.

SystemLog/ConsoleColorSet.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    internal class ConsoleColorSet
+    {
+        #region Constants
+
+        public static readonly ConsoleColorSet Debug = new ConsoleColorSet(ConsoleColor.Cyan);
+        public static readonly ConsoleColorSet Info = new ConsoleColorSet(ConsoleColor.White);
+        public static readonly ConsoleColorSet Warning = new ConsoleColorSet(ConsoleColor.Yellow);
+        public static readonly ConsoleColorSet Error = new ConsoleColorSet(ConsoleColor.Red);
+        public static readonly ConsoleColorSet Fatal = new ConsoleColorSet(ConsoleColor.Red);
```
vncoelho: Fatal? ajajja

SystemLog/Logger.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public void Log(string source, LogLevel level, string message)
+        {
+            var log = $"[{DateTime.Now.TimeOfDay:hh\\:mm\\:ss\\.fff}] {message}";
+
+            if (Settings.Default.ConsoleOutput)
+            {
+                var currentColor = new ConsoleColorSet();
+
+                switch (level)
```
vncoelho: There is no other way, right @shargon? It is strange to set this color all the time.
With this we could also debug different classes with different levels, yes? Because a cache for each part of the code could be more efficiently, however...if we are printing efficiency is not a focus anyway.

A major configuration could easily switch the logs for different parts of the codes.

PR #125, opened 2019-08-19
Fix RpcNep5Tracker

### LINKED ISSUE #116: Update RpcNep5Tracker to work in NEO3
This plugin depends on UTXO, it must be refactored to work with NEO 3

From @shargon:
https://github.com/neo-project/neo-plugins/issues/114#issuecomment-518576175

erikzhang:
Fixes #116

### DIFF
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -200,7 +200,7 @@ namespace Neo.Plugins
             return true;
         }
 
-        private void AddTransfers(byte dbPrefix, UInt160 userScriptHash, uint startTime, uint endTime,
+        private void AddTransfers(byte dbPrefix, UInt160 userScriptHash, ulong startTime, ulong endTime,
             JArray parentJArray)
         {
             var prefix = new[] { dbPrefix }.Concat(userScriptHash.ToArray()).ToArray();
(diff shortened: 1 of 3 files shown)


RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -67,10 +67,14 @@ private void ResetBatch()
         private void RecordTransferHistory(Snapshot snapshot, UInt160 scriptHash, UInt160 from, UInt160 to, BigInteger amount, UInt256 txHash, ref ushort transferIndex)
         {
             if (!_shouldTrackHistory) return;
+
+            Header header = snapshot.Height == 0
```
vncoelho: Very good.

shargon:
https://github.com/neo-project/neo-plugins/pull/125/files#diff-2c8a32803157f91fab2b7beedc206bb9R108 Now we call the `Transfer` event, no `MintTokens`, we should preserve it?

superboyiii:
```getnep5balances``` works now. ```getnep5transfers``` still need to fix.
image

**Comment by erikzhang** [comment]
`getnep5transfers` only returns transactions in 7 days by default. You need to add parameters `startTime` and `endTime`. To get all transactions, you can just send the parameters:

```json
"params": ["AbT3X8KusycQKZiZBQhnHCuWFo1fpPuXWX", 0]
```

superboyiii:
@erikzhang  I know this, but I make these transactions in two days which should display( in 7 days). It still doesn't work. For example, I use 2017/01/01 to be the start timestamp.
image

superboyiii:
Works well now. Good job! 
image

PR #128, opened 2019-08-22
Fixes deadlock in RpcNep5Tracker

### LINKED ISSUE #127: Two nodes aren't able to run in one PC at the same time if they all have installed RpcNep5Tracker.
Two nodes aren't able to run in one PC at the same time if they all have installed RpcNep5Tracker, 
even though there's no port conflict. If one of them uninstalls this plugin, they all work well.

erikzhang:
Fixes #127


superboyiii:
Now neo3 works. Maybe we should apply it on neo2.

PR #131, opened 2019-08-26
Fix RpcNep5Tracker for uncollectible chain snapshot & ready for plugin-v2.10.3 release

superboyiii:
Fix RpcNep5Tracker for uncollectible chain snapshot & ready for plugin-v2.10.3 release.

### DIFF
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -12,6 +12,7 @@ using Neo.VM;
 using Neo.Wallets;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Numerics;
 using System.Text;
@@ -40,7 +41,7 @@ namespace Neo.Plugins
             if (_db == null)
             {
                 var dbPath = GetConfiguration().GetSection("DBPath").Value ?? "Nep5BalanceData";
-                _db = DB.Open(dbPath, new Options { CreateIfMissing = true });
+                _db = DB.Open(Path.GetFullPath(dbPath), new Options { CreateIfMissing = true });
             }
             _shouldTrackHistory = (GetConfiguration().GetSection("TrackHistory").Value ?? true.ToString()) != false.ToString();
             _recordNullAddressHistory = (GetConfiguration().GetSection("RecordNullAddressHistory").Value ?? false.ToString()) != false.ToString();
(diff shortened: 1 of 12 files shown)


RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -67,10 +68,10 @@ private void ResetBatch()
         private void RecordTransferHistory(Snapshot snapshot, UInt160 scriptHash, UInt160 from, UInt160 to, BigInteger amount, UInt256 txHash, ref ushort transferIndex)
         {
             if (!_shouldTrackHistory) return;
+            Header header = snapshot.GetHeader(snapshot.Height);
```
erikzhang: Need to fix the deadlock.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -191,7 +191,7 @@ public void OnPersist(Snapshot snapshot, IReadOnlyList<Blockchain.ApplicationExe
                     script = sb.ToArray();
                 }
 
-                ApplicationEngine engine = ApplicationEngine.Run(script, snapshot);
+                ApplicationEngine engine = ApplicationEngine.Run(script, snapshot,testMode:true);
```
erikzhang: Don't use `testMode` because it can't prevent infinite loop attacks.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -40,7 +41,7 @@ public override void Configure()
             if (_db == null)
             {
                 var dbPath = GetConfiguration().GetSection("DBPath").Value ?? "Nep5BalanceData";
-                _db = DB.Open(dbPath, new Options { CreateIfMissing = true });
+                _db = DB.Open(Path.GetFullPath(dbPath), new Options { CreateIfMissing = true });
```
shargon: Is this well rebased? i think that this change comes from another PR

RpcNep5Tracker/RpcNep5Tracker.cs
superboyiii: Got it.

RpcNep5Tracker/RpcNep5Tracker.cs
superboyiii: Yes, I sync it from NEO3's PR. When two nodes installed RpcNep5Tracker, they aren't able to run together on the same PC because they're using the same dbPath (relative path). This PR is able to solve this issue . I've tested it on my local.

RpcNep5Tracker/RpcNep5Tracker.cs
erikzhang: NOOOOO... Don't just remove it. You need to add a few fees.

erikzhang: Please follow my original PR. #125

superboyiii: Done.

SimplePolicy.UnitTests/SimplePolicy.UnitTests.csproj
```diff
@@ -19,6 +19,7 @@
     <PackageReference Include="MSTest.TestAdapter" Version="1.4.0" />
     <PackageReference Include="MSTest.TestFramework" Version="1.4.0" />
     <PackageReference Include="Moq" Version="4.7.63" />
+    <PackageReference Include="Neo" Version="2.10.3" />
```
erikzhang: We don't need this line, I think.

superboyiii: Yes, a mistake. Removed.

shargon:
I checked with the neo3, and this is for neo2, @erikzhang could you revert my last pull?

erikzhang:
Reverted.

PR #132, opened 2019-08-28
Fixes output of `getbalance`

erikzhang:
See https://github.com/neo-project/neo/issues/1069


superboyiii:
Nice fix, much unified now.
image

PR #133, opened 2019-08-29
Fix RpcSystemAssetTracker

superboyiii:
Fix RpcSystemAssetTracker for multi nodes running on one PC.

### DIFF
--- a/RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
+++ b/RpcSystemAssetTracker/RpcSystemAssetTrackerPlugin.cs
@@ -11,6 +11,7 @@ using System.Linq;
 using Neo.Ledger;
 using Neo.Persistence;
 using Snapshot = Neo.Persistence.Snapshot;
+using System.IO;
 
 namespace Neo.Plugins
 {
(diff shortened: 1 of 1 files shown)


superboyiii:
Tested.
image

vncoelho:
Hi @superboyiii, I think I did not see this before because usually we run on indivual containers when settling privatenets.

I did not understand the problem very well...aehuahea
But I believe that your fix are always good!
If you could guide me the procedure for having this problem I can try to reproduce and test.

superboyiii:
> Hi @superboyiii, I think I did not see this before because usually we run on indivual containers when settling privatenets.
> 
> I did not understand the problem very well...aehuahea
> But I believe that your fix are always good!
> If you could guide me the procedure for having this problem I can try to reproduce and test.

Hi @vncoelho  In the before case, if we installed RpcSystemAssetTracker plugin in two nodes and start them together, there'll be a conflict between these nodes since they're using the same relative dbPath. The second node wasn't able to start. If turn it to full Path, there'll never be any conflict. This situation can always be seen on running multi nodes on one PC.

vncoelho:
@superboyiii with superpower's...good to see you merging this! 
As I said, since we were always running with docker we did not see this problem. It is good that you found.

superboyiii:
@vncoelho Ahaahaaha.....Thanks.

PR #135, opened 2019-09-11
Fix/memory

### LINKED ISSUE #129: Export block makes extremely high cost of memory which usually cause system killed on Linux.
Export block makes extremely high cost of memory which usually cause system killed on Linux.
On mainnet export, our server which has 16G memory, is only able to hold on  about 2 million blocks' export. After that, it's been killed by ubuntu. Here's the line graph. I think ImportBlocks plugin is necessary to improve algorithm for export.
image

KickSeason:
fix #129


vncoelho:
Looks good idea.

KickSeason:
@shargon  I will.

ImportBlocks/ImportBlocks.cs
```diff
@@ -48,8 +48,9 @@ private bool OnExport(string[] args)
                     if (start <= end)
                         fs.Write(BitConverter.GetBytes(count), 0, sizeof(uint));
                     fs.Seek(0, SeekOrigin.End);
-                    using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
-                        for (uint i = start; i <= end; i++)
+                    for (uint i = start; i <= end; i++)
+                    {
+                        using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
```
erikzhang: I think that create new snapshot each time is not a good idea. We can use `Blockchain.Singleton.Store.GetBlock()`. It won't create any snapshot and read block from inner database directly.

shargon: Maybe use the `store` directly in order to prevent the cache? or something for disable the cache of the objects in the snapshots?

shargon: Yes, is better directly with the `Store`

erikzhang: It won't increase the memory, because they will be released by GC. You can test it.

shargon: But is stored in the internal `dictionary`, so we will have the same problem, we need 
```csharp
public TValue TryGet(TKey key, bool cache=true)
```

KickSeason: So does `[]` operation,

ImportBlocks/ImportBlocks.cs
```diff
@@ -42,57 +33,65 @@ private bool OnExport(string[] args)
...
+        }
+
+        private bool OnExport(string[] args)
+        {
+            bool writeStart;
+            uint start, count;
+            string path;
+            if (args.Length < 2) return false;
+            if (!string.Equals(args[1], "block", StringComparison.OrdinalIgnoreCase)
+                && !string.Equals(args[1], "blocks", StringComparison.OrdinalIgnoreCase))
```
shargon: seems duplicated

ImportBlocks/ImportBlocks.cs
erikzhang: No. I'm **SURE** it will be released by GC automatically.

erikzhang: https://github.com/neo-project/neo/blob/197a6ec216fe9ef2d17ac20f8ec6d95ed01fb1c3/neo/Persistence/Store.cs#L9

```c#
DataCache<UInt256, TrimmedBlock> IPersistence.Blocks => GetBlocks();
```

You can see the `=>`.

And https://github.com/neo-project/neo/blob/197a6ec216fe9ef2d17ac20f8ec6d95ed01fb1c3/neo/Persistence/LevelDB/LevelDBStore.cs#L44-L47

```c#
public override DataCache<UInt256, TrimmedBlock> GetBlocks()
{
    return new DbCache<UInt256, TrimmedBlock>(db, null, null, Prefixes.DATA_Block);
}
```

ImportBlocks/ImportBlocks.cs
erikzhang: What is this refactor?

ImportBlocks/ImportBlocks.cs
```diff
@@ -17,23 +17,14 @@ public override void Configure()
             Settings.Load(GetConfiguration());
         }
 
-        private bool OnExport(string[] args)
```
vncoelho: maybe keep the `OnExport` function declaration before `WriteBlocks`.

ImportBlocks/ImportBlocks.cs
```diff
@@ -17,23 +17,14 @@ public override void Configure()
...
+        private void WriteBlocks(uint start, uint count, string path, bool writeStart)
         {
-            if (args.Length < 2) return false;
-            if (!string.Equals(args[1], "block", StringComparison.OrdinalIgnoreCase)
-                && !string.Equals(args[1], "blocks", StringComparison.OrdinalIgnoreCase))
-                return false;
-            if (args.Length >= 3 && uint.TryParse(args[2], out uint start))
+            uint end = start + count - 1;
+            using (FileStream fs = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None, 4096, FileOptions.WriteThrough))
```
vncoelho: Can you add a comment on these options?

ImportBlocks/ImportBlocks.cs
```diff
@@ -42,54 +33,65 @@ private bool OnExport(string[] args)
                         fs.Seek(sizeof(uint), SeekOrigin.Begin);
                     }
                     else
+                    {
+                        byte[] buffer = new byte[sizeof(uint)];
```
vncoelho: Maybe declare `buffer` outside this `if else`

ImportBlocks/ImportBlocks.cs
shargon: I told him that both conditions `if/else` was very similar, so i asked him to join this two logics in only one

ImportBlocks/ImportBlocks.cs
KickSeason: will do

ImportBlocks/ImportBlocks.cs
KickSeason: will do

KickSeason: will do

ImportBlocks/ImportBlocks.cs
KickSeason: I will  try and test.

ImportBlocks/ImportBlocks.cs
KickSeason: This is not duplicated. I think this is to make `export` support `export blocks` and `export block`

ImportBlocks/ImportBlocks.cs
KickSeason: * `FileOption.WriteThrough` indicates that the system should write through any intermediate cache and go directly to disk.
* `4096` is the default size of `buffer`.

KickSeason:
@erikzhang @vncoelho I update.

erikzhang:
I don't understand why we have to refactor it?

shargon:
The code is easier to maintain if is unified, the logic between was very similar.

ImportBlocks/ImportBlocks.cs
```diff
@@ -19,79 +20,83 @@ public override void Configure()
...
-                        {
-                            Block block = snapshot.GetBlock(i);
-                            byte[] array = block.ToArray();
-                            fs.Write(BitConverter.GetBytes(array.Length), 0, sizeof(int));
-                            fs.Write(array, 0, array.Length);
-                            Console.SetCursorPosition(0, Console.CursorTop);
-                            Console.Write($"[{i}/{end}]");
-                        }
+                    {
+                        DataCache<UInt256, BlockState> blocks = Blockchain.Singleton.Store.GetBlocks();
```
shargon: This doesn't compile in my branch
```suggestion
                        DataCache<UInt256, TrimmedBlock> blocks = Blockchain.Singleton.Store.GetBlocks();
```

ImportBlocks/ImportBlocks.cs
```diff
@@ -19,79 +20,83 @@ public override void Configure()
...
+                        DataCache<UInt256, TransactionState> transactions = Blockchain.Singleton.Store.GetTransactions();
+                        UInt256 hash = Blockchain.Singleton.GetBlockHash(i);
+                        Block block = blocks.TryGet(hash).TrimmedBlock.GetBlock(transactions);
```
shargon: Same here:

```suggestion
                        Block block = blocks.TryGet(hash).GetBlock(transactions);
```

ImportBlocks/ImportBlocks.cs
```diff
@@ -19,79 +20,83 @@ public override void Configure()
...
-                    if (start <= end)
-                        fs.Write(BitConverter.GetBytes(count), 0, sizeof(uint));
-                    fs.Seek(0, SeekOrigin.End);
+                }
+                if (start <= end)
+                    fs.Write(BitConverter.GetBytes(count), 0, sizeof(uint));
+                fs.Seek(0, SeekOrigin.End);
+                for (uint i = start; i <= end; i++)
+                {
                     using (Snapshot snapshot = Blockchain.Singleton.GetSnapshot())
```
shargon: Snapshot is not used

ImportBlocks/ImportBlocks.cs
shargon: It will be disposed if we call `GetBlocks();` inside the for.

KickSeason:
Much cleaner now. Thanks @shargon .

shargon:
The code looks good to me, @superboyiii could you test it if this solve the issue?

KickSeason:
@nicolegys is testing.

ImportBlocks/ImportBlocks.cs
```diff
@@ -19,79 +19,80 @@ public override void Configure()
...
+                }
+                if (start <= end)
+                    fs.Write(BitConverter.GetBytes(count), 0, sizeof(uint));
+                fs.Seek(0, SeekOrigin.End);
+                for (uint i = start; i <= end; i++)
+                {
+                    var blocks = Blockchain.Singleton.Store.GetBlocks();
+                    var transactions = Blockchain.Singleton.Store.GetTransactions();
+                    UInt256 hash = Blockchain.Singleton.GetBlockHash(i);
+                    Block block = blocks.TryGet(hash).GetBlock(transactions);
```
erikzhang: There is no difference between this way and `Blockchain.Singleton.Store.GetBlock()`.

KickSeason: Different.
`GetBlock()`  
https://github.com/neo-project/neo/blob/e1fdde307153879f0a7ae64bd3e11d042284896e/neo/Persistence/Helper.cs#L31
`GetBlocks()`  
https://github.com/neo-project/neo/blob/e1fdde307153879f0a7ae64bd3e11d042284896e/neo/Persistence/LevelDB/LevelDBStore.cs#L46

erikzhang: I mean:

```c#
var blocks = Blockchain.Singleton.Store.GetBlocks();
var transactions = Blockchain.Singleton.Store.GetTransactions();
UInt256 hash = Blockchain.Singleton.GetBlockHash(i);
Block block = blocks.TryGet(hash).GetBlock(transactions);
```

```c#
Blockchain.Singleton.Store.GetBlock(i);
```

There is no difference between these two ways.

KickSeason: This is Store.GetBlock(index)
```
public static Block GetBlock(this IPersistence persistence, uint index)
 {
       if (index == 0) return Blockchain.GenesisBlock;
       UInt256 hash = Blockchain.Singleton.GetBlockHash(index);
       if (hash == null) return null;
       return persistence.GetBlock(hash);
 }
```
It call GetBlock(hash)
```
public static Block GetBlock(this IPersistence persistence, UInt256 hash)
 {
            TrimmedBlock state = persistence.Blocks.TryGet(hash);
            if (state == null) return null;
            if (!state.IsBlock) return null;
            return state.GetBlock(persistence.Transactions);
 }
```
Which get block from `Blocks` cache and transactions from `Transactions` cache, not new cache.
We need get block from new cache every time.

erikzhang: They are the same. You can test it.

KickSeason: Will test.

KickSeason: Oh! yes. 
`DataCache<UInt256, BlockState> IPersistence.Blocks => GetBlocks();`
You told me. Didn't get it. Anyway let me have a test.
Will make a new commit.

KickSeason: 🙏dalao

KickSeason:
I think we can merge now finally.

superboyiii:
> The code looks good to me, @superboyiii could you test it if this solve the issue?

Nicole's result shows it works effectively and greatly reduces time and memory consume.

nicolegys:
Good performance!
It took about 3h to export the mainnet blocks. 
Here are the monitoring records of cpu and memory.

PR #139, opened 2019-09-30
Fix-RpcNep5Tracker, prevent it from leveldb bug. (2x)

superboyiii:
Fix-RpcNep5Tracker, prevent it from leveldb bug.

### DIFF
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -35,7 +35,7 @@ namespace Neo.Plugins
         private uint _maxResults;
         private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
-        private Fixed8 maxGas = Fixed8.FromDecimal(1000000000m);
+        private Fixed8 maxGas = Fixed8.FromDecimal(100000000m);
 
         public override void Configure()
         {
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -35,6 +35,7 @@ namespace Neo.Plugins
         private uint _maxResults;
         private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
+        private Fixed8 maxGas = Fixed8.FromDecimal(1000000000m);
 
         public override void Configure()
         {
@@ -191,7 +192,7 @@ namespace Neo.Plugins
                     script = sb.ToArray();
                 }
 
-                ApplicationEngine engine = ApplicationEngine.Run(script, snapshot, extraGAS: new Fixed8(100000000));
+                ApplicationEngine engine = ApplicationEngine.Run(script, snapshot, extraGAS: maxGas);
                 if (engine.State.HasFlag(VMState.FAULT)) continue;
                 if (engine.ResultStack.Count <= 0) continue;
                 nep5BalancePair.Value.Balance = engine.ResultStack.Pop().GetBigInteger();
--- a/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -35,7 +35,7 @@ namespace Neo.Plugins
         private uint _maxResults;
         private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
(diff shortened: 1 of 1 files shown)


erikzhang:
What's this?

superboyiii:
@erikzhang We found if new a variable for ExtraGas evertime when execute ApplicationEngine.Run(), it will cause the bug of leveldb. Details please see here. https://github.com/neo-project/neo-cli/issues/472
We've already tested it on mainet, within this hotfix, chain number grows stablely. Without this, chain number grows rapidly which causes no space on disk. It's not a problem from our code but this could avoid causing leveldb bug.

shargon:
This doesn't seems related to levelDB :S

superboyiii:
> This doesn't seems related to levelDB :S

At first, I'm not sure. But when I saw the issue reproduced again, I just try to readjust it. And I test it several times, it truely avoids this issue. I resync the node within this version, it works stablely.

shargon:
Maybe was a random error. Because this change seems that only change the scope of a variable not related with LevelDB.

superboyiii:
@shargon Could you try to sync two nodes on mainet on two different server with previous one and this one installed.  You'll find the previous one will crash because of no space on disk and this one runs stablely.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -35,6 +35,7 @@ public class RpcNep5Tracker : Plugin, IPersistencePlugin, IRpcPlugin
         private uint _maxResults;
         private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
+        private Fixed8 maxGas = Fixed8.FromDecimal(100000000m);
```
vncoelho: private const?

vncoelho:
I am not sure, @superboyiii, however, your tests are usually always right! ahuehaeahuea

If it solves something I think you should proceed because there is not a notorious dangerous on this.

superboyiii:
Hey, Vitor, my man. Thanks for your trust. This story begins here: https://github.com/google/leveldb/issues/320.
Details can be seem here: https://github.com/neo-project/neo-cli/issues/472
What I did was tested to prevent neo-cli from this issue, that's an accidental chance when I found that changing code to this way could avoid this issue. Could you help me to double check? 
You need to do these:
1. Sync a node with master-2.x's version, build RpcNep5Tracker, and put it into ```Plugins``` folder. Try to sync to the latest height with offline package: https://sync.ngd.network/. After about ten hours, you'll find your disk has few space left.
2. Build this version and do the same steps again. After several hours, you'll find everything is normal and your node has already sync to the latest height.

RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -35,6 +35,7 @@ public class RpcNep5Tracker : Plugin, IPersistencePlugin, IRpcPlugin
         private uint _maxResults;
         private bool _shouldTrackNonStandardMintTokensEvent;
         private Neo.IO.Data.LevelDB.Snapshot _levelDbSnapshot;
+        private Fixed8 maxGas = Fixed8.FromDecimal(1m);
```
erikzhang:
```suggestion
        private static readonly Fixed8 maxGas = Fixed8.FromDecimal(1m);
```

erikzhang:
Please test it.

vncoelho:
Surely, @superboyiii, my pleasure.
It is always great to hear the history behind, these ideas should be valued because of that, it is always not easy to find any problem. But, when we find we are happy ;D

Give me some couple of days.
Can you describe 1. again, please?
I need to sync the node first and then include the plugin? Can not just include the plugin and sync with bootstrap?

shargon:
We can upgrade leveldb in order to prevent future issues

superboyiii:
@shargon Yes, that's the best solution. Please take a look at here. https://github.com/neo-project/neo-cli/issues/472 @Ashuaidehao has already build the latest leveldb library, and we've test it on three different OS: Win10，Ubuntu 16.04 and MacOS. We use base case to make function and rpc test, all pass. The final step we have to do is compare it with previous data to ensure it's compatible. @nicolegys is doing this, it will take several days, let's wait for it.

superboyiii:
> Surely, @superboyiii, my pleasure.
> It is always great to hear the history behind, these ideas should be valued because of that, it is always not easy to find any problem. But, when we find we are happy ;D
> 
> Give me some couple of days.
> Can you describe 1. again, please?
> I need to sync the node first and then include the plugin? Can not just include the plugin and sync with bootstrap?

Bootstrap is available. Recommend to use bootstrap.

vncoelho:
@superboyiii, can we merge it?

I did not had time to test. However, the change is trivial and does not produces any danger.

vncoelho:
If merged, perhaps we need to port to master branch.

PR #141, opened 2019-11-05
Fix/memory (#135) (master-2x)

KickSeason:
Apply to 2.x

* use Singleton.Store.GetBlock optimize memory cause

* refactor


superboyiii:
Tested. It takes about 4 hours for exporting mainnet chain data and no memory exception.
image
image

PR #142, opened 2019-11-11
Fix cell on maxNBlocksPerDay (3.X)

superboyiii:
uint (86400/15000*10000) equals 5000 but not 5760 because ```maxNBlocksPerDay``` is uint but not float.

### DIFF
--- a/CoreMetrics/CoreMetrics.cs
+++ b/CoreMetrics/CoreMetrics.cs
@@ -33,7 +33,7 @@ namespace Neo.Plugins
         private JObject GetBlocksTime(uint nBlocks, uint lastHeight)
         {
             // It is currently limited to query blocks generated in the last 24hours (86400 seconds)
-            uint maxNBlocksPerDay = 86400 / Blockchain.MillisecondsPerBlock * 1000;
+            uint maxNBlocksPerDay = 86400 / (Blockchain.MillisecondsPerBlock / 1000);
             if (lastHeight != 0)
             {
                 if (lastHeight >= Blockchain.Singleton.Height)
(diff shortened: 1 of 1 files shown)


CoreMetrics/CoreMetrics.cs
```diff
@@ -60,13 +60,7 @@ private JObject GetBlocksTime(uint nBlocks, uint lastHeight)
                 JObject json = new JObject();
                 return json["error"] = "Requested number of blocks timestamps " + nBlocks + " exceeds quantity of known blocks " + Blockchain.Singleton.Height;
             }
-
-            if (nBlocks <= 0)
```
shargon: ==?

superboyiii: @shargon  Uint can't be < 0,  because input -1 means ```nBlocks``` = uint 4294967295.

superboyiii: image

shargon: > UInt can't be < 0, because input -1 means nBlocks = uint 4294967295.

That's the reason for change to == 0.

CoreMetrics/CoreMetrics.cs
```diff
@@ -33,7 +33,7 @@ public JObject OnProcess(HttpContext context, string method, JArray _params)
         private JObject GetBlocksTime(uint nBlocks, uint lastHeight)
         {
             // It is currently limited to query blocks generated in the last 24hours (86400 seconds)
-            uint maxNBlocksPerDay = 86400 / Blockchain.MillisecondsPerBlock * 1000;
+            uint maxNBlocksPerDay = 86400 / (Blockchain.MillisecondsPerBlock / 1000);
```
shargon: Good catch

CoreMetrics/CoreMetrics.cs
superboyiii: > > UInt can't be < 0, because input -1 means nBlocks = uint 4294967295.
> 
> That's the reason for change to == 0.

Oh, got it!

PR #143, opened 2019-11-14
Moving RPC Server to neo-plugins

melanke:
Hello,
As discussed on [Neo's #1139 PR](https://github.com/neo-project/neo/pull/1139),
We are moving the RPC Server to neo-plugins, adding Rest compatibility and an easier way to implement new RPC methods, using "Controllers".

I had to change IRpcPlugin interface to be able to use multiple Rpc Servers vs multiple Rpc Plugins.
Instead of implementing 3 methods you only need to implement one and use RpcServer methods to configure the plugin. Beyond functionalities included in the previous PR like bindOperation and injectSpecialParameter you can now add and remove request and response interceptors to listen and modify any request.


RpcServer/RpcServer.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Version>3.0.0-preview1</Version>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <RootNamespace>Neo.Plugins</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\neo\neo\neo.csproj" />
```
erikzhang: We should add references from NuGet.

neo-plugins.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28729.10
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.271
```
erikzhang: Please don't change these lines.

erikzhang:
The indentations are not good.

erikzhang:
By the way, have we decided to move the RpcServer to the plugin?

shargon:
We need a way for interact between plugins in order to move the rpc to a plugin

erikzhang:
Why we need to interact between plugins?

RpcServer/RpcServer.csproj
melanke: Alright, I will just keep this on the draft to be able to test with neo changes I made. I will make sure to change this before the final PR.

melanke:
The IRpcServer plugin and IRpcPlugin are bound together like this:

RpcServer calls BeforeStartServer of IRpcPlugin:
```C#
		public override void Configure()
		{
			_settings = new RpcServerSettings(GetConfiguration());
			InjectSpecialParameter(ctx => ctx);
			InjectSpecialParameter(ctx => _settings);
			
			foreach (IRpcPlugin plugin in Plugin.RpcPlugins.ToList())
			{
				plugin.BeforeStartServer(this);
			}

			Start();
		}
```

And on the IRpcPlugin it can configure itself:
```C#
		public void BeforeStartServer(IRpcServer server)
		{
			server.BindController<NeoDefaultRpcController>(); 
                        // here you can implement interceptors and inject parameters aswell.
		}
```

I am thinking about to make a small documentation on how the developer can create plugins dor the RpcServer.

shargon:
> Why we need to interact between plugins?

Because now we can make plugins for RPC, if is moved to a new one, we can't do that without interact

RpcServer/RpcServer/RcpTargetAndMethod.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Reflection;
+
+namespace Neo.Plugins.RpcServer
+{
+    internal class RcpTargetAndMethod
```
erikzhang: typo

RpcServer/RpcServer/RpcServer.cs
```diff
@@ -0,0 +1,744 @@
...
+        }
+
+        private JObject CreateResponse(bool isNewVersion, string id)
+        {
+            var response = new JObject();
+
+            if (isNewVersion)
+            {
+                response["restResponse"] = "1.0";
+            }
```
erikzhang: What's this?

erikzhang:
I have a good way to interact between plugins.

But if we move RpcServer to a plugin, then we need to move RpcClient too.

lock9:
@erikzhang  maybe it doesn`t have to be a plugin if we can detach the storage from the library.
This way the library would support both a node using local data or using 'remote storage' through RPC.

RpcServer/RpcServer/RpcServer.cs
melanke: if the request comes with the REST format we can't answer with a jsonRpc version, so it is setting a "restResponse" version

erikzhang: I think it's not a good idea to mix JSON-RPC with REST.

lock9: Maybe we need a plugin for Http and then a plugin for RPC and another for Rest.

melanke: I can easily separate http, json-rpc and rest, this is the best solution?
But without json-rpc or rest, there will be no neo methods implemented.

melanke: Ok, it was not easy, but now we have an HttpServer plugin, a RpcInterceptor and RestInterceptor.
They are all in the same package for now but I will separate them.

HttpServer/HttpServer/HttpServer.cs
```diff
@@ -0,0 +1,596 @@
...
+        }
+
+        public override void OnPluginsLoaded()
+        {
+            ConfigurePlugins();
+        }
+
+        public async void ConfigurePlugins()
+        {
+            await Task.Delay(500);
```
melanke: I believe this isn't the best solution, but the app crashes when I dont wait for plugins initialization with this delay

HttpServer/HttpServer/HttpServer.cs
```diff
@@ -0,0 +1,596 @@
...
+using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.Extensions.DependencyInjection;
+using Neo.IO.Json;
+
+
+namespace Neo.Plugins.HttpServer
+{
+    using Interceptor = Action<IHttpOperationPayload>;
+
+    public class HttpServer : Plugin, IHttpServer
```
erikzhang: What is `IHttpServer`?

lock9: Hi @erikzhang, the common code between RPC and Rest is the "HTTP engine". The IHttpServer is the "HTTP provider".
This implementation allows RPC and REST independently. You add interceptors to the requests and responses, allowing you to add RPC and REST endpoints independently.

erikzhang: I need to see the interface details.

lock9: I'm 90% confident (because I've seen the code) that this interface has only these 2 methods:
AddRequestInterceptor and AddResponseInterceptor.
Check https://github.com/neo-project/neo-plugins/pull/143/files#diff-506d69ca242c5fb9ebccbe04e0baa14f for current RPC implementation.

erikzhang: This interface should be in the plugin, not in Neo.dll.

lock9: If we do that, we are not going to allow different 'HTTP server' implementations. If that is ok for you, I agree.

lock9: Actually, the interface _has_ to be on neo, otherwise, it won't be possible for people to implement it without adding a reference to this plugin.

erikzhang: Plugins can also reference each other. I don't want all plugin interactions to depend on changes to Neo.

lock9: Ok, I see the issue. I thought that we wanted to allow plugins to communicate without knowing that other plugins exist. If we expect plugins to reference other plugins, I agree that it can be moved to the plugins.

melanke: But all plugins have an interface:
- ILogPlugin
- IMemoryPoolTxObserverPlugin
- IP2PPlugin
- IPersistencePlugin

So IHttpServer is just another plugin interface

melanke: @erikzhang I am following the Plugin Implementation pattern, are you planning to change this pattern? I am pretty sure it would be impossible to have multiple implementation of HttpServer without this interface inside Neo.dll, it is just the way all plugins are working.

melanke: Because the HttpServer is the base plugin, and Rpc and Rest are other plugins that depend on it.

melanke:
```C#
    public abstract class Plugin : IDisposable
    {
        public static readonly List<Plugin> Plugins = new List<Plugin>();
        private static readonly List<ILogPlugin> Loggers = new List<ILogPlugin>();
        internal static readonly List<IHttpServer> HttpServers = new List<IHttpServer>();
        protected static readonly List<IHttpPlugin> HttpPlugins = new List<IHttpPlugin>();
        internal static readonly List<IPersistencePlugin> PersistencePlugins = new List<IPersistencePlugin>();
        internal static readonly List<IP2PPlugin> P2PPlugins = new List<IP2PPlugin>();
        internal static readonly List<IMemoryPoolTxObserverPlugin> TxObserverPlugins = new List<IMemoryPoolTxObserverPlugin>();
```

```C#
protected Plugin()
        {
            Plugins.Add(this);

            if (this is ILogPlugin logger) Loggers.Add(logger);
            if (this is IP2PPlugin p2p) P2PPlugins.Add(p2p);
            if (this is IHttpServer rpcServer) HttpServers.Add(rpcServer);
            if (this is IHttpPlugin rpc) HttpPlugins.Add(rpc);
            if (this is IPersistencePlugin persistence) PersistencePlugins.Add(persistence);
            if (this is IMemoryPoolTxObserverPlugin txObserver) TxObserverPlugins.Add(txObserver);

            Configure();
        }
```

melanke:
I will need to fix the other plugins that depends on RPC Server when we decide this draft is good

HttpServer/HttpServer/HttpServer.cs
erikzhang: `HttpPlugin` can reference to `HttpServer` directly. So it doesn't need an interface. In this way we don't need to modify Neo.

erikzhang:
@melanke I have a proposal. I will first port `RpcServer` to the plugin, and then you work based on the ported plugin. Is it okay for you?

melanke:
> @melanke I have a proposal. I will first port `RpcServer` to the plugin, and then you work based on the ported plugin. Is it okay for you?

I already ported to the plugin. Don't you want to continue from by branch? I think it would be easier and time saving for us both.

HttpServer/HttpServer/HttpServer.cs
melanke: @erikzhang I could drop IHttpServer but then the Http Plugins could only work with this HttpServer implementation, not a generic one, I can easily do that if that is your intention.

erikzhang: I really don't think we need multiple HttpServer implementations.

melanke: Alright, so I will change that, thanks for the clarification.

melanke: Done

RpcServer/RpcServer.csproj
erikzhang: Can we use nuget package now?

melanke: I will need to send my changes on Neo first, but they will break the plugins that rely on RPC, so, it is kinda the egg-or-the-chicken situation hahaha.

I will fix this plugins now, so we can do this whole change faster.

erikzhang: You can send the PR to Neo first to let us know what changes are made.

PR #145, opened 2019-11-19
Add fee information in ApplicationLogs plugin

shargon:
Close https://github.com/neo-project/neo-plugins/issues/137

### DIFF
--- a/ApplicationLogs/LogReader.cs
+++ b/ApplicationLogs/LogReader.cs
@@ -53,6 +53,16 @@ namespace Neo.Plugins
             {
                 JObject json = new JObject();
                 json["txid"] = appExec.Transaction.Hash.ToString();
+                if (appExec.Transaction != null)
+                {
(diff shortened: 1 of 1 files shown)


shargon:
@hal0x2328 please review it if this can solve the problem

hal0x2328:
It's a good thing to have the fees split out separately in the log but I don't think it solves the problem if the notification from the GAS contract still shows address B receiving the entire balance from address A - how do you know in which case to subtract the fee from that total, since it only happens when the amount transferred is A's entire balance?

Although you could make the plugin subtract the fee from the notification, it seems like it should probably be fixed in the GAS contract itself.

shargon:
it's for neo2 or neo3?

hal0x2328:
> it's for neo2 or neo3?

Neo3

shargon:
I will transfer the issue to the neo repository

PR #150, opened 2019-12-01
Add LevelDBStore

erikzhang:


ApplicationLogs/ApplicationLogs.csproj
```diff
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="3.0.0-CI00804" />
+    <ProjectReference Include="..\LevelDBStore\LevelDBStore.csproj" />
```
shargon: Should be great if the plugin use the current Store. Something like a `Clone(string path)` in `IStore`

LevelDBStore/IO/Data/LevelDB/DB.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public void Put(WriteOptions options, byte[] key, byte[] value)
+        {
+            Native.leveldb_put(handle, options.handle, key, (UIntPtr)key.Length, value, (UIntPtr)value.Length, out IntPtr error);
+            NativeHelper.CheckError(error);
+        }
+
+        public void Write(WriteOptions options, WriteBatch write_batch)
+        {
+            // There's a bug in .Net Core.
```
shargon: Wait for https://github.com/neo-project/neo-cli/issues/472#issuecomment-560330816

LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -0,0 +1,40 @@
...
+            using Iterator it = db.NewIterator(options);
+            for (it.Seek(startKey); it.Valid(); it.Next())
+            {
+                byte[] key = it.Key();
+                if (key.AsSpan().SequenceCompareTo(endKey) > 0) break;
+                yield return resultSelector(key, it.Value());
+            }
+        }
+
+        internal static byte[] ToByteArray(this IntPtr data, UIntPtr length)
```
shargon:
```suggestion
        internal static byte[] ToByteArray(this IntPtr data, int length)
```

LevelDBStore/LevelDBStore.csproj
```diff
@@ -0,0 +1,21 @@
...
+
+  <ItemGroup>
+    <None Update="LevelDBStore\config.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Neo" Version="3.0.0-CI00810" />
```
shargon: Time to move levelDB to a nuget in order to avoid library issues? (maybe hosted by neo)

ApplicationLogs/ApplicationLogs.csproj
erikzhang: What do you mean?

LevelDBStore/LevelDBStore.csproj
erikzhang: What library issues?

LevelDBStore/IO/Data/LevelDB/Helper.cs
erikzhang: The caller uses `UIntPtr` to receive the length from C++ library.

ApplicationLogs/ApplicationLogs.csproj
shargon: I mean that if the node is using RocksDB, this plugin use mandatory LevelDB

LevelDBStore/LevelDBStore.csproj
shargon: If you compile from the source, you need the levelDB dll or you can't run the node, with release there is no problem

ApplicationLogs/ApplicationLogs.csproj
erikzhang: This plugin uses leveldb with a different store. We can improve it in the future.

LevelDBStore/LevelDBStore.csproj
erikzhang: Currently only windows needs it.

shargon: Ok, we can improve it later !

PR #151, opened 2019-12-02
Remove Travis and use Github Actions

erikzhang:


.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,24 @@
...
+
+env:
+  DOTNET_VERSION: 3.0.100
+
+jobs:
+
+  Test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Chectout
```
shargon:
```suggestion
    - name: Checkout
```

vncoelho: I also noticed this on the NEO, we can fix together with another PR there.

erikzhang:
We can automatically publish the release if the version changed.

lock9:
Hmm, not sure if I understand. I can't see from the current configuration release instructions. Are you sure the release process is automated? Where can I find the rules?

erikzhang:
@lock I need to do some research first. I don't know how to implement it. We need to improve it later.

PR #153, opened 2019-12-05
Fix leveldb

erikzhang:
Fixes neo-project/neo#1327


superboyiii:
Tested, it works.
image

PR #157, opened 2019-12-07
FASTER plugin

shargon:
Close https://github.com/neo-project/neo-plugins/issues/155
Require UT from https://github.com/neo-project/neo-plugins/pull/144

I need to do some test, but I think that FASTER wont be a valid option for this reason.

- The snapshot system I think that is not the same as LevelDB or RocksDB, you can't (or I need to do more research) make multiples Snapshots, because `CompleteCheckpoint` doesn't accept the checkpoint Id. you can `Revert` to this checkpoint, but `Commit` only to the last of them. https://github.com/microsoft/FASTER/blob/54f23a4524d3d5f86df56663eb984f81490ca600/cs/playground/ManagedSampleCore/ICustomFaster.cs#L133-L138

***Note: I have troubles with `.github/workflows/dotnetcore.yml` in my own repo, or I remove the file or I can't make a pull request :S***


shargon:
I will open it again in neo repository

PR #158, opened 2019-12-07
FASTER Storage plugin

shargon:
**DRAFT-DON'T MERGE**

Close https://github.com/neo-project/neo-plugins/issues/155
Require UT from https://github.com/neo-project/neo-plugins/pull/144


shargon:
Ready to review and test the first draft, but there are something that I don't like it.

- `Read/Write/Delete` are not relative to the snapshot. And, if you restart the storage, the unique way for recover the previous state that I find it's with snapshots.... I made a work around for this "issue" saving in the `session.id` file the last valid snapshot.
  - Ppossible problems, Dispose the store before the snapshot. It's possible that the snapshot that I made will be related to the current snapshot, because the operations are no related to the snapshot.
- I can't find a good way for remove snapshots.

shargon:
This UT fail only in this `FASTER` implementation. Write in the store in the middle of a snapshot. We need to solve this before merge it.

image

src/FasterDBStore/FasterDBStore.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <ItemGroup>
+    <None Update="RocksDBStore\config.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.FASTER" Version="2019.11.18.1" />
+    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" />
```
erikzhang: What's this?

src/RocksDBStore/Plugins/Storage/Store.cs
```diff
@@ -87,9 +87,7 @@ public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
             for (it.Seek(prefix); it.Valid(); it.Next())
             {
                 var key = it.Key();
-                byte[] y = prefix;
-                if (key.Length < y.Length) break;
-                if (!key.AsSpan().StartsWith(y)) break;
+                if (key.Length < prefix.Length) break;
```
erikzhang: Is this correct??

src/FasterDBStore/FasterDBStore.csproj
shargon: Without this, travis doesn't work... I don't know why

src/RocksDBStore/Plugins/Storage/Store.cs
shargon: I will check it

src/FasterDBStore/FasterDBStore.csproj
shargon: https://github.com/neo-project/neo-plugins/runs/338134148

vncoelho:
> This UT fail only in the FASTER implementation. Write in the store in the middle of a snapshot. We need to solve this before merge it.

I had this same problem in the UTs of the consensus, @shargon.
I needed to refresh the snapshot with the one from `Singleton` before doing a proper Delete.

shargon:
@vncoelho are different errors, because this is at low level. The storage, apparently, is not able to write outside the Snapshot when a snapshot it's made

src/FasterDBStore/Store.cs
```diff
@@ -0,0 +1,186 @@
...
+            if (File.Exists(Path.Combine(path, "session.id")))
+            {
+                var data = new byte[16];
+                using (var file = File.OpenRead(Path.Combine(path, "session.id")))
+                {
+                    if (file.Read(data, 0, 16) == 16)
+                    {
+                        // Recover the last storage state
+
+                        db.Recover(new Guid(data));
```
bettybao1209: This line may throw the FasterException `Invalid index commit metadata for ID ***`, which is caused by the absence of the `completed.dat` file in the corresponding guid directory (under the snapshot directory).

tests/Neo.Plugins.Storage.Tests/StoreTest.cs
```diff
@@ -70,40 +94,98 @@ private void TestStorage(IStore store)
...
+                using (var snapshot = store.GetSnapshot())
+                {
+                    // Create a dummy snapshot and remove from the storage
+
+                    store.Delete(0x10, key);
+                }
 
-                if (shouldExist) CollectionAssert.AreEqual(new byte[] { 0x04, 0x05, 0x06 }, ret);
-                else Assert.IsNull(ret);
+                Assert.IsNull(store.TryGet(0x10, key));
```
bettybao1209: I think the non-null result occurs because you recover the db when disposing the snapshot.

src/FasterDBStore/Store.cs
bettybao1209: Maybe we can use try-catch to deal with the exception caused by the file missing issue.

erikzhang:
@shargon Maybe this can be closed?

PR #160, opened 2019-12-07
Add RpcServer

erikzhang:


erikzhang:
Should we move the RpcClient to this plugin? Or should we create a new repo?

vncoelho:
@erikzhang, I think that a new repo may not be the best direction, because, perhaps, soon, other similar implementations will be requested, such as a REST Client.

vncoelho:
Maybe later we could merge `RpcNep5Tracker` into `RpcServer`. What do you think?

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -0,0 +1,204 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Neo.Plugins
+{
+    partial class RpcServer
+    {
+        [RpcMethod]
+        private JObject GetBestBlockHash(JArray _params)
```
lock9: Isn't this going to change all RPC endpoint names? For me it is ok to change, just to be sure.

src/RpcServer/RpcServer.cs
```diff
@@ -0,0 +1,243 @@
...
+                app.UseResponseCompression();
+                app.Run(ProcessAsync);
+            })
+            .ConfigureServices(services =>
+            {
+                services.AddResponseCompression(options =>
+                {
+                    // options.EnableForHttps = false;
+                    options.Providers.Add<GzipCompressionProvider>();
+                    options.MimeTypes = ResponseCompressionDefaults.MimeTypes.Append("application/json-rpc");
```
lock9: Hi Erik, we checked this online and it seems that json-rpc doesn't exist, we should use "application/json" instead.

erikzhang:
@lock9 `RpcServer.Blockchain.cs` is not a separate class. All `RpcServer.*.cs` file is a single class.

src/RpcServer/RpcServer.Blockchain.cs
erikzhang: https://github.com/neo-project/neo-modules/blob/ae5cdee38cc02ec062ba0d5536ea9655497a5e77/src/RpcServer/RpcServer.cs#L238

It won't change the name.

src/RpcServer/RpcServer.cs
erikzhang: Fixed

shargon:
We need to release this plugin in order to test https://github.com/neo-project/neo/pull/1349

shargon:
@superboyiii could you test this copying the dependency manually ?

superboyiii:
@shargon Fine.

src/RpcServer/RpcServer.cs
```diff
@@ -0,0 +1,243 @@
...
+            catch
+            {
+                return false;
+            }
+
+            string[] authvalues = authstring.Split(new string[] { ":" }, StringSplitOptions.RemoveEmptyEntries);
+            if (authvalues.Length < 2)
+                return false;
+
+            return authvalues[0] == Settings.Default.RpcUser && authvalues[1] == Settings.Default.RpcPass;
```
superboyiii: A conflict of namespace here about ```Settings``` on ApplicationLogs and RpcServer which is able to return false here.

superboyiii:
```
authvalues[0]
"aaa"
Settings.Default.RpcUser
error CS0433: The type 'Settings' exists in both 'ApplicationLogs, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null' and 'RpcServer, Version=3.0.0.0, Culture=neutral, PublicKeyToken=null'
```
image

erikzhang: No problem here. They are both `internal`.

superboyiii: ApplicationLogs is using RpcServer as references. Is it related to this? Loading sequence issue?
image

src/ApplicationLogs/LogReader.cs
```diff
@@ -23,31 +21,24 @@ public class LogReader : Plugin, IRpcPlugin, IPersistencePlugin
         public LogReader()
         {
             db = DB.Open(GetFullPath(Settings.Default.Path), new Options { CreateIfMissing = true });
+            RpcServer.RegisterMethods(this);
```
superboyiii: ApplicationLogs is using RpcServer as references. Is it related to this? Loading sequence issue?

superboyiii: I'm not able to pas ```CheckAuth```.
image

erikzhang: Did you add `Authorization` entry in the HTTP headers?

superboyiii: > Did you add `Authorization` entry in the HTTP headers?

Yes, Of course.
image
image
image
You can see they're the same.

erikzhang: Then it should return `true`.

erikzhang: Did you open wallet via RPC? Opening it in CLI has no effect. Only wallets opened in RPC can be accessed by RPC.

erikzhang: You need to use `openwallet` command to open the wallet in RPC.

superboyiii: Yes, but that's not the point. The issue is that `Settings.Default.RpcUser` and `Settings.Default.RpcPass` can't get value because of `Settings` conflict on ApplicationgLogs and RpcServer. So it can't return `true` here on authvalues[].
image

superboyiii: > You need to use `openwallet` command to open the wallet in RPC.

Yes, that's the truth. It works. But I'm a little confused that why wallet has to be opened by RPC to continue invoking other wallet related RPC?  Will it bring risks?

erikzhang: There is no conflict at the runtime. Because they are both `internal` class. The conflict is in your debugger. You can't use the immediate window to get the values.

superboyiii:
Tested, all rpc methods works well on mine.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -1,187 +1,92 @@
...
-                    tx = null;
-                result["tx"] = tx?.ToArray().ToHexString();
-            }
-        }
+        private Wallet wallet;
 
         private void CheckWallet()
         {
-            if (Wallet is null)
+            if (wallet is null)
```
shargon: Maybe we can have another Attribute for `RequireWallet` ?

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -1,187 +1,92 @@
...
-            {
-                return context.ToJson();
-            }
+            wallet = null;
+            return true;
         }
 
-        private JObject DumpPrivKey(UInt160 scriptHash)
+        [RpcMethod]
+        private JObject DumpPrivKey(JArray _params)
```
shargon: We should have another Attribute for `RequireAuth` in order to call wallet methods.

src/RpcServer/RpcServer.Wallet.cs
erikzhang: We don't need it. Because if you need a wallet, you should at least be able to access the wallet object. Now that you have access, you can easily determine if the wallet object is empty.

src/RpcServer/RpcServer.Wallet.cs
erikzhang: Currently if you have set `RpcUser` and `RpcPass` in config.json, you need to authenticate for any method.

shargon: Maybe we should disallow some methods by default in the config. And if the user want to use it, they can allow them and set a password

erikzhang: https://github.com/neo-project/neo-modules/blob/fae8c17baffde1cf105bd21f8eed35e9a0f96eda/src/RpcServer/RpcServer/config.json#L12

`openwallet` is disabled by default.

shargon: But if you install the rpc, and open a wallet by your self, you will have troubles. We should disable by default all the wallet methods

superboyiii: Hey Shargon, due to my tests，all wallet methods are disabled by default unless you invoke "openwallet" methods，it's absolutely not related to you open your wallet or not in local, they are different business.

erikzhang: https://github.com/neo-project/neo-modules/pull/160#discussion_r356434877

shargon: Then is good to me, thanks for clarify!

erikzhang:
Merge?

PR #164, opened 2019-12-09
Remove CoreMetrics

erikzhang:
I think this plugin should not be in `neo-modules`.


Tommo-L:
Or could we create a `extensions` or `plugins` folder, it will be more clear.

```
- LevelDBStore
- RocksDBStore
- RpcServer
- Extensions
          - ApplicationLogs
          - CoreMetrics
          - SystemLog
          - StatesDumper
```

vncoelho:
I agree with @Tommo-L, if this repository. In fact, neo-plugins was not a bad name, @erikzhang.
Someplace need to host such "Extensions/Plugins", officially or not.

erikzhang:
I don't know what is this. It seems this plugin is never used.

vncoelho:
In fact, @erikzhang, I think that some people were using it, because this is a simple way to get block times without requiring a great load from the RPCs.

The idea was to merge other types of metrics into this plugin in a near future.

PR #171, opened 2019-12-31
adapt RpcClient to modules

chenquanyu:
1. Support plugins RPC Methods
2. Add Basic Auth for RpcClient
3. Support Wallet RPC Methods
4. Auto calculate the network fee in RpcClient


chenquanyu:
Hi @vncoelho, 
`RpcServer` is a plugin that will be used in the `Plugins` folder, but `RpcClient` is a C# SDK which will be supplied as a NuGet package. So I don't think they will be enabled together.

For this PR I'm not sure whether we should support `openwallet` methods in `RpcClient`. And it would be better if we can reconstruct the `RpcServer` and plugins with `Neo.Network.RPC.Models`.

vncoelho:
@chenquanyu, I still do not have it fully clear. But I believe it will be good for developers.

When it is ready to merge let us know.

vncoelho:
@chenquanyu, can you proceed with the renaming suggested by @wanglongfei88 in #175?
If you agree, perhaps **RpcSDK** may look good. Fell free to adjust.

erikzhang:
I don't want to rename it. 😂

wanglongfei88:
> I don't want to rename it. 😂

😂😂，OK

vncoelho:
I think it should be renamed, @erikzhang, RpcClient does not look easy to be abstracted.

vncoelho:
@chenquanyu, I just merged into your branch.

Remember to create a branch in your repo next time. Otherwise your master becomes a little bit confusing...aehauea

src/RpcClient/TransactionManager.cs
```diff
@@ -91,9 +91,6 @@ public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[]
             throw new InvalidOperationException($"Insufficient GAS in address: {sender.ToAddress()}");
         }
 
-        /// <summary>
-        /// Calculate NetworkFee
-        /// </summary>
         /// <param name="isEstimate">assuming the witnesses are basic Signature Contract if set to true</param>
```
shargon: why remove this comment?

src/RpcClient/Models/RpcApplicationLog.cs
```diff
@@ -0,0 +1,87 @@
...
+using Neo.VM.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Neo.Network.RPC.Models
+{
+    public class RpcApplicationLog
+    {
+        public UInt256 TxId { get; set; }
```
shargon: TxHash?

src/RpcClient/Models/RpcNep5Transfers.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public ushort TransferNotifyIndex { get; set; }
+
+        public UInt256 TxHash { get; set; }
+
+        public JObject ToJson()
+        {
+            JObject json = new JObject();
+            json["timestamp"] = TimestampMS;
+            json["asset_hash"] = AssetScriptHash.ToString();
```
shargon: Camelcase for all? `assetHash`?

src/RpcClient/RpcClient.cs
```diff
@@ -277,35 +321,49 @@ public RpcPlugin[] ListPlugins()
...
+        /// This method is provided by the plugin RpcNep5Tracker.
         /// </summary>
-        public bool SubmitBlock(byte[] block)
+        /// <param name="address">The address to query the transaction information.</param>
+        /// <param name="startTimestamp">The start block Timestamp, default to seven days before UtcNow</param>
+        /// <param name="endTimestamp">The end block Timestamp, default to UtcNow</param>
+        public RpcNep5Transfers GetNep5Transfers(string address, ulong? startTimestamp = default, ulong? endTimestamp = default)
         {
-            return RpcSend("submitblock", block.ToHexString()).AsBoolean();
+            startTimestamp ??= (DateTime.UtcNow - TimeSpan.FromDays(7)).ToTimestampMS();
```
shargon: 0 by default?

src/RpcClient/TransactionManager.cs
chenquanyu: Add back😂

src/RpcClient/Models/RpcNep5Transfers.cs
chenquanyu: Hi @shargon,
Do you mean also change "asset_hash" in Plugin `RpcNep5Tracker`?

shargon: I mean, Ussually we use this format `assetHash` instead of `asset_hash` isn't it ?

chenquanyu: Yes, how about move these models to `RpcServer`, then we can use the `ToJson()` method in all the plugins.

src/RpcClient/Models/RpcNep5Transfers.cs
```diff
@@ -0,0 +1,81 @@
...
+        public List<RpcNep5Transfer> Sent { get; set; }
+
+        public List<RpcNep5Transfer> Received { get; set; }
+
+        public JObject ToJson()
+        {
+            JObject json = new JObject();
+            json["sent"] = Sent.Select(p => p.ToJson()).ToArray();
+            json["received"] = Received.Select(p => p.ToJson()).ToArray();
+            json["address"] = UserScriptHash.ToAddress();
```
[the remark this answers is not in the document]
chenquanyu: It seems incorrect to name an `UInt160` type as `Address`.

src/RpcClient/Models/RpcNep5Transfers.cs
chenquanyu: Hi @shargon, I have move models to `RpcServer`. Camelcase for all will change the JSON in several responses, I will do the change if you think it's necessary.

erikzhang: No....... These models are not belong to `RpcServer`...

nicolegys:
@chenquanyu Is this pr ready for testing now?

src/RpcClient/Models/RpcNep5Transfers.cs
chenquanyu: OK, I have move models back😂

chenquanyu:
> @chenquanyu Is this pr ready for testing now?

Yes, it's ready for testing, I will look into the `dotnet-format` error.

shargon:
dotnet format errors are global, something was changed in the last version of dotnet format...

nicolegys:
FromJson method in Contract.cs has been removed😂
https://github.com/neo-project/neo/tree/65215823bc320ed8d4c6249683700c9d8199c103

chenquanyu:
> FromJson method in Contract.cs has been removed😂

@shargon, Shall we add it back?

shargon:
> Shall we add it back?

I think that this module can implement an extension, if a module require something and it could solve by himself, the module should do that.

src/RpcClient/RpcClient.cs
```diff
@@ -199,34 +210,34 @@ public RpcRawMemPool GetRawMempoolBoth()
...
         {
-            return RpcTransaction.FromJson(RpcSend("getrawtransaction", txid, true));
+            return RpcTransaction.FromJson(RpcSend("getrawtransaction", txHash, true));
         }
 
         /// <summary>
         /// Returns the stored value, according to the contract script hash and the stored key.
         /// </summary>
-        public string GetStorage(string script_hash, string key)
+        public string GetStorage(string scriptHash, string key)
```
nicolegys: Maybe this param name could be changed? Since we can also use `Id` to get storage.

shargon: We should accept both, because it's easier for the user know the script hash.

chenquanyu: Yes, I have changed the code, please have a check.

src/RpcClient/Models/RpcInvokeResult.cs
```diff
@@ -8,29 +9,45 @@ public class RpcInvokeResult
...
+            try
+            {
+                json["stack"] = new JArray(Stack.Select(p => p.ToJson()));
+            }
+            catch (InvalidOperationException)
+            {
+                // ContractParameter.ToJson() may cause InvalidOperationException
+                json["stack"] = "error: recursive reference";
+            }
+            if (Tx != null) json["tx"] = Tx;
```
shargon: `string.IsNulOrEmpty`?

src/RpcClient/Models/RpcNep5Transfers.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        public ushort TransferNotifyIndex { get; set; }
+
+        public UInt256 TxHash { get; set; }
+
+        public JObject ToJson()
+        {
+            JObject json = new JObject();
+            json["timestamp"] = TimestampMS;
+            json["asset_hash"] = AssetHash.ToString();
```
shargon: We should define a json coding style rules. Use '_'  or camelcase.

chenquanyu: Now, `snake_case` is more common in current JSON.  Camelcase seems only appear in  `entryPoint`, `returnType`, `tcpPort` and `vmState`.

shargon: Maybe we should modify manifest to use `snake_case` but I think that we should unify the behaviour

chenquanyu: Yes, the unifying will change both `neo` and `neo-modules` code. Maybe we can raise an issue and have a discussion on using `snake_case` or `camelCase`.

src/RpcClient/RpcClient.cs
```diff
@@ -17,22 +19,45 @@ namespace Neo.Network.RPC
...
     public class RpcClient : IDisposable
     {
-        private readonly HttpClient httpClient;
+        private HttpClient httpClient;
 
-        public RpcClient(string url)
+        public RpcClient(string url, string rpcUser = default, string rpcPass = default)
         {
             httpClient = new HttpClient() { BaseAddress = new Uri(url) };
+            if (rpcUser != default && rpcPass != default)
```
shargon:
```suggestion
            if (!string.IsNullOrEmpty(rpcUser) && !string.IsNullOrEmpty(rpcPass))
```

src/RpcClient/RpcClient.cs
```diff
@@ -199,34 +210,39 @@ public RpcRawMemPool GetRawMempoolBoth()
...
-            return RpcSend("getrawtransaction", txid).AsString();
+            return RpcSend("getrawtransaction", txHash).AsString();
         }
 
         /// <summary>
         /// Returns the corresponding transaction information, based on the specified hash value.
         /// verbose = true
         /// </summary>
-        public RpcTransaction GetRawTransaction(string txid)
+        public RpcTransaction GetRawTransaction(string txHash)
```
shargon: vebose as argument?

src/RpcClient/RpcClient.cs
```diff
@@ -199,34 +210,39 @@ public RpcRawMemPool GetRawMempoolBoth()
...
         {
-            return RpcTransaction.FromJson(RpcSend("getrawtransaction", txid, true));
+            return RpcTransaction.FromJson(RpcSend("getrawtransaction", txHash, true));
         }
 
         /// <summary>
-        /// Returns the stored value, according to the contract script hash and the stored key.
+        /// Returns the stored value, according to the contract script hash(or Id) and the stored key.
```
shargon:
```suggestion
        /// Returns the stored value, according to the contract script hash (or Id) and the stored key.
```

src/RpcClient/RpcClient.cs
```diff
@@ -245,13 +281,41 @@ public RpcVersion GetVersion()
...
+        /// </summary>
+        public string SendRawTransaction(Transaction transaction)
+        {
+            return SendRawTransaction(transaction.ToArray());
+        }
+
+        /// <summary>
+        /// Broadcasts a serialized block over the NEO network.
+        /// </summary>
+        public string SubmitBlock(byte[] block)
```
shargon: I think that the return should be UInt256 instead of strings.

src/RpcClient/RpcClient.cs
```diff
@@ -277,35 +345,154 @@ public RpcPlugin[] ListPlugins()
...
         }
 
         /// <summary>
-        /// Verifies that the address is a correct NEO address.
+        /// Returns the balance of the corresponding asset in the wallet, based on the specified asset number.
+        /// This method applies to assets that conform to NEP-5 standards.
         /// </summary>
-        public RpcValidateAddressResult ValidateAddress(string address)
+        /// <returns>new address as string</returns>
+        public BigInteger GetBalance(string assetId)
```
shargon: BigDecimal and take the decimals from the contract?

src/RpcClient/RpcClient.cs
chenquanyu: The return types are not the same, I think using two methods is more convenient.

nicolegys:
All APIs test passed.
ContractClient               √
Nep5API                        √
PolicyAPI                       √
RpcClient                      √
TransactionManager     √
WalletAPI                      √

nicolegys:
Could we go on with this PR? Is it OK for you? @erikzhang @shargon

PR #172, opened 2020-01-03
Rpc limits (3x)

shargon:
Close https://github.com/neo-project/neo/issues/1393 for 3x

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -92,7 +92,7 @@ namespace Neo.Plugins
             host = new WebHostBuilder().UseKestrel(options => options.Listen(Settings.Default.BindAddress, Settings.Default.Port, listenOptions =>
             {
                 // Default value is unlimited
-                options.Limits.MaxConcurrentConnections = Peer.DefaultMaxConnections;
+                options.Limits.MaxConcurrentConnections = Settings.Default.MaxConcurrentConnections;
                 // Default value is 2 minutes
                 options.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(1);
                 // Default value is 30 seconds
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -91,8 +91,11 @@ namespace Neo.Plugins
         {
             host = new WebHostBuilder().UseKestrel(options => options.Listen(Settings.Default.BindAddress, Settings.Default.Port, listenOptions =>
             {
-                options.Limits.MaxConcurrentConnections = 40;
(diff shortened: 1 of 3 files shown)


shargon:
> MaxConnections default looks to be unlimited.

This was the worst problem according to our tests.

vncoelho:
Ok, maybe let the other ones as default. I think it is not worth to reduce by half for now, @shargon.

shargon:
The best way is a with a readTimeout, but I need to find a good solution for this.

superboyiii:
@shargon I think it has the same design problem with https://github.com/neo-project/neo/pull/1394 ,
we should adjust config.json of  neo-cli to open interface for setting `MaxConnections `.

shargon:
@superboyiii take a look, here is easier to configure

shargon:
With 100 it could be vulnerable to Denial Of Service attacks in regular computers

superboyiii:
Looks much reasonable now. I'll test it soon.

superboyiii:
> With 100 it could be vulnerable to Denial Of Service attacks in regular computers

@shargon  Find an issue.
At first everything went smoothly. 
1. First I tried options.Limits.MaxConcurrentConnections, make it as 10. I opened 12 groups and each group has 10 threads. Requests were sent in 1 second and only 10 responses returned. It works well.
2. Secondly, I make one of these groups keeping loop for 2 minutes. It means 10 threads will keep sending requests. After 1min and 30 sec, no reponse returned. Seems 30 sec more it cost than expected but at least it worked finally.
3. Thirdly, I returned to test step 1 again, something unexpected appeared.  Almost half of 120 requests returned(the limit I set is always 10). However after I restarted neo-cli, it worked normally again.

I think it's not from your PR but some mechanism defect somewhere. But it could decrease effectiency of PR a lot.

shargon:
> Thirdly, I returned to test step 1 again, something unexpected appeared. Almost half of 120 requests returned(the limit I set is always 10). However after I restarted neo-cli, it worked normally again.

Although it's limited to 10 concurrent, does not mean that can not serve 120 , as long as they are not concurrent they should be waiting, I think it should be the normal behavior of the server.

superboyiii:
It doesn''t means 120 requests were sent one by one.  What I did is creating 120 threads and each one sent 1 request in 1 second at the same time. Step1 return expected result，but step3 returned different result.  I just repeat step1 in step3，they're the same behaviour.

shargon:
You created 120 concurrent requests, but the server could enqueue and serve only 10 at same time, this dosen't mean that you can't send 120 concurrent request, means that it will spend more time, because some request need to wait and maybe some of them can fail.

PR #174, opened 2020-01-06
Update NuGet versions of all modules

vncoelho:
closes #173

PR #177, opened 2020-01-07
Change from public to protected as in level db

### LINKED ISSUE #173: RocksDB Building error
```cs
Plugins/Storage/RocksDBStore.cs(10,30): error CS0507: 'RocksDBStore.Configure()': cannot change access modifiers when overriding 'protected' inherited member 'Plugin.Configure()' [/opt/neo-plugins/src/RocksDBStore/RocksDBStore.csproj]
```

@shargon, do you have an idea?

vncoelho:
closes #173 

@shargon, take a look, perhaps it might be this.

### DIFF
--- a/src/RocksDBStore/Plugins/Storage/RocksDBStore.cs
+++ b/src/RocksDBStore/Plugins/Storage/RocksDBStore.cs
@@ -7,7 +7,7 @@ namespace Neo.Plugins.Storage
         /// <summary>
         /// Configure
         /// </summary>
-        public override void Configure()
+        protected override void Configure()
         {
             Settings.Load(GetConfiguration());
         }
--- a/src/RocksDBStore/RocksDBStore.csproj
+++ b/src/RocksDBStore/RocksDBStore.csproj
@@ -14,7 +14,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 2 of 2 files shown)


vncoelho:
https://github.com/neo-project/neo-modules/blob/c08ce9017ea62ef59ec69136005d40fda642537b/src/LevelDBStore/Plugins/Storage/LevelDBStore.cs#L9

shargon:
The change it's good

vncoelho:
Perfect, I will try to compile again as soon as possible.

shargon:
The problem is the neo-version

vncoelho:
The NEO version? O.o
What effectly changed from 844 to 846?

vncoelho:
image

Maybe it is not, @shargon,

that two last commits did not changed anything related.
Anyway it might be good to update to `protected`, yes?

shargon:
Now it pass the UT

vncoelho:
@shargon, perhaps we can merge this PR with #174 as @cloud8little suggested.

shargon:
I think that we should update the version of the nuget for all plugins here, what do you think?

vncoelho:
Here or in the other PR?

shargon:
We can close this, and join all in the other

PR #178, opened 2020-01-08
change network fee calculation

chenquanyu:


chenquanyu:
Hi @shargon @vncoelho 
I found the size of the transaction is 1 byte longer than the calculate method. Because the `VarSize` of Witnesses is one byte longer than the `VarSize` of verification and invocation. Is this as designed?
```
 _size = HeaderSize +
        Attributes.GetVarSize() +   //Attributes
        Cosigners.GetVarSize() +    //Cosigners
        Script.GetVarSize() +       //Script
        Witnesses.GetVarSize();     //Witnesses
```

chenquanyu:
> Looks good, but could you add some UT for check that the fee is well calculated?

Hi @shargon, I have added UT for network fee in `TestSign()`.

vncoelho:
Great job, @chenquanyu!

PR #180, opened 2020-01-15
update faster plugin

bettybao1209:
Related to PR #158 
1. upgrade fasterdb version to `2020.1.14.1`,  and update corresponding codes;
2. add serialnum(sno) to distinguish write operations between store and snapshot.


src/FasterDBStore/Store.cs
```diff
@@ -142,22 +143,33 @@ public IEnumerable<(byte[] Key, byte[] Value)> Find(byte table, byte[] prefix)
...
+
             var k = new BufferKey(table, key);
             var v = new BufferValue(value);
 
-            if (db.Upsert(ref k, ref v, Empty.Default, 0) == Status.PENDING)
+            session.Upsert(ref k, ref v, Empty.Default, 0);
+            var status = session.Read(ref k, ref input, ref output, Empty.Default, 0);
+
+            if (status == Status.PENDING)
             {
```
shargon: We don't need to check the Upsert status?

bettybao1209: we should check the status of upsert and delete, since they are the operation type that can be pending.

shargon: Before read it? two checks or with one it's enough?

bettybao1209: I think checking the read status after the upsert/delete operation should be ok

shargon:
We can merge it and improve it in the other PR

Issue #179, opened 2020-01-15
[RpcNep5Tracker]Getnep5balances, balance won't change when a transaction is not `transfer`. 


nicolegys:
Describe the bug
Getnep5balances, balance won't change when a tx is not transfer. Then, make a transfer tx, the gas balance value will return to normal.

To Reproduce
1.open wallet 6.json, deploy a smartcontract, list asset, gas balance = 8.9850761
![image](https://user-images.githubusercontent.com/43694095/72419520-f65c2c00-37b7-11ea-96bd-dd9ce8d51320.png)

2.rpc getnep5balances, gas balance didn't change, gas balance = 10
![image](https://user-images.githubusercontent.com/43694095/72419617-23a8da00-37b8-11ea-9c5e-5173025bc383.png)

3.make a transfer transaction, list asset, gas balance = 6.9723922
![image](https://user-images.githubusercontent.com/43694095/72419641-2c99ab80-37b8-11ea-9960-6d0b9e3dd38f.png)

4.rpc getnep5balances, gas balance changed, gas balance = 6.9723922
![image](https://user-images.githubusercontent.com/43694095/72419664-3ae7c780-37b8-11ea-9941-87bd005f014b.png)

nicolegys:
@Tommo-L :)

Tommo-L:
Got, we'll fix it

PR #181, opened 2020-01-18
fix balance not change for non-transfer tx

bettybao1209:
fix issue #179


src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -107,6 +107,8 @@ private void RecordTransferHistory(StoreView snapshot, UInt160 scriptHash, UInt1
             var eventName = stateItems[0].GetString();
             if (eventName != "Transfer") return;
             if (stateItems.Count < 4) return;
+            if (stateItems[1] is VM.Types.Null) stateItems[1] = null;
```
vncoelho: Better to do this conversion right on the beginning instead of using VM.Types.Null?

nicolegys:
Test passed.

src/RpcNep5Tracker/RpcNep5Tracker.cs
erikzhang: I think it is not necessary to convert it to null. We can check it by calling `stateItems[1].IsNull`.

bettybao1209: yes, I have removed it.

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -120,17 +121,19 @@ private void RecordTransferHistory(StoreView snapshot, UInt160 scriptHash, UInt1
             byte[] toBytes = stateItems[2]?.GetSpan().ToArray();
             if (toBytes?.Length != 20) toBytes = null;
```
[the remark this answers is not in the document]
bettybao1209: for gas burn case, stateItems[2] is null, and it works well.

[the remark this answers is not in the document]
shargon: https://github.com/neo-project/neo/blob/35c73cecf6538f70005fab1567367a6384b24092/src/neo/SmartContract/Native/Tokens/Nep5Token.cs#L115

I think that this notification will not pass the line 114 https://github.com/neo-project/neo-modules/pull/181/files#diff-d41367a5b8a7e11c7927addce76c2639R114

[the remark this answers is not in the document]
bettybao1209: Here the ScriptContainer is null, so notifications added in `Burn` will not be handled. But the systemfee+networkfee will be handled in the second `HandleNotification` (L164-L167).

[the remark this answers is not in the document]
erikzhang: It is not for GAS only. Any NEP-5 asset can burn. We should handle it correctly.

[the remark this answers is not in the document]
bettybao1209: For other Nep5 assets, their ScriptContainer is not null, so the notifications in Burn function can be handled in the above HandleNotification.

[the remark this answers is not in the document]
erikzhang: L112 and L114 will return in that case.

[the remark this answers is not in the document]
bettybao1209: So should I add back the coversion from VM.Types.NULL to null ?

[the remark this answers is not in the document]
erikzhang: No. You should use `StackItem.IsNull` to check it. You can't modify the parameters. You are not the only subscriber of the event.

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -97,6 +98,11 @@ private void RecordTransferHistory(StoreView snapshot, UInt160 scriptHash, UInt1
             transferIndex++;
         }
 
+        private bool IsNull(VM.Types.StackItem item)
+        {
+            return item == null || item.IsNull;
```
erikzhang: I think it is impossible for it to be `null`.

shargon: Yes, we never send `null` as state

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -156,6 +159,12 @@ public void OnPersist(StoreView snapshot, IReadOnlyList<Blockchain.ApplicationEx
                     HandleNotification(snapshot, transaction, notifyEventArgs.ScriptHash, stateItems,
                         nep5BalancesChanged, ref transferIndex);
                 }
+
+                // Handle SystemFee and NetworkFee GAS burn for each transaction
+                if (!(appExecuted.Transaction is Transaction tx)) continue;
```
erikzhang: What does this mean?

bettybao1209: To deal with the fees for those non-null transactions.

erikzhang: `appExecuted.Transaction is null`?

bettybao1209: image

For each block persisted, here the ScriptContainer is null, so the ApplicationExecuted.Transaciton is null.

superboyiii:
@bettybao1209 @erikzhang Please go on this PR, based on our test, it does fix this bug.

src/RpcNep5Tracker/RpcNep5Tracker.cs
erikzhang: I mean, you should modify it to `appExecuted.Transaction is null`.

nicolegys:
Throw exception when starting neo-cli.
image

version of RpcNep5tracker `ef4b506c556498d78ebdebf9ef84deb728ef935a`

bettybao1209:
> Throw exception when starting neo-cli.
> image
> 
> version of RpcNep5tracker `ef4b506c556498d78ebdebf9ef84deb728ef935a`

Now it passed my test cases locally. Plz check.

nicolegys:
Thank you @bettybao1209 
Now it can run normaly and the test passed.
There is only one small point, the consensus address will get net fee after packaging a tx, but giving it the net fee is not a tx, so the nep5balance(gas) of it will not be refreshed. I think it's resonable.

shargon:
> There is only one small point, the consensus address will get net fee after packaging a tx, but giving it the net fee is not a tx, so the nep5balance(gas) of it will not be refreshed. I think it's resonable.

The genesis is not computed? Because the initialization of NEO and GAS call the Transfer event.

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -156,6 +159,12 @@ public void OnPersist(StoreView snapshot, IReadOnlyList<Blockchain.ApplicationEx
                     HandleNotification(snapshot, transaction, notifyEventArgs.ScriptHash, stateItems,
                         nep5BalancesChanged, ref transferIndex);
                 }
+
+                // Handle SystemFee and NetworkFee GAS burn for each transaction
+                if (appExecuted.Transaction is null) continue;
+                HandleNotification(snapshot, appExecuted.Transaction, NativeContract.GAS.Hash,
+                    new VM.Types.Array(new VM.Types.StackItem[] { "Transfer", appExecuted.Transaction.Sender.ToArray(), VM.Types.StackItem.Null, appExecuted.Transaction.SystemFee + appExecuted.Transaction.NetworkFee }),
+                    nep5BalancesChanged, ref transferIndex);
```
erikzhang: I think we should not handle the burn events for each transaction, but for the `SYSTEM` trigger only.

https://github.com/neo-project/neo/blob/c1b9b9712ee638768014150e2369ac76e02863db/src/neo/Ledger/Blockchain.cs#L504

bettybao1209: Here the ScriptContainer (transaction) of the ApplicationEngine with the SYSTEM trigger is null.
https://github.com/neo-project/neo/blob/c1b9b9712ee638768014150e2369ac76e02863db/src/neo/Ledger/Blockchain.cs#L499.
The reason we trace each transaction is to refresh the sender's balance and the transfer history.

bettybao1209: I wonder whether we should record the transfer history for non-tranfer transactions. If not, maybe we should just loose the restrictions on `notifyEventArgs.ScriptContainer is Transaction transaction`, just to trace the balance change.
https://github.com/neo-project/neo-modules/blob/c75c0766fc15832fe3cd5edce4c3bb4b68e56253/src/RpcNep5Tracker/RpcNep5Tracker.cs#L156

nicolegys:
Test passed.
The value of `nep5balances` was consistent with what in neo-cli, and the `nep5transfers` for non-tranfer transactions(e.g. paying netfee and sysfee) were not recorded.

superboyiii:
@erikzhang Could you make a final review? Is this OK for you?

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -108,33 +109,38 @@ private void RecordTransferHistory(StoreView snapshot, UInt160 scriptHash, UInt1
...
                 return;
-            byte[] fromBytes = stateItems[1]?.GetSpan().ToArray();
-            if (fromBytes?.Length != 20) fromBytes = null;
-            byte[] toBytes = stateItems[2]?.GetSpan().ToArray();
-            if (toBytes?.Length != 20) toBytes = null;
+            byte[] fromBytes = stateItems[1].IsNull ? null : stateItems[1].GetSpan().ToArray();
+            if (fromBytes?.Length != UInt160.Length) fromBytes = null;
+            byte[] toBytes = stateItems[2].IsNull ? null : stateItems[2].GetSpan().ToArray();
+            if (toBytes?.Length != UInt160.Length) toBytes = null;
             if (fromBytes == null && toBytes == null) return;
```
shargon: If someone send `new (byte[50],new byte[20])` it will be treated as `(null,new byte[20])` We should distinguish in L120 between a format error and `null`.

bettybao1209: so maybe we should just `return` if the address format is invalid?

shargon: I think so

nicolegys:
OK for me now, @erikzhang is it OK for you?

PR #183, opened 2020-02-07
Fix RpcServer[GetStorage]

superboyiii:
Reference: https://github.com/neo-project/neo/pull/1400  Using id to be the key.


shargon:
We need to fix the `dotnet-format` errors, I created https://github.com/neo-project/neo-modules/pull/185 for tests it, without luck

belane:
Would this change the rpc call, requiring consumers to know the index assigned to each smart contract? It may be better to maintain queries by scripthash and convert it internally.

superboyiii:
@shargon Your PR looks much more reasonable now. And I also had a little advice here:https://github.com/neo-project/neo-node/issues/540
Since id is unique for each sc on blockchain, it's good when make contract.update. We need to record and remember it. Maybe we could make [getstorage] support both id and scripthash as param[0].

superboyiii:
Test Step:
1. deploy a nep-5 contract and invoke `deploy` method.
2. Get contract id via RPC `getcontractstate` .
3. Make id as param[0], and hex of totalSupply ("746f74616c537570706c79") as param[1] via PRC `getstorage`. (expect pass)
4. Make scriptHash of sc as param[0], and hex of totalSupply ("746f74616c537570706c79") as param[1] via PRC `getstorage`. (expect pass)
5. Make "aa" as param[0], and hex of totalSupply ("746f74616c537570706c79") as param[1] via PRC `getstorage`. (expect fail)
6. Make -1 as param[0], and hex of totalSupply ("746f74616c537570706c79") as param[1] via PRC `getstorage`. (expect pass [return null])

Test Result: [**PASS**]

Issue #195, opened 2020-03-18
[RpcNep5Tracker] Using contract id instead of contract scriptHash.


nicolegys:
Now the plugin uses contract scriptHash to creat a key in storage.
![image](https://user-images.githubusercontent.com/43694095/76944298-bcd0aa00-693b-11ea-9eb3-d49f9fd02b75.png)
Shall we use contract id instead? Since now we support migrating/destroying a contract, the scriptHash could be changed/deleted. 
Maybe we also need to handle notifications other than `transfer`, to handle the migration/destroy scenes.

shargon:
Yes, we should do it

ProDog:
This can be reopened?  the  scriptHash could not be changed if contract migrated. @Tommo-L

Tommo-L:
Yes, plz reopen it.

PR #196, opened 2020-03-25
Deal with contract destory

bettybao1209:
partly close https://github.com/neo-project/neo-modules/issues/195

### DIFF
--- a/src/RpcNep5Tracker/RpcNep5Tracker.cs
+++ b/src/RpcNep5Tracker/RpcNep5Tracker.cs
@@ -286,6 +286,11 @@ namespace Neo.Plugins
             foreach (var (key, value) in dbCache.Find(prefix))
             {
                 JObject balance = new JObject();
+                if (Blockchain.Singleton.View.Contracts.TryGet(key.AssetScriptHash) is null)
+                {
+                    dbCache.Delete(key);
+                    continue;
+                }
                 balance["asset_hash"] = key.AssetScriptHash.ToString();
                 balance["amount"] = value.Balance.ToString();
                 balance["last_updated_block"] = value.LastUpdatedBlock;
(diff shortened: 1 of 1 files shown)


erikzhang:
You shouldn't modify the database when querying.

erikzhang:
https://github.com/neo-project/neo-modules/blob/bbcfa56171a931c1c6f2b50f9289350dd7c052c0/src/RpcNep5Tracker/RpcNep5Tracker.cs#L148

You can delete them in `OnPersist()`.

bettybao1209:
> You can delete them in `OnPersist()`.
But I think it is hard to delete records for non-owner address.

nicolegys:
Thank you @bettybao1209 

Test partly passed.
After the contract being destroyed, getnep5balances won't return its nep5 asset.

Steps:
1.deploy a nep5 contract.
2.make some txs and then getnep5balances.
image
3.destroy the contract.
image
4.getnep5balances
image
The asset was deleted from the list.

Some concerns:
1.Now after the nep5 contract is destroyed, it will not be listed when calling `getnep5balances`, but its storage items still exist.
Once the owner deploys the same contract again, when we call `getnep5balances`, it's OK for the owner address, because the value will be covered, but for the non-owner addresses, the old value will be listed until the address have a `transfer` tx to cover the old value.
I know the probability for someone to re-deploy his destroyed contract is quite low, but maybe we should find a way to delete the storage item in the future.
2.Also, after the nep5 contract is migrated, its "asset_hash" won't be refreshed when calling `getnep5balances`. Every holder address must make a `transfer` tx to refresh its "asset_hash". 
I think we can find a better solution in the future.

src/RpcNep5Tracker/RpcNep5Tracker.cs
```diff
@@ -286,6 +286,8 @@ public JObject GetNep5Balances(JArray _params)
             foreach (var (key, value) in dbCache.Find(prefix))
             {
                 JObject balance = new JObject();
+                if (Blockchain.Singleton.View.Contracts.TryGet(key.AssetScriptHash) is null)
+                    continue;
```
shargon: Maybe we shoud do that only in the first iteration, a break or a flag should be ok

bettybao1209: But we should list all assets in user's address, so we can't just break the iteration.

PR #197, opened 2020-03-25
Fix verifyresult on RPCServer

vncoelho:


erikzhang:
This is incorrect. You need to monitor the event stream to receive the `RelayResult` messages.

PR #199, opened 2020-03-26
Remove GetBlockSysFee

bettybao1209:


erikzhang:
Should `RpcClient` be updated too?

bettybao1209:
> Should `RpcClient` be updated too?

Yes, I just removed it.

vncoelho:
But we need a call for getting the burned amount.

Issue #198, opened 2020-03-25
Unhandled exception when using Plugin RpcServer


vncoelho:
```cs
neo> Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.AspNetCore.ResponseCompression, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. The system cannot find the file specified.

File name: 'Microsoft.AspNetCore.ResponseCompression, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
   at Neo.Plugins.RpcServer.<>c.<OnPluginsLoaded>b__20_2(IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.<>c__DisplayClass10_0.<ConfigureServices>b__0(WebHostBuilderContext _, IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.BuildCommonServices(AggregateException& hostingStartupErrors)
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()
   at Neo.Plugins.RpcServer.OnPluginsLoaded()
   at Neo.NeoSystem..ctor(String storageEngine) in /opt/neoLib/src/neo/NeoSystem.cs:line 39
   at Neo.CLI.MainService.Start(String[] args)
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_1(Object state)
   at System.Threading.QueueUserWorkItemCallbackDefaultContext.Execute()
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading._ThreadPoolWaitCallback.PerformWaitCallback()
```

shargon:
We will need dependecies in plugins

vncoelho:
Do you know which dependencies to install, @shargon?
Currently this problem is blocking the use on Ubuntu Bionic.

vncoelho:
I read about adding `<PackageReference Include="Microsoft.AspNetCore.App" />`, however, I had problem of compatibility with `netstandard2.1`.

shargon:
It should be `Microsoft.AspNetCore.ResponseCompression`

vncoelho:
[...]

PR #200, opened 2020-03-27
Add relayresult

bettybao1209:


src/RpcServer/RpcServer.Node.cs
```diff
@@ -70,16 +74,50 @@ private JObject GetVersion(JArray _params)
...
             Block block = _params[0].AsString().HexToBytes().AsSerializable<Block>();
-            VerifyResult reason = System.Blockchain.Ask<VerifyResult>(block).Result;
-            return GetRelayResult(reason, block.Hash);
+            return Send(block);
+        }
+
+        private JObject Send(IInventory inventory)
+        {
+            var a = System.ActorSystem.ActorOf(RpcActor.Props());
+            System.ActorSystem.EventStream.Subscribe(a, typeof(RelayResult));
```
erikzhang: This can be a static variable.

bettybao1209: Fixed, please review again.

src/RpcServer/RpcServer.cs
```diff
@@ -16,17 +17,43 @@
...
         private IWebHost host;
+        private static readonly IActorRef rpcActor = System.ActorSystem.ActorOf(RpcActor.Props());
 
         public RpcServer()
         {
             RegisterMethods(this);
+            System.ActorSystem.EventStream.Subscribe(rpcActor, typeof(RelayResult));
+        }
+
+        private class RpcActor : UntypedActor
```
erikzhang: One file one class.

bettybao1209: Fixed.

src/RpcServer/RelayActor.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            switch (message)
+            {
+                case IInventory inventory:
+                    this.inventory = inventory;
+                    this.sender = Sender;
+                    blockchain.Tell(inventory);
+                    break;
+                case RelayResult reason:
+                    if (reason.Inventory.Hash.Equals(inventory.Hash))
```
chenquanyu: The inventory may change when another request comes before the first return. Maybe we can maintain a `Dictionary<UInt256, RelayResult>` to avoid this error.

erikzhang: You are right! There should be a dictionary.

vncoelho:
close #205

vncoelho:
I am facing problem to test RpcServer because of the dependencies problem reported on #198, could you give a hand on that @bettybao1209 @superboyiii ?

src/RpcServer/RelayActor.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                case IInventory inventory:
+                    {
+                        senders.Add(inventory.Hash, Sender);
+                        neoSystem.Blockchain.Tell(inventory);
+                        break;
+                    }
+                case RelayResult reason:
+                    {
+                        if (senders.Remove(reason.Inventory.Hash, out var actor))
```
shargon: If we never receive this message, the size of `sender` will increase without limit.

nicolegys:
Test passed.

Test steps:

 **`sendrawtransaction`**
1. `sendrawtransaction` - tx not exsited
result: return tx hash
image
2. `sendrawtransaction` - tx already exsited
result: return error `AlreadyExists`
image

 **`submitblock`**
1.show state
image
2. `submitblock` - block not exsited
result: return block hash
image
3.show state
result: block height increased by 1
image
4. `submitblock` - block already exsited
result: return error `AlreadyExists`
image
5.show state
result: block height didn't change
image

src/RpcServer/RelayActor.cs
erikzhang: Why do we not receive this message?

shargon: According to the Akka documentation (https://doc.akka.io/docs/akka/current/general/message-delivery-reliability.html)
```
local sends can fail in Akka-specific ways:

if the mailbox does not accept the message (e.g. full BoundedMailbox)
if the receiving actor fails while processing the message or is already terminated
```

Also, this it's a module, so it's easy to forget if we make any change in the core.

erikzhang: Any solution.

chenquanyu: Maybe we can add expire time for this dictionary.

src/RpcServer/RelayActor.cs
```diff
@@ -0,0 +1,55 @@
...
+            switch (message)
+            {
+                case IInventory inventory:
+                    {
+                        UInt256 hash = inventory.Hash;
+                        senders.Add(hash, Sender);
+                        neoSystem.Blockchain.Tell(inventory);
+                        // the sender will clean the record after expireMs
+                        Task.Run(async () =>
+                        {
```
shargon, line 33: I prefer only one task and check the whole list every X seconds, because one task per inventory could be slower.

shargon: Or maybe we can have a list with max X items, and trim it if we exceed this limit. I am doing this class for oracles

bettybao1209: please review it again.

src/RpcServer/FixedDictionary.cs
```diff
@@ -0,0 +1,51 @@
+using System.Collections.Generic;
+
+namespace Neo.Plugins
+{
+    public class FixedDictionary<TKey, TValue>
```
erikzhang, line 5: Is this the `FIFOCache`?

https://github.com/neo-project/neo/blob/master/src/neo/IO/Caching/FIFOCache.cs

bettybao1209: FIFOCache is defined as internal,I'm afraid we cannot directly use it in rpcserver.

erikzhang: Please consider using KeyedCollection.

https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.keyedcollection-2

bettybao1209: Already done, please review it again.

tests/Neo.Network.RPC.Tests/Neo.Network.RPC.Tests.csproj
```diff
@@ -17,6 +17,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\src\RpcClient\RpcClient.csproj" />
+    <ProjectReference Include="..\..\src\RpcServer\RpcServer.csproj" />
```
erikzhang, line 20: Can you create a new UT project for RpcServer?

erikzhang:
Please wait for neo-project/neo#1548

tests/Neo.Network.RPC.Tests/Neo.Network.RPC.Tests.csproj
erikzhang: The name should be `Neo.Plugins.RpcServer.Tests`.

bettybao1209: Yes, my latest commit has done this.

erikzhang: Yours is `Neo.Plugin.RpcServer.Tests`.

bettybao1209: sorry for that, fixed already.

erikzhang:
neo-project/neo#1548 has been merged.

bettybao1209:
since https://github.com/neo-project/neo/pull/1548 is merged, maybe we can use https://github.com/neo-project/neo-modules/pull/220 instead.

bettybao1209:
> [neo-project/neo#1548](https://github.com/neo-project/neo/pull/1548) has been merged.

Ok, I open a new PR https://github.com/neo-project/neo-modules/pull/220 and I'll close this one.

PR #201, opened 2020-03-30
Allow different cases of transfer

bettybao1209:


erikzhang:
Event names must be capitalized to distinguish them from method names.

bettybao1209:
So should we emphasis this point in NEP5 and NEP14?

vncoelho:
I think so, @bettybao1209, as better described there better will it be.

PR #206, opened 2020-03-30
Enable programmatic configuration of RpcServer plugin

### LINKED ISSUE #203: RpcServer plugin needs to be programmatically configurable
neo-express will need to programmatically create and configure the RpcServer plugin w/o depending on config.json file.

devhawk:
This PR modifies the RpcServer plugin so that it can be used and configured independently of the disk-based plugin infrastructure of neo-cli. Neo-express does not use the same file based plugin configuration that neo-cli does, yet needs to be able to configure and run some of the standard plugins such as RpcServer.

In particular, this PR:
* separates plugin load functionality out into separate RpcServerPlugin class
* eliminates shared settings state in favour of injecting settings via constructor parameter
* enabled construction of RpcServerSettings instance w/o requiring IConfigurationSection

fixes #203

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -86,7 +86,7 @@ namespace Neo.Plugins
             }
         }
 
-        public void StartHost()
+        public void StartRpcServer()
         {
             host = new WebHostBuilder().UseKestrel(options => options.Listen(settings.BindAddress, settings.Port, listenOptions =>
             {
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -77,7 +77,7 @@ namespace Neo.Plugins
             return response;
         }
 
-        public  void Dispose()
+        public void Dispose()
         {
             if (host != null)
             {
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -19,19 +19,24 @@ using System.Threading.Tasks;
 
 namespace Neo.Plugins
 {
-    public sealed partial class RpcServer : Plugin
+    public sealed partial class RpcServer : IDisposable
     {
-        private static readonly Dictionary<string, Func<JArray, JObject>> methods = new Dictionary<string, Func<JArray, JObject>>();
-        private IWebHost host;
+        private readonly Dictionary<string, Func<JArray, JObject>> methods = new Dictionary<string, Func<JArray, JObject>>();
 
-        public RpcServer()
+        private IWebHost host;
+        private readonly RpcServerSettings settings;
+        private readonly NeoSystem system;
+
+        public RpcServer(NeoSystem system, RpcServerSettings settings)
         {
+            this.system = system;
+            this.settings = settings;
             RegisterMethods(this);
         }
 
         private bool CheckAuth(HttpContext context)
         {
-            if (string.IsNullOrEmpty(Settings.Default.RpcUser)) return true;
+            if (string.IsNullOrEmpty(settings.RpcUser)) return true;
 
             context.Response.Headers["WWW-Authenticate"] = "Basic realm=\"Restricted\"";
 
@@ -50,12 +55,7 @@ namespace Neo.Plugins
             if (authvalues.Length < 2)
                 return false;
 
-            return authvalues[0] == Settings.Default.RpcUser && authvalues[1] == Settings.Default.RpcPass;
-        }
-
-        protected override void Configure()
-        {
-            Settings.Load(GetConfiguration());
+            return authvalues[0] == settings.RpcUser && authvalues[1] == settings.RpcPass;
         }
 
         private static JObject CreateErrorResponse(JObject id, int code, string message, JObject data = null)
@@ -77,9 +77,8 @@ namespace Neo.Plugins
             return response;
         }
 
-        public override void Dispose()
+        public  void Dispose()
         {
-            base.Dispose();
             if (host != null)
             {
                 host.Dispose();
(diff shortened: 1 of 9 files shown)


vncoelho:
Sounds good, @devhawk.
I will take a look asap.

Can you give a basic guidelines for reproducing how to configure it now?

devhawk:
> Can you give a basic guidelines for reproducing how to configure it now?

For neo-cli, it still loads like as plugin so there is no difference. for neo-express, I'm loading it like this:

``` csharp
using var system = new NeoSystem(storageEngineName);
var rpcSettings = new RpcServerSettings(port: rpcPort);
var rpcServer = new RpcServer(system, rpcSettings);
rpcServer.StartHost(); // this isn't a great name. Maybe it should be StartRpcServer?
system.StartNode(channelsConfig);
system.StartConsensus(wallet);
```

src/RpcServer/RpcServer.cs
```diff
@@ -77,39 +77,38 @@ private static JObject CreateResponse(JObject id)
             return response;
         }
 
-        public override void Dispose()
+        public  void Dispose()
```
shargon:
```suggestion
        public void Dispose()
```

src/RpcServer/RpcServerPlugin.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    public sealed class RpcServerPlugin : Plugin
+    {
+        static List<object> handlers = new List<object>();
+        RpcServer server;
+        RpcServerSettings settings;
+
+        protected override void Configure()
+        {
+            settings = new RpcServerSettings(GetConfiguration());
```
shargon: We need to update the settings in `server` when this method was called. (Auto-reconfiguration)

src/RpcServer/RpcServerSettings.cs
```diff
@@ -19,9 +20,32 @@ internal class Settings
...
 
-        public static Settings Default { get; private set; }
+        public RpcServerSettings(IPAddress bindAddress = null,
+            ushort port = 10332,
+            string sslCert = "",
+            string sslCertPassword = "",
+            string[] trustedAuthorities = null,
+            string rpcUser = "",
+            string rpcPass = "",
+            string masGasInvoke = "10",
```
shargon: fees as `BigDecimal`?

src/RpcServer/RpcServerSettings.cs
```diff
@@ -19,9 +20,32 @@ internal class Settings
         public string[] DisabledMethods { get; }
         public int MaxConcurrentConnections { get; }
 
-        public static Settings Default { get; private set; }
+        public RpcServerSettings(IPAddress bindAddress = null,
```
shargon: I am not very fan of constructors with all the arguments, I prefer to set the properties manually with the `get` and `set`

devhawk: I designed it this way so that the properties would continue to be read only, but I'm fine with changing it. I will point out that all the ctor args have defaults, which makes it easy to specify only the non-default fields via named arguments.

``` csharp
// all the fields except port use the defaults
var rpcSettings = new RpcServerSettings(port: rpcPort);
```

src/RpcServer/RpcServerSettings.cs
devhawk: I was a little worried about that since the original ctor needs to know NativeContract.GAS.Decimals, but yeah you're right these should be BigDecimals. Will fix

src/RpcServer/RpcServerPlugin.cs
devhawk: would it make sense to separate the settings that can be updated from those that cannot? Several of the settings (SslCert, SslCertPassword, MaxConcurrentConnections, TrustedAuthorities) are only used when creating the WebHost instance. If the user changes the RPC port after the system is running, the running WebHost port is not torn down and re-created.

src/RpcServer/RpcServerSettings.cs
devhawk: Actually, it would make sense to move to R/W properties in order to support auto-reconfiguration - at least for the settings that can be modified while running (RpcUser, RpcPass, MaxGasInvoke, MaxFee, DisabledMethods)

src/RpcServer/RpcServerPlugin.cs
devhawk: fixed

src/RpcServer/RpcServerSettings.cs
devhawk: I changed maxGasInvoke and maxFee ctor params to be decimals.

nicolegys:
I tried to update the settings, such as `DisabledMethods`, but not everytime succeeded, sometimes the neo-cli would throw exception and crash.
Then I tried to debug, found out that the exception was threw in `GetConfiguration()`
image
It seems that here the json file wasn't added successfully in time, because if I catch the exception and try addjsonfile again, it will pass, and then the settings updated successfully.
image

devhawk:
> I tried to update the settings, such as `DisabledMethods`, but not everytime succeeded, sometimes the neo-cli would throw exception and crash.
> Then I tried to debug, found out that the exception was threw in `GetConfiguration()`

Thanks for testing this PR @nicolegys. in the core Neo.dll. Master branch version of RpcServer calls `GetConfiguration()` in a similar fashion. Would you mind checking to see if we see a similar problem using the master branch version of RpcServer? That would help us figure out if the issue is in this PR or somewhere else.

nicolegys:
You're right @devhawk , there existed similar problem when using the master branch version of RpcServer. I'll create an issue in neo repository.

nicolegys:
1.smoke test passed.
2.update settings failed, because of https://github.com/neo-project/neo/issues/1531

Version
neo-cli: 5a46401b3635b0f2f6a6b09141c20ea5ab361623
neo: 3.0.0-CI00863 (I didn't use the latest one because we need https://github.com/neo-project/neo-modules/pull/200 to make `RpcServer` be compatible with the latest neo)

nicolegys:
@devhawk This PR is OK for me, the hot reload problem will be fixed in neo code.

shargon:
Good change!

devhawk:
> Good change!

So do you merge this @shargon or should I?

nicolegys:
@erikzhang @shargon @vncoelho Could we merge this PR?

shargon:
There are conflicts

devhawk:
> There are conflicts

Fixed

Issue #209, opened 2020-04-01
RpcClient needs to support async versions of RPC wrapper methods 


devhawk:
For example, WalletApi.Transfer needs a WalletApi.TransferAsync counterpart

chenquanyu:
We can change all the methods in `RpcClient` to `async`, but now the methods in `Neo` are `sync`. Shall we keep both versions of them? @erikzhang @shargon

shargon:
I don't like async methods :P, I prefer both, but not only async

devhawk:
any async call can be waited synchronously via the Task.Result property. Since HttpClient doesn't provide sync methods, RpcClient.Send method [uses Result property](https://github.com/neo-project/neo-modules/blob/master/src/RpcClient/RpcClient.cs#L83) to make it a synchronous call. 

Instead of parallel sync/async implementations, wouldn't it make more sense to have a single set of async methods? Folks that want to synchronously wait on the result can simply use the Result property.

PR #213, opened 2020-04-03
add multi-sig-transfer

### LINKED ISSUE #210: WalletAPI.Transfer does not work for sending from genesis account
Attempting to send 1000 neo from genesis account to a standard address using a single node Neo3 chain (master branch). The consensus node wallet has two accounts, the node-specific account (`NU78azX2nqbuDPthu5UGMaCXTVwpAkpvye`) and the multi-sig account (`NfC9EW6UJEbfX4TKCnqGf72AKHXUfR71ma`). I am calling GetKey() on the multi-sig WalletAccount and passing it to WalletApi.Transfer. Transfer throws an exception `Insufficient GAS in address: NU78azX2nqbuDPthu5UGMaCXTVwpAkpvye`. That is the node-specific account address, NOT the multi-sig account address.

Stepping thru the transfer function, the exception is throwing in the TransactionManager.Sign() function. Sign is checking the gas balance, finding the gas balance to be zero and throwing.

### LINKED ISSUE #211: Utility.ToBigInteger should be public

### LINKED ISSUE #223: TransactionManager.MakeTransaction throws exception
Attempting to implement [transfer from multi-sig account](https://github.com/neo-project/docs/blob/release-neo3/docs/en-us/tooldev/sdk/transaction.md#constructing-a-transaction-to-transfer-from-multi-signature-account). TransactionManager.MakeTransaction throws an exception: `Exception thrown: 'Neo.Network.RPC.RpcException' in RpcClient.dll: 'Unable to cast object of type 'Neo.IO.Json.JString' to type 'Neo.IO.Json.JArray'.'`

It appears that RpcClient.InvokeScript is encoding the parameters incorrectly.  [RpcServer `invokescript`](https://github.com/neo-project/neo-modules/blob/d62b18bcea99186614c3043603fcaa8e6214ba77/src/RpcServer/RpcServer.SmartContract.cs#L89) looks like it's expecting CheckWitnessHashes to be an array of it's own while RpcClient.InvokeScript is simply appending the witness hashes to the parameter array

chenquanyu:
fix #211 
fix  #210 
fix #223


src/RpcClient/TransactionManager.cs
```diff
@@ -84,8 +84,6 @@ public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[]
                 long remainder = Tx.SystemFee % d;
                 if (remainder > 0)
                     Tx.SystemFee += d - remainder;
-                else if (remainder < 0)
```
shargon, line 87: Why this change?

chenquanyu: image
This case can never be met for the `Tx.SystemFee > 0`.

src/RpcClient/WalletAPI.cs
```diff
@@ -156,6 +157,23 @@ public Transaction Transfer(UInt160 scriptHash, KeyPair from, UInt160 to, BigInt
...
+        /// Transfer NEP5 token from multi-sig account
+        /// </summary>
+        /// <param name="scriptHash">contract script hash</param>
+        /// <param name="m">multi-sig min signature count</param>
+        /// <param name="pubKeys">multi-sig pubKeys</param>
+        /// <param name="keys">sign keys</param>
+        /// <param name="to">to account</param>
+        /// <param name="amountInteger">transfer amount</param>
+        /// <returns></returns>
+        public Transaction Transfer(UInt160 scriptHash, int m, ECPoint[] pubKeys, KeyPair[] keys, UInt160 to, BigInteger amountInteger)
```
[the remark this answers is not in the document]
chenquanyu: Yes, we may wrap async methods for code reuse. https://github.com/chenquanyu/neo-modules/blob/add-async-methods/src/RpcClient/RpcClient.Async.cs

src/RpcClient/TransactionManager.cs
```diff
@@ -150,6 +148,21 @@ public TransactionManager AddMultiSig(KeyPair key, int m, params ECPoint[] publi
...
+        /// Add Multi-Signature
+        /// </summary>
+        /// <param name="keys">The KeyPairs to sign transction</param>
+        /// <param name="m">The least count of signatures needed for multiple signature contract</param>
+        /// <param name="publicKeys">The Public Keys construct the multiple signature contract</param>
+        public TransactionManager AddMultiSig(KeyPair[] keys, int m, params ECPoint[] publicKeys)
+        {
+            foreach (var key in keys)
+            {
+                AddMultiSig(key, m, publicKeys);
```
shargon, line 161: We should avoid the line 146 (https://github.com/neo-project/neo-modules/pull/213/files#diff-b0495f29c2f433fb99528ffde111175eR146), I think that it's better to duplicate the logic here.

src/RpcClient/Nep5API.cs
```diff
@@ -117,5 +118,32 @@ public Transaction CreateTransferTx(UInt160 scriptHash, KeyPair fromKey, UInt160
...
+        /// <param name="scriptHash">contract script hash</param>
+        /// <param name="m">multi-sig min signature count</param>
+        /// <param name="pubKeys">multi-sig pubKeys</param>
+        /// <param name="fromKeys">sign keys</param>
+        /// <param name="to">to account</param>
+        /// <param name="amount">transfer amount</param>
+        /// <returns></returns>
+        public Transaction CreateTransferTx(UInt160 scriptHash, int m, ECPoint[] pubKeys, KeyPair[] fromKeys, UInt160 to, BigInteger amount)
+        {
+            if (m > fromKeys.Length) throw new System.Exception($"Need at least {m} KeyPairs for signing!");
```
shargon, line 134: || m<0 ?

chenquanyu: Maybe this is also an ArgumentExcception.

image

src/RpcClient/TransactionManager.cs
chenquanyu: More keys can be added after this method, so I didn't check keys count in this method.

src/RpcClient/TransactionManager.cs
```diff
@@ -84,8 +84,6 @@ public TransactionManager MakeTransaction(byte[] script, TransactionAttribute[]
                 long remainder = Tx.SystemFee % d;
                 if (remainder > 0)
                     Tx.SystemFee += d - remainder;
```
shargon, line 86: I think that now we can remove the `remainder`

chenquanyu: Done by #219

src/RpcClient/WalletAPI.cs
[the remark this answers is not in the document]
devhawk: The underlying HttpClient doesn't expose sync versions of Get, Post, etc since Http requests are inherently async. Shouldn't all the RpcWrapper methods follow the same strategy? Why are there sync methods for any of these RPC wrapper calls?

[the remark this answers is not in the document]
chenquanyu: I think we can add async methods after the discussion in #209.

ProDog:
There are some exceptions when `SendRawTransaction`, maybe need update `CreateTransferTx`?

image

chenquanyu:
Tested on the preview2 version.
image

chenquanyu:
The signature verification failed after https://github.com/neo-project/neo/commit/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101

ProDog:
> The signature verification failed after [neo-project/neo@bf16df9](https://github.com/neo-project/neo/commit/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101)

Only use default `"Magic": 5195086` can work.

shargon:
Did you use the same `protocol.json` for sign?

Default `Magic` it's loaded from this file

https://github.com/neo-project/neo/commit/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101#diff-dd6fe58934e6f742301fccc0758b474fR13

https://github.com/neo-project/neo/blob/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101/src/neo/ProtocolSettings.cs#L36

ProDog:
RpcClient always uses the default `Magic` by reference, but test net or private net not necessarily used default `Magic`.

shargon:
You need to test it with the same `protocol.json` of the remote server

ProDog:
> You need to test it with the same `protocol.json` of the remote server

Also need `protocol.json`  in RpcClient? :worried:

shargon:
Yes, because the signature must be network dependant.

ProDog:
RpcServer's `invokefunction` has the same problem, how to modify its `Magic` :thinking:

chenquanyu:
RpcServer use the settings of CLI @ProDog

PR #217, opened 2020-04-08
Fix RPCServer configuration falure

Qiao-Jin:
Close [#1531](https://github.com/neo-project/neo/issues/1531)

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -3,6 +3,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.ResponseCompression;
 using Microsoft.AspNetCore.Server.Kestrel.Https;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Neo.IO;
 using Neo.IO.Json;
@@ -15,6 +16,7 @@ using System.Net.Security;
 using System.Reflection;
 using System.Security.Cryptography.X509Certificates;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Neo.Plugins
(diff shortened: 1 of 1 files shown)


src/RpcServer/RpcServer.cs
```diff
@@ -55,7 +57,22 @@ private bool CheckAuth(HttpContext context)
...
+                try
+                {
+                    config = GetConfiguration();
+                }
+                catch (FormatException)
+                {
+                    remainingTimes--;
+                    Thread.Sleep(10);
+                }
+            }
```
erikzhang, line 73: I think we can just catch the exception and ignore it.

Qiao-Jin: If we ignore the exception without try once more, a null config will be loaded, which will give rise to a null-pointer exception and crash the node

erikzhang: If you don't invoke `Settings.Load()`, then it won't.

Qiao-Jin: Got it, have changed correspondingly

erikzhang:
Please test it. If it works, we can update all the projects.

shargon:
Without any log it's hard to know when something was wrong, you can think that it's well configured and it's not

Qiao-Jin:
> Without any log it's hard to know when something was wrong, you can think that it's well configured and it's not

I'm not sure whether we need to log it, @erikzhang could you give us some advice?
https://github.com/neo-project/neo/issues/1531#issuecomment-610233266

erikzhang:
https://github.com/neo-project/neo/blob/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101/src/neo/Plugins/Plugin.cs#L69

How about we catch exceptions here? @Qiao-Jin @shargon

Qiao-Jin:
> https://github.com/neo-project/neo/blob/bf16df9d5a090dfcaff90d7c9d5454cf5ad09101/src/neo/Plugins/Plugin.cs#L69
> 
> How about we catch exceptions here? @Qiao-Jin @shargon

I think it's good to catch exceptions here

shargon:
The code looks well but I prefer to know get the error, otherwise it's impossible to know where it fail.

PR #222, opened 2020-04-15
Fix SystemLog

erikzhang:


shargon:
I think that both solutions are fine, this is the protection and https://github.com/neo-project/neo/pull/1571 it's the fix

src/SystemLog/Logger.cs
```diff
@@ -39,6 +40,9 @@ void ILogPlugin.Log(string source, LogLevel level, string message)
 
                 if (!string.IsNullOrEmpty(Settings.Default.Path))
                 {
+                    StringBuilder sb = new StringBuilder(source);
+                    foreach (char c in GetInvalidFileNameChars())
+                        sb.Replace(c, '-');
```
Tommo-L, line 45: Throw exception or fix in neo-core? I think we shouldn't modify the `source`, which is the log file name. 🤔

src/SystemLog/Logger.cs
```diff
@@ -39,6 +40,9 @@ void ILogPlugin.Log(string source, LogLevel level, string message)
...
                     var path = Combine(Settings.Default.Path, source);
```
shargon, line 46:
```suggestion
                    var path = Combine(Settings.Default.Path, sb.ToString());
```

src/SystemLog/Logger.cs
shargon: I think that we should merge both

erikzhang: I think we should only merge this one. `source` is not filename. We use it as filename in SystemLog, so we fix it in SystemLog.

Tommo-L: The problem is, user want a log file named `A`, but we create `B`, this is unreasonable.

erikzhang:
@shargon `StringBuilder` is more efficient.

src/SystemLog/Logger.cs
erikzhang: As I have said, the parameter `source` is not a filename. It is the log source. We just use it as filename in SystemLog.

Tommo-L: Got.

shargon:
But `Replace` returns a new `StringBuilder`, that's why I thought that in this case maybe it's not better.

src/SystemLog/Logger.cs
```diff
@@ -39,7 +40,10 @@ void ILogPlugin.Log(string source, LogLevel level, string message)
 
                 if (!string.IsNullOrEmpty(Settings.Default.Path))
                 {
-                    var path = Combine(Settings.Default.Path, source);
+                    StringBuilder sb = new StringBuilder(source);
+                    foreach (char c in GetInvalidFileNameChars())
+                        sb.Replace(c, '-');
```
shargon, line 45:
```suggestion
                        sb = sb.Replace(c, '-');
```

erikzhang: It doesn't returns new `StringBuilder`. It returns the old one. There will be no new one be created.

PR #224, opened 2020-04-16
3.0.0-preview2

erikzhang:


shargon:
This will downgrade the version of the plugins, it's only for create the release?

Tommo-L:
> This will downgrade the version of the plugins, it's only for create the release?

Yes, we'll update these plugins after release.

Tommo-L:
wait for #213

erikzhang:
Why will it downgrade the version of the plugins?

erikzhang:
I will create release on this branch, but won't merge it to master.

Issue #232, opened 2020-05-03
Save errors.log


shargon:
In order to close https://github.com/neo-project/neo/issues/1245 we need to save `Error` and `Fatal` log in a file.

shargon:
We already do this, only require update the nuget.

PR #233, opened 2020-05-04
Update nugets

shargon:
Close https://github.com/neo-project/neo-modules/issues/232 https://github.com/neo-project/neo/issues/1245

### DIFF
--- a/src/SystemLog/Logger.cs
+++ b/src/SystemLog/Logger.cs
@@ -14,8 +14,35 @@ namespace Neo.Plugins
             Settings.Load(GetConfiguration());
         }
 
+        private static void GetErrorLogs(StringBuilder sb, Exception ex)
+        {
+            sb.AppendLine(ex.GetType().ToString());
+            sb.AppendLine(ex.Message);
+            sb.AppendLine(ex.StackTrace);
+            if (ex is AggregateException ex2)
+            {
+                foreach (Exception inner in ex2.InnerExceptions)
+                {
+                    sb.AppendLine();
+                    GetErrorLogs(sb, inner);
+                }
+            }
+            else if (ex.InnerException != null)
+            {
+                sb.AppendLine();
+                GetErrorLogs(sb, ex.InnerException);
+            }
+        }
+
         void ILogPlugin.Log(string source, LogLevel level, object message)
         {
+            if (message is Exception ex)
+            {
+                var sb = new StringBuilder();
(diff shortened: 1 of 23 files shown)


superboyiii:
> Close #232 [neo-project/neo#1245](https://github.com/neo-project/neo/issues/1245)

Maybe we should upgrade nuget to 3.0.0-CI00907, since neo-cli's dependency now is changed to that. And since the incompatible change here: [3.0.0-CI00906](https://github.com/neo-project/neo/blob/946fb73d01ae942635570430b550c5a41f747654/src/neo/Plugins/ILogPlugin.cs#L5),`message` type should be changed to `object` in `void ILogPlugin.Log(...)` of SystemLog.
It works well on my local after these are applied,.
image

shargon:
@superboyiii please, could you test it again ?

src/SystemLog/Logger.cs
```diff
@@ -14,7 +14,7 @@ protected override void Configure()
             Settings.Load(GetConfiguration());
         }
 
-        void ILogPlugin.Log(string source, LogLevel level, string message)
+        void ILogPlugin.Log(string source, LogLevel level, object message)
```
erikzhang, line 17: Need to consider whether the type of `message` is `Exception` or other.

shargon: What should be the behaviour when it's an Exception?

erikzhang: https://github.com/neo-project/neo-node/commit/82033b46d36a69b31e24e6f421e5e174a2ca0d88#diff-2ad153cafa6b784ceb93f0f6a668bcbdL31-L49

superboyiii:
@shargon Find a issue on `RpcBlockHeader.FromJson`.
https://github.com/neo-project/neo-modules/issues/238

shargon:
@chenquanyu could you help me with the RPC ut?

joeqian10:
@shargon @chenquanyu I will look into this.

shargon:
@superboyiii could you check it again please?

superboyiii:
@shargon I'll check it soon.

superboyiii:
@shargon I've tested all RpcClient mehods related to modified parts, like `GetBlockHeader`, `GetBlock`, `GetContractState`, they all works well, return the same response as RPC invocation result. And SystemLogs also works well.

Issue #236, opened 2020-05-07
Plugin load sequence can cause method not loaded normally


superboyiii:
I found an issue on Neo-cli v3.0.0-Preview2 that if RpcServer plugin is loaded on the top of plugin list, some plugins which need the dependency of RpcServer will not work. For example:
![image](https://user-images.githubusercontent.com/38493437/81281437-6c98dd00-908c-11ea-8f76-f0744075c596.png)
In this way, RpcServer is loaded on the top of list which makes ApplicationLogs and RpcNepTracker not work normally, they can sync chain data but aren't able to be invoked by RPC methods, like this:
![image](https://user-images.githubusercontent.com/38493437/81281926-14aea600-908d-11ea-948b-3a67ea18895b.png)

And if Plugins are loaded in this sequence (RpcServer is loaded in the buttom), other Plugins are all worked well. 
![image](https://user-images.githubusercontent.com/38493437/81282514-d82f7a00-908d-11ea-82b4-314deed4f69e.png)

And if in this case:
![image](https://user-images.githubusercontent.com/38493437/81283117-b8e51c80-908e-11ea-9205-bf77cd3f7c92.png)
ApplicationLogs will work normally but RpcNep5Tracker will not be found in RPC method invoking.

Till now I found plugins work in the correct sequence in windows but mostly incorrect in linux. I'm not sure how to control the loading sequence. During my test it's not related to create time or plugin name! :disappointed_relieved:

shargon:
I will add a dependency

Tommo-L:
@shargon  just go ahead, plz add the dependency.

PR #239, opened 2020-05-09
Fix RpcServer RegisterMethods

ProDog:
`ApplicationLogs` and `RpcNep5Tracker` can't added in handlers if `RpcServerPlugin` loaded first.
This could lead to https://github.com/neo-project/neo-modules/issues/236 .


superboyiii:
Tested: PASS
I replace this version on my older version, it works normally now.

shargon:
Then, what is the sense of this method? https://github.com/neo-project/neo-modules/blob/57f2ab56e5bc2946e5adc7f37bbeb82bd3c5e6d1/src/ApplicationLogs/LogReader.cs#L33

ProDog:
There  registered it if the plugin in `handlers`:
https://github.com/neo-project/neo-modules/blob/57f2ab56e5bc2946e5adc7f37bbeb82bd3c5e6d1/src/RpcServer/RpcServerPlugin.cs#L41

gsmachado:
Nice! This bug was bugging me for some time. 😄

Issue #237, opened 2020-05-08
Neo-cli will stop response if systemlog file is open


superboyiii:
Neo-cli will stop response if systemlog file is open. For example, if a node start consensus and someone opens the consensus log to have a look, node will `on stop` and never response again unless restart it. It's a stability bug.
![image](https://user-images.githubusercontent.com/38493437/81375687-a2919c00-9134-11ea-80d6-7586039408b2.png)
![image](https://user-images.githubusercontent.com/38493437/81375420-0ebfd000-9134-11ea-8efc-f13411b18391.png)

PR #240, opened 2020-05-11
Allow to share logs in read mode

shargon:
Close https://github.com/neo-project/neo-modules/issues/237

### DIFF
--- a/src/SystemLog/Logger.cs
+++ b/src/SystemLog/Logger.cs
@@ -70,10 +70,16 @@ namespace Neo.Plugins
                     StringBuilder sb = new StringBuilder(source);
                     foreach (char c in GetInvalidFileNameChars())
                         sb.Replace(c, '-');
+
                     var path = Combine(Settings.Default.Path, sb.ToString());
                     Directory.CreateDirectory(path);
                     path = Combine(path, $"{now:yyyy-MM-dd}.log");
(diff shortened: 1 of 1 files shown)


superboyiii:
@shargon The same crash.
image

shargon:
I think that the unique solutions are:
- Open the log file at the beginning.
- Show a error if it's open.

superboyiii:
> I think that the unique solutions are:
> 
> * Open the log file at the beginning.
> * Show a error if it's open.

Yes, should be catch and not cause `onstop`

src/SystemLog/Logger.cs
```diff
@@ -70,10 +70,23 @@ void ILogPlugin.Log(string source, LogLevel level, object message)
...
+                    {
+                        using (var fs = new FileStream(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.Read))
+                        using (var writer = new StreamWriter(fs))
+                        {
+                            writer.WriteLine($"[{level}]{log}");
+                        }
+                    }
+                    catch (IOException)
+                    {
+                        Console.WriteLine("Error writing the log file: " + path);
```
superboyiii, line 88: @shargon Tested, expction is catched successfully. But without File.AppendAllLines(), this can only record one line.
image

superboyiii:
Now it works well.

PR #253, opened 2020-05-29
Add BloomFilter to levelDB

Qiao-Jin:
This PR adds BloomFIlter option and corresponding configuration into module levelDB.


shargon:
Looks good, did you test the speed benefits?

Qiao-Jin:
> Looks good, did you test the speed benefits?

Have tested in [#1599](https://github.com/neo-project/neo/pull/1599) as an alternative, I will provide detailed result later

erikzhang:
Leveldb will create bloom filter by default.

Tommo-L:
> Leveldb will create bloom filter by default.



We need to config `BloomFilterBitsPerKey` parameter.

src/LevelDBStore/LevelDBStore/config.json
```diff
@@ -1,5 +1,6 @@
-﻿{
+{
   "PluginConfiguration": {
-    "Path": "Data_LevelDB_{0}"
+    "Path": "Data_LevelDB_{0}",
+    "BloomFilterBitsPerKey": 20
```
erikzhang, line 4: I think this can be hardcoded.

Qiao-Jin: Maybe we should keep this modifible by user due to different hardware / node usage?

erikzhang: Hardware?

Qiao-Jin: No, justing depends on data. Maybe some TPS tests are needed to make sure of this figure if it should be hard-coded instead of tunable?

Tommo-L:
Merge?

src/LevelDBStore/LevelDBStore/config.json
erikzhang: Find a best value and fix it.

vncoelho: I agree with @erikzhang.
@Qiao-Jin, I believe it may be possible to even adjust it during running time in order to help the search for the best. You can slightly keep changing and tracking it during execution. For example:

* Increase in 1 unit
* Decrease in 1 unit
* Increase in 5 units
* Decrease in 5 units
* Increase in 10 units
* Decrease in 10 units

This is a like a Variable Neighborhood Search.
You slightly modify the bloom filter and observe the results, if not good you revert.
Basically you need an average value and some observation after change.

On the real code I believe that it is not possible because the microbenchmark is costly.

shargon: @Tommo-L If we resolve this conversation we can merge, the code looks good to me

Tommo-L: Okay, we'll add the benchmark with it.

Qiao-Jin: After testing it seems that bitsPeyKey = 15 is a good value to choose, with highest TPS.

Qiao-Jin:
> Hard-code, please.

Have hard-coded correspondingly

PR #257, opened 2020-06-04
Add plugin desc and fix ut

Tommo-L:
1. Add plugin desciption
2. Sync neo-core, fix the calculation of networkFee, as interopservice updated https://github.com/neo-project/neo/pull/1655


ProDog:
> Very good, description for RPCClient is missing?

It doesn't need show in the neo-cli.

vncoelho:
That is true...aheauhuea

PR #262, opened 2020-06-12
Change json field names for GetVersion

### LINKED ISSUE #261: Difference between implementation and documentation of GetVersion RPC method
**Describe the bug**
Implementation of [`GetVersion`](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.Node.cs#L60) RPC method assumes lover-cased underscored JSON fields `tcp_port`, `ws_port` and `user_agent`. Documentation [example](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/getversion.html) differs from that and describes `tcpPort`, `wsPort` and `useragent`.

**Expected behavior**
Documentation should match implementation.

joeqian10:
Closes #261

### DIFF
--- a/src/RpcServer/RpcServer.Node.cs
+++ b/src/RpcServer/RpcServer.Node.cs
@@ -60,10 +60,10 @@ namespace Neo.Plugins
         private JObject GetVersion(JArray _params)
         {
             JObject json = new JObject();
-            json["tcp_port"] = LocalNode.Singleton.ListenerTcpPort;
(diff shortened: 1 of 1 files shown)


joeqian10:
@shargon @Tommo-L Maybe we should revert this since other json responses are using lover-cased underscored fields. I think it's time to make this consistent. The documentation is also needed to change.

PR #276, opened 2020-06-29
Change json fields to all lower case for consistency

### LINKED ISSUE #261: Difference between implementation and documentation of GetVersion RPC method

joeqian10:
Close #261


erikzhang:
I think we don't need `_`.

src/RpcClient/Models/RpcRequest.cs
```diff
@@ -18,7 +18,7 @@ public static RpcRequest FromJson(JObject json)
             return new RpcRequest
             {
                 Id = json["id"],
-                JsonRpc = json["jsonrpc"].AsString(),
+                JsonRpc = json["json_rpc"].AsString(),
```
erikzhang, line 21: This breaks the jsonrpc specification.

joeqian10:
> I think we don't need `_`.

I will change everything to all lower-cased. Actually I also think the snake case need to waste more resourses when passing requests and responses back and forth, though tiny.

Issue #278, opened 2020-07-01
Issue for invokefunction add signature


ProDog:
**Describe the bug**

Now `invokefunction` and `invokescript` supports check witness, and add signature when wallet has opened in rpc.
But now add signature is abnormal:
I have opened wallet in rpc and call `invokefunction` to get tx, the stack in result is `true`, next I call `sendrawtransaction` to send the tx to node, then I check the transaction's stack is `false`.

**To Reproduce**
1. Open the wallet in rpc
2. Call `invokefunction` and `invokescript` with witness and get tx
3. Call `sendrawtransaction` to send tx

**Additional context**
Result for invokefunction:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "script": "180c14e3137eddba5f6485cad334a79bdb67c43273171f0c141c0357464b777ecf6b5f3ac3893ace1f8b1621f613c00c087472616e736665720c14bcaf41d684c7d4ad6ee0d99da9707b9d1f0c8e6641627d5b52",
        "state": "HALT",
        "gas_consumed": "9007960",
        "stack": [
            {
                "type": "Boolean",
                "value": true
            }
        ],
        "tx": "001477b7261c0357464b777ecf6b5f3ac3893ace1f8b1621f658738900000000002ef11200000000003d2b20000054180c14e3137eddba5f6485cad334a79bdb67c43273171f0c141c0357464b777ecf6b5f3ac3893ace1f8b1621f613c00c087472616e736665720c14bcaf41d684c7d4ad6ee0d99da9707b9d1f0c8e6641627d5b5201420c407b9d29b7316249f6b6ad6d4b6dfb401407cd4eee0b0ea54afff0c01340398e67b2b20f80f70b0b36104f6d9ea1ab889b3a8109fad1d106d10fb29efb676b4ba5290c210222d8515184c7d62ffa99b829aeb4938c4704ecb0dd7e340e842e9df1218263430b4195440d78"
    }
}
```

Result for sendrawtransaction:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "hash": "0xfb5e74fb8748409bf5678e3d8d4819954d6273acbe469e17e41fdd1f651de7b4"
    }
}
```

Get the transaction's notify:
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "result": {
[...]

Issue #260, opened 2020-06-10
invokefunction and invokescript RPC calls should support scoped signatures


roman-khimov:
**Summary or problem description**
`invokefunction` and `invokescript` RPC calls support passing script hashes for verifying that are to be used by `System.Runtime.CheckWitness` syscall (thanks to neo-project/neo#335 and #214 changes). It's a very useful feature, but these hashes are not processed the way they will be processed when they're added as Cosigners for a transaction (and that's what is supposed to happen with a lot of these test invocations, trying first and then sending a real transaction) because transactions use scoped signatures since neo-project/neo#968 merge. It may affect the result of test invocation and lead to differences with execution on real network.

**Do you have any solution you want to propose?**
I propose replacing current script hashes parameter (which is just the last parameter for current `invokefunction` and `invokescript` that is supposed to be an array of hex-encoded 20-byte script hashes) with cosigners parameter that is to be represented as an array of JSON-serialized `Cosigner` structures (that are well-defined).

Example request using this format (similar to https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokefunction.html):
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "invokefunction",
  "params": [
    "0x806b7fa0db3b46d6c42e1e1b0a7fd50db9d4a9b0",
    "transfer",
    [
      {
        "type": "Hash160",
        "value": "0xcadb3dc2faa3ef14a13b619c9a43124755aa2569"
      },
      {
        "type": "Hash160",
        "value": "0x2916eba24e652fa006f3e5eb8f9892d2c3b00399"
      },
      {
        "type": "Integer",
        "value": "1000"
      }
    ],
    [
      {
        "account": "0xcadb3dc2faa3ef14a13b619c9a43124755aa2569",
        "scopes": "CalledByEntry"
      }
    ]
  ]
}
```
**Where in the software does this update applies to?**
- RPC (HTTP)

roman-khimov:
Side note: we have it now implemented as an extension in neo-go, so it accepts both an array of hashes (that are treated like cosigners with Global scope) and an array of proper Cosigners.

superboyiii:
@roman-khimov #280 was merged, your issue should be solved.

PR #280, opened 2020-07-02
fix add signature in invokefunction and invokescript

ProDog:
Close https://github.com/neo-project/neo-modules/issues/278 .
I have added signature like https://github.com/neo-project/neo-modules/issues/260#issue-636540014.

Call invokefunction:
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "invokefunction",
  "params": [
    "0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc",
    "transfer",
    [
      {
        "type":"Hash160",
        "value":"0xf621168b1fce3a89c33a5f6bcf7e774b4657031c"
        },
      {
        "type":"Hash160",
        "value":"0x1f177332c467db9ba734d3ca85645fbadd7e13e3"
        },
      {
        "type":"Integer",
        "value":"8"
        }        
    ],
    [
        {
          "account": "0xf621168b1fce3a89c33a5f6bcf7e774b4657031c",
          "scopes": "CalledByEntry"      
        }
    ]
  ]
}
```

If we have opened wallet in rpc, the resulting tx can be sent directly by `sendrawtransaction`.


src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -138,11 +138,14 @@ private JObject OpenWallet(JArray _params)
...
 
-        private void ProcessInvokeWithWallet(JObject result)
+        private void ProcessInvokeWithWallet(JObject result, CheckWitnessHashes checkWitnessHashes = null)
         {
-            if (wallet != null)
+            if (wallet != null && checkWitnessHashes != null)
             {
-                Transaction tx = wallet.MakeTransaction(result["script"].AsString().HexToBytes());
+                UInt160[] accounts = wallet.GetAccounts().Where(p => !p.Lock && !p.WatchOnly).Select(p => p.ScriptHash).ToArray();
+                Cosigner[] cosigners = checkWitnessHashes.GetCosigners().Where(p => accounts.Contains(p.Account)).ToArray();
```
shargon, line 146: Before the change, the cosigners could be null and it work

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -138,11 +138,14 @@ private JObject OpenWallet(JArray _params)
             return true;
         }
 
-        private void ProcessInvokeWithWallet(JObject result)
+        private void ProcessInvokeWithWallet(JObject result, Cosigners cosigners = null)
         {
-            if (wallet != null)
+            if (wallet != null && cosigners != null)
```
shargon, line 143: I think that we should throw an exception if doesn't have the wallet or cosigners, otherwise it doesn't do anything

ProDog: Added :-)

shargon: If wallet or cosigners are null, nothig happend

ProDog: If wallet or cosigners are null, it just return the stack, don't include the tx, need anything else?

shargon: But the rpc client will notice that this happend?

ProDog: Set the default tx to null ?

PR #281, opened 2020-07-07
Fix seek

Tommo-L:
- Fix Rocksdb.Seek
- Fix Leveldb.Seek, avoid cross table.

@ProDog plz, try it again.


src/LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -39,5 +41,14 @@ internal static byte[] ToByteArray(this IntPtr data, UIntPtr length)
             Marshal.Copy(data, buffer, 0, (int)length);
             return buffer;
         }
+
+        public static byte[] CreateKey(byte table, byte[] key = null)
```
shargon, line 45: We should try to detach RocksDb form LevelDb

Tommo-L: ? This method is only for LevelDb, and there is no dependence between them.

ProDog:
I have tested, it's OK.

src/LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -39,5 +41,14 @@ internal static byte[] ToByteArray(this IntPtr data, UIntPtr length)
...
+        {
+            if (key is null) return new[] { table };
```
shargon, line 47: or length ==0?

Tommo-L: Done

PR #286, opened 2020-07-14
update signers for invokefunction and invokescript

ProDog:
Apply https://github.com/neo-project/neo/pull/1752.

Update WitnessScope for signers.

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -61,7 +61,7 @@ namespace Neo.Plugins
 
         private JObject GetInvokeResult(byte[] script, Cosigners cosigners = null)
         {
-            using ApplicationEngine engine = ApplicationEngine.Run(script, cosigners, extraGAS: settings.MaxGasInvoke);
+            using ApplicationEngine engine = ApplicationEngine.Run(script, cosigners, gas: settings.MaxGasInvoke);
             JObject json = new JObject();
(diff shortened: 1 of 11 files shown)


src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -74,31 +76,42 @@ private JObject GetInvokeResult(byte[] script, Cosigners cosigners = null)
...
                 json["stack"] = "error: recursive reference";
             }
-            ProcessInvokeWithWallet(json, cosigners);
+            ProcessInvokeWithWallet(json, signers);
             return json;
         }
 
+        private static Signers SignersFromJson(JArray _params)
+        {
+            return new Signers(_params.Select(u => new Signer()
```
shargon, line 85: We can serialize it and deserialize it in order to check that all the information it's good

https://github.com/neo-project/neo/blob/54784e03091aa217e133fbe3e70288d36c985e58/src/neo/Network/P2P/Payloads/Signer.cs#L28-L39

Tommo-L: But it's not convenient for the caller.

shargon: If he fill a wrong signer, he's tx will never arrive

ProDog: @shargon can you help modify this😅

Issue #290, opened 2020-07-15
RocksDB Store should use SeekForPrev instead of Seek for SeekDirection.Backwards


devhawk:
RocksDB `Store.Seek` method currently calls `db.Seek` to position the iterator on the first key that matches the prefix parameter, regardless of `SeekDirection`. If `SeekDirection` is `Backwards`, the method will call `iterator.Prev()`, which will - by definition - never return a value since the original `Seek` call placed the iterator at the first key matching the prefix.

devhawk:
FYI, levelDB store probably has same problem

PR #291, opened 2020-07-16
Fix db.Seek

Tommo-L:
Close https://github.com/neo-project/neo-modules/issues/290

### DIFF
--- a/src/LevelDBStore/IO/Data/LevelDB/Helper.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/Helper.cs
@@ -7,19 +7,46 @@ namespace Neo.IO.Data.LevelDB
 {
     public static class Helper
     {
+        private static int ByteArrayCompare(byte[] x, byte[] y)
+        {
+            int length = Math.Min(x.Length, y.Length);
+            for (int i = 0; i < length; i++)
+            {
+                int r = x[i].CompareTo(y[i]);
+                if (r != 0) return r;
+            }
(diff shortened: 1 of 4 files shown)


src/LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -7,19 +7,46 @@ namespace Neo.IO.Data.LevelDB
 {
     public static class Helper
     {
+        private static int ByteArrayCompare(byte[] x, byte[] y)
```
devhawk, line 10: Since LevelDBStore project is netstandard2.1, you can use Span<T>.SequenceCompare instead of having to write your own ByteArrayCompare function

src/LevelDBStore/IO/Data/LevelDB/Helper.cs
```diff
@@ -7,19 +7,46 @@ namespace Neo.IO.Data.LevelDB
...
-                yield return resultSelector(it.Key(), it.Value());
+                // SeekForPrev
 
-                if (direction == SeekDirection.Forward)
-                    it.Next();
-                else
+                it.Seek(target);
+                if (!it.Valid())
+                    it.SeekToLast();
+                else if (ByteArrayCompare(it.Key(), target) > 0)
```
devhawk, line 41: use `it.Key().AsSpan().SequenceCompareTo(target)` here

PR #292, opened 2020-07-16
fix main.yml

cloud8little:


shargon:
what status left?

superboyiii:
@shargon Do we need to do something to shut down Travis to pass the check? I just removed the setting file but it seems still `Waiting for status to be reported`.

shargon:
@superboyiii 
image

superboyiii:
@shargon Seems the issue is here. Could you do this?
https://github.community/t/expected-waiting-for-status-to-be-reported/18001

shargon:
Done, this was the problem. Now require 

image

PR #294, opened 2020-07-17
Solve workflow conflict

superboyiii:
Solve workflow conflict from master-2.x and master


shargon:
This is for master only, right?

superboyiii:
> This is for master only, right?

Yes

PR #299, opened 2020-07-22
Merge master2.x to neox-2.x

superboyiii:
Merge master2.x to neox-2.x


superboyiii:
@shargon I'll fix the workflow after this merge. Change on push to neox-2.x and choose actions/checkout@v2.

superboyiii:
@shargon Could you have a look? We need these improvement on neox-2.x

PR #301, opened 2020-07-23
update UT

ProDog:
Update UT for invokescript


Tommo-L:
It's included in https://github.com/neo-project/neo-modules/pull/302 now.

PR #302, opened 2020-07-27
Add Utility.StackItemFromJson

Tommo-L:
Close https://github.com/neo-project/neo-modules/issues/293 Close https://github.com/neo-project/neo-modules/issues/300

PR #303, opened 2020-07-30
Add Utility.StackItemFromJson (#302)（preview3）

superboyiii:
* fix invoke tojson

* add StackItemFromJson

* optimize code

* remove src/RpcClientTests/UtilityTests.cs

Co-authored-by: Luchuan <luchuan@neo.org>


shargon:
@superboyiii You can create a tag instead of a branch for preview3

superboyiii:
@shargon We need preview3 branch for test and fix, but tag is a certain checkpoint and not allowed to change. If we release, we'll surely add release tag, but before that, we need a release branch to test and ensure everything is OK until release.

PR #307, opened 2020-08-04
Merge Preview3 to Master

superboyiii:
Merge Preview3 to Master

### DIFF
--- a/src/RpcClient/RpcClient.csproj
+++ b/src/RpcClient/RpcClient.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Version>3.0.0-preview2</Version>
+    <Version>3.0.0-preview3</Version>
     <TargetFramework>netstandard2.1</TargetFramework>
     <RootNamespace>Neo.Network.RPC</RootNamespace>
     <Authors>The Neo Project</Authors>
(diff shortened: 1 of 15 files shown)


src/RpcClient/RpcClient.csproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="3.0.0-CI00972" />
+    <PackageReference Include="Neo" Version="3.0.0-preview3-00" />
```
shargon, line 17: I think we should use the latest version for nugets in master

shargon: or do you want to merge for create the tag?

superboyiii: > or do you want to merge for create the tag?

Yes, should use the latest version of nuget. I've already created the tag on preview3 branch, no need on master since there will be extra commit on master.

superboyiii: Shall we merge first and I upgrade the latest neo dependency in master later? I think they are two workflows after release.

shargon: I think that we can update them here

superboyiii: > I think that we can update them here

I can't do this since `preview3` is the release branch and should not be changed from nuget `3.0.0-preview3-00`. I have to wait until master is merged.

shargon: ok! approved :)

PR #309, opened 2020-08-06
update signers and add sender for rpc api

ProDog:
- add signers for send asset api
- supports address and script hash
- add sender for invoke contract api

### DIFF
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -313,7 +313,7 @@ namespace Neo.Plugins
         }
 
         internal static UInt160 AddressToScriptHash(string address)
-        {            
+        {
             if (UInt160.TryParse(address, out var scriptHash))
             {
                 return scriptHash;
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -42,7 +42,7 @@ namespace Neo.Plugins
         private JObject DumpPrivKey(JArray _params)
         {
             CheckWallet();
-            UInt160 scriptHash = StringToScriptHash(_params[0].AsString());
+            UInt160 scriptHash = AddressToScriptHash(_params[0].AsString());
             WalletAccount account = wallet.GetAccount(scriptHash);
             return account.GetKey().Export();
         }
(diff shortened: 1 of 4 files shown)


src/RpcServer/RpcServer.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Version>3.0.0-preview3</Version>
+    <Version>3.0.0-preview2</Version>
```
shargon, line 4:
```suggestion
    <Version>3.0.0-preview3</Version>
```

ProDog:
Contract calls support custom scope, like `invokefunction` and `invokescript`.
For the wallet transfer function, I think `CalledByEntry` is enough, it is easier for users, and they can call it by `invoke` api if they want to customize it.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -299,5 +311,21 @@ private JObject SignAndRelay(Transaction tx)
                 return context.ToJson();
             }
         }
+
+        internal static UInt160 TryScriptHash(string input)
```
Tommo-L, line 315: Need a better name, `TryXXXX` should return bool value.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -299,5 +311,21 @@ private JObject SignAndRelay(Transaction tx)
                 return context.ToJson();
             }
         }
+
+        internal static UInt160 StringToScriptHash(string input)
```
vncoelho, line 315: Maybe it is better to create a function just for AssetID and other function for scripthash.

ProDog: To AssetID can use UInt160.Parse() directly, now this just for scripthash.

ProDog:
Can merge?

shargon (approved):

shargon:
The rpc specifications will need to be updated.

Tommo-L (approved):

Tommo-L:
> The rpc specifications will need to be updated.

I have opened the issue https://github.com/neo-project/docs/issues/1129, we'll update it later.

PR #317, opened 2020-08-12
fix listplugins

### LINKED ISSUE #316: listplugins is inconsistent with plugins command
**Describe the bug**
listplugins when RpcServer is installed. 

**To Reproduce**
1. Install RpcServer, ApplicationLogs, LevelDBStore and RpcNep5Tracker  
2. Start the node
3. call listplugins request, and input plugins in the command.

in neo-cli, 
```
Loaded plugins:
        ApplicationLogs Synchronizes the smart contract log with the NativeContract log (Notify)
        LevelDBStore    Uses LevelDB to store the blockchain data
        RpcNep5Tracker  Enquiries NEP-5 balances and transaction history of accounts through RPC
        RpcServer       Enables RPC for the node
```
listplugins response:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "name": "ApplicationLogs",
      "version": "3.0.0.0",
      "interfaces": [
        "IPersistencePlugin"
      ]
    },
    {
      "name": "LevelDBStore",
      "version": "3.0.0.0",
      "interfaces": []
    },
    {
      "name": "RpcNep5Tracker",
      "version": "3.0.0.0",
      "interfaces": [
        "IPersistencePlugin"
      ]
    },
    {
      "name": "RpcServer",
      "version": "3.0.0.0",
      "interfaces": []
    },
    {
      "name": "SystemLog",
      "version": "3.0.0.0",
      "interfaces": [
        "ILogPlugin"
      ]
    }
  ]
}
```
**Expected behavior**:
1. SystemLog should not show.
2. interfaces for RpcServer should not be empty.

Screenshot:
f17fbfbeab95bf89fb500a22a8d38e1

02aaf7a349c124f39e2bc048e1542c7

**Platform:**
 - OS: [e.g. Windows 10 x64]
 - Version preview3

cloud8little:
fix #316

### DIFF
--- a/src/RpcServer/RpcServer.Utilities.cs
+++ b/src/RpcServer/RpcServer.Utilities.cs
@@ -13,14 +13,14 @@ namespace Neo.Plugins
         private JObject ListPlugins(JArray _params)
         {
             return new JArray(Neo.Plugins.Plugin.Plugins
-                .OrderBy(u => u.Name)
+                .OrderBy(u => u.Name).Where(u => !(u is ILogPlugin))
                 .Select(u => new JObject
                 {
                     ["name"] = u.Name,
                     ["version"] = u.Version.ToString(),
                     ["interfaces"] = new JArray(u.GetType().GetInterfaces()
                         .Select(p => p.Name)
-                        .Where(p => p.EndsWith("Plugin"))
+                        .Where(p => p.EndsWith("Plugin") || p.EndsWith("Provider"))
                         .Select(p => (JObject)p))
                 }));
         }


src/RpcServer/RpcServer.Utilities.cs
```diff
@@ -13,14 +13,14 @@ partial class RpcServer
...
-                .OrderBy(u => u.Name)
+                .OrderBy(u => u.Name).Where(u => !(u is ILogPlugin))
                 .Select(u => new JObject
                 {
                     ["name"] = u.Name,
                     ["version"] = u.Version.ToString(),
                     ["interfaces"] = new JArray(u.GetType().GetInterfaces()
                         .Select(p => p.Name)
-                        .Where(p => p.EndsWith("Plugin"))
+                        .Where(p => p.EndsWith("Plugin") || p.EndsWith("Provider"))
```
shargon, line 23: This seems that we will have a lot of || in a future, we should change this logic

erikzhang: Why do we need to filter it? I think we can just list all the interfaces.

cloud8little: > Why do we need to filter it? I think we can just list all the interfaces.

do you want IDisposable to show in all plugins?

src/RpcServer/RpcServer.Utilities.cs
```diff
@@ -13,14 +13,14 @@ partial class RpcServer
         private JObject ListPlugins(JArray _params)
         {
             return new JArray(Neo.Plugins.Plugin.Plugins
-                .OrderBy(u => u.Name)
+                .OrderBy(u => u.Name).Where(u => !(u is ILogPlugin))
```
erikzhang, line 16: All `ILogPlugin` will not be displayed? I think it's incorrect.

cloud8little: > All `ILogPlugin` will not be displayed? I think it's incorrect.

You are right, my intention is to exclude SystemLog. 
refs: https://github.com/neo-project/neo-node/blob/59da50115537c1941eee9365ae79febe7883146c/neo-cli/CLI/MainService.Plugins.cs#L109-L110

shargon: Maybe you can exclude them if his assembly is `neo-node`

erikzhang:
Is this PR still updated?

PR #319, opened 2020-08-13
fix #312

### LINKED ISSUE #312: RpcServer.GetInvokeResult doesn't compile against neo master
ApplicationEngine.Run with signers argument appears to have been removed in neo master

cloud8little:


erikzhang:
What's the difference?

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -63,7 +63,7 @@ public void SerializeUnsigned(BinaryWriter writer)
 
         private JObject GetInvokeResult(byte[] script, Signers signers = null)
         {
-            using ApplicationEngine engine = ApplicationEngine.Run(script, signers, gas: settings.MaxGasInvoke);
+            using ApplicationEngine engine = ApplicationEngine.Run(script, container: signers, gas: settings.MaxGasInvoke);
```
Tommo-L, line 66:
```cs
public static ApplicationEngine Run(byte[] script, StoreView snapshot = null, IVerifiable container = null, Block persistingBlock = null, int offset = 0, long gas = TestModeGas)
```
https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/ApplicationEngine.cs#L290

`signers` will be assigned to `snapshot`.

PR #322, opened 2020-08-14
Allow to try to repair leveldb before opening

erikzhang:


src/LevelDBStore/Plugins/Storage/LevelDBStore.cs
```diff
@@ -15,6 +18,8 @@ protected override void Configure()
 
         public IStore GetStore()
         {
+            if (Environment.CommandLine.Split(' ').Any(p => p == "/repair" || p == "--repair"))
```
shargon, line 21: We should add a flag in order avoid a second repair if we call twice to `GetStore`

erikzhang: If you call `GetStore()` with the same path twice, the second call will be blocked. So you must close the db first. But once you closed it, it may be damaged again. So there is no need to prevent multiple repairs. And even if there are multiple repairs, there is no harm.

Issue #298, opened 2020-07-17
Move Blockchain Toolkit storage providers to Neo Modules project


devhawk:
The Neo3 version of [Neo Express](https://github.com/neo-project/neo-express) and [Neo Debugger](https://github.com/neo-project/neo-debugger) both depend on a RocksDB and Checkpoint implementation of Neo.Persistence.IStore. Currently, these live in the [ngdseattle/persistence](https://github.com/ngdseattle/neo-persistence) project. However, it feels like neo-modules is the correct repo for these storage providers.

The Blockchain Toolkit RocksDB provider is similar to the existing neo-modules RocksDB provider, with the following improvements

* **Create column families on demand for faster startup**. In preview 2 testing, the current neo-modules RocksDB provider took 10-15 seconds to start as it creates 256 column families on first run. For a production node, that will likely go thru this initialization once in its lifetime, this delay is perfectly fine. For a developer node, this delay is unacceptable. 
* **Checkpoint Management**  Blockchain Toolkit uses RocksDB for persistence because it can checkpoint a database [while it's running](https://github.com/facebook/rocksdb/wiki/Checkpoints). The Blockchain Toolkit RocksDB provider includes methods to create, restore and validate checkpoints.
* **Read-Only Mode** The Blockchain Toolkit RocksDB supports OpenReadOnly for use with the checkpoint provider (described below)

The Checkpoint Provider supports neo-express's [checkpoint commands](https://github.com/neo-project/neo-express/blob/master/docs/command-reference.md#checkpoint-command). This provider uses a Neo.Persistence.IReadOnlyStore instance and stores all changes in memory. This is primarily useful for test and debug scenarios.  

Note, I don't feel strongly about where these developer focused storage providers should live. If the neo core dev community felt that neo-modules repo is for production modules and that these developer-scenario focused storage providers belong somewhere else. we can easily create a new repo with a name like neo-dev-modules for these (and future dev-scenario modules) to live.

erikzhang:
@shargon

shargon:
> Create column families on demand for faster startup. In preview 2 testing, the current neo-modules RocksDB provider took 10-15 seconds to start as it creates 256 column families on first run. For a production node, that will likely go thru this initialization once in its lifetime, this delay is perfectly fine. For a developer node, this delay is unacceptable.

Agree, we should move prefix to an enum in order to know the current columns.

[...]

PR #333, opened 2020-08-25
Optimize RocksDb startup

shargon:
Related to https://github.com/neo-project/neo-modules/issues/298

### DIFF
--- a/src/RocksDBStore/Plugins/Storage/Store.cs
+++ b/src/RocksDBStore/Plugins/Storage/Store.cs
@@ -20,8 +20,16 @@ namespace Neo.Plugins.Storage
         public Store(string path)
         {
             var families = new ColumnFamilies();
-            for (int x = 0; x <= byte.MaxValue; x++)
-                families.Add(new ColumnFamilies.Descriptor(x.ToString(), new ColumnFamilyOptions()));
+
+            try
+            {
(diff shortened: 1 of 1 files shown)


src/RocksDBStore/Plugins/Storage/Store.cs
```diff
@@ -20,8 +20,16 @@ internal class Store : IStore
         public Store(string path)
         {
             var families = new ColumnFamilies();
-            for (int x = 0; x <= byte.MaxValue; x++)
-                families.Add(new ColumnFamilies.Descriptor(x.ToString(), new ColumnFamilyOptions()));
+
+            try
+            {
+                foreach (var family in RocksDb.ListColumnFamilies(Options.Default, Path.GetFullPath(path)))
```
devhawk, line 26: How to column families get added after this change? For a newly created RocksDb store, there will be no column families and it doesn't look like you've added a call to CreateColumnFamily anywhere (though I could have missed that)

shargon: They are created when not found here.
https://github.com/neo-project/neo-modules/blob/496d390f803bfafffc0038867c281ddf80f75664/src/RocksDBStore/Plugins/Storage/Store.cs#L76-L82

Issue #311, opened 2020-08-10
RpcClient.CalculateNetworkFee no longer compiles against neo master branch


devhawk:
Wallet.CalculateNetworkFee signature has changed.

ProDog:
That changed in https://github.com/neo-project/neo/pull/1800/files#diff-bac7c9fe6b52541341827f089612cfeaL356-L399;
RpcClient need calculate network fee in https://github.com/neo-project/neo-modules/blob/master/src/RpcClient/TransactionManager.cs#L105
@shargon do you have any ideas?😀

shargon:
```csharp
public TransactionManager Sign()
        {
            // Calculate NetworkFee
            Tx.NetworkFee = Wallet.CalculateNetworkFee(Blockchain.Singleton.GetSnapshot(), Tx);
            var gasBalance = nep5API.BalanceOf(NativeContract.GAS.Hash, Tx.Sender);
        
...
```
?

ProDog:
It's good, but now `Wallet.CalculateNetworkFee` is non-static.

shargon:
Please, take a look to https://github.com/neo-project/neo/pull/1839

PR #338, opened 2020-09-02
Update dependencies

shargon:
Require https://github.com/neo-project/neo/pull/1898
Close https://github.com/neo-project/neo-modules/issues/311 https://github.com/neo-project/neo-modules/issues/346

### DIFF
--- a/src/LevelDBStore/IO/Data/LevelDB/DB.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/DB.cs
@@ -60,7 +60,7 @@ namespace Neo.IO.Data.LevelDB
                     Native.leveldb_free(error);
                     return false;
                 }
-                return true;
+                return value != IntPtr.Zero;
             }
             finally
             {
--- a/src/LevelDBStore/IO/Data/LevelDB/DB.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/DB.cs
@@ -50,6 +50,24 @@ namespace Neo.IO.Data.LevelDB
             }
         }
 
+        public bool Contains(ReadOptions options, byte[] key)
+        {
+            IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
+            try
+            {
+                if (error != IntPtr.Zero)
+                {
+                    Native.leveldb_free(error);
+                    return false;
+                }
+                return true;
+            }
+            finally
+            {
+                if (value != IntPtr.Zero) Native.leveldb_free(value);
+            }
+        }
+
         public Snapshot GetSnapshot()
         {
             return new Snapshot(handle);
--- a/src/LevelDBStore/IO/Data/LevelDB/DB.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/DB.cs
@@ -35,37 +35,40 @@ namespace Neo.IO.Data.LevelDB
         public byte[] Get(ReadOptions options, byte[] key)
         {
             IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
-            try
+
+            if (error != IntPtr.Zero)
             {
-                if (error != IntPtr.Zero)
-                {
-                    Native.leveldb_free(error);
-                    return null;
-                }
-                return value.ToByteArray(length);
+                Native.leveldb_free(error);
+                return null;
             }
-            finally
+
+            if (value != IntPtr.Zero)
             {
-                if (value != IntPtr.Zero) Native.leveldb_free(value);
+                var ret = value.ToByteArray(length);
+                Native.leveldb_free(value);
+                return ret;
             }
+
+            return null;
         }
 
         public bool Contains(ReadOptions options, byte[] key)
         {
             IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
-            try
+
+            if (error != IntPtr.Zero)
             {
-                if (error != IntPtr.Zero)
-                {
-                    Native.leveldb_free(error);
-                    return false;
-                }
-                return value != IntPtr.Zero;
+                Native.leveldb_free(error);
+                return false;
             }
-            finally
+
+            if (value != IntPtr.Zero)
             {
-                if (value != IntPtr.Zero) Native.leveldb_free(value);
+                Native.leveldb_free(value);
+                return true;
             }
+
+            return false;
         }
 
         public Snapshot GetSnapshot()
--- a/src/LevelDBStore/IO/Data/LevelDB/DB.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/DB.cs
@@ -35,12 +35,7 @@ namespace Neo.IO.Data.LevelDB
         public byte[] Get(ReadOptions options, byte[] key)
         {
             IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
-
-            if (error != IntPtr.Zero)
-            {
-                Native.leveldb_free(error);
-                return null;
-            }
+            NativeHelper.CheckError(error);
 
             if (value != IntPtr.Zero)
             {
@@ -55,12 +50,7 @@ namespace Neo.IO.Data.LevelDB
         public bool Contains(ReadOptions options, byte[] key)
         {
             IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
-
-            if (error != IntPtr.Zero)
-            {
-                Native.leveldb_free(error);
-                return false;
-            }
+            NativeHelper.CheckError(error);
 
             if (value != IntPtr.Zero)
             {
--- a/src/LevelDBStore/IO/Data/LevelDB/DB.cs
+++ b/src/LevelDBStore/IO/Data/LevelDB/DB.cs
@@ -49,7 +49,7 @@ namespace Neo.IO.Data.LevelDB
 
         public bool Contains(ReadOptions options, byte[] key)
         {
-            IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
+            IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out _, out IntPtr error);
             NativeHelper.CheckError(error);
 
             if (value != IntPtr.Zero)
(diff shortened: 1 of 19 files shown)


shargon:
@superboyiii could your team test if it close https://github.com/neo-project/neo-modules/issues/311 succesfully?

cloud8little:
> @ProDog @superboyiii could your team test if it close #311 succesfully?

Sure.

erikzhang:
@shargon Can you try to use `leveldb_iter_seek()` + `leveldb_iter_key()` and see if it improves?

shargon:
```Without seek:

Exist: 00:00:00.0588924
Don't exist: 00:00:00.0366033

With seek:

Exist: 00:00:00.2610647
Don't exist: 00:00:00.2114671
```

Seek Contains:

```csharp
public bool Contains(ReadOptions options, byte[] key)
        {
            using var iterator = NewIterator(options);

            iterator.Seek(key);
            var read = iterator.Key();
            if (read.Length != key.Length) return false;

            for (int x = read.Length - 1; x >= 0; x--)
            {
                if (read[x] != key[x]) return false;
            }

            return true;
        }
```

Test:

```csharp
[TestMethod]
        public void TestSpeed()
        {
            var key = new byte[20];
            var data = new byte[1025];

            var rand = new Random();
            rand.NextBytes(key);
            rand.NextBytes(data);

            using (var plugin = new Neo.Plugins.Storage.LevelDBStore())
            {
                // Test all with the same store

                var store = plugin.GetStore();
                store.Delete(1, key);
                store.Put(1, key, data);

                var st = new Stopwatch(); st.Start();

                for (int x = 0; x < 100_000; x++)
                {
                    Assert.IsTrue(store.Contains(1, key));
                }

                st.Stop();
                Console.WriteLine("Exist: " + st.Elapsed);

                key[0]++;
                st = new Stopwatch(); st.Start();

                for (int x = 0; x < 100_000; x++)
                {
                    Assert.IsFalse(store.Contains(1, key));
                }

                st.Stop();
                Console.WriteLine("Don't exist: " + st.Elapsed);
            }
        }
```

src/LevelDBStore/IO/Data/LevelDB/DB.cs
```diff
@@ -35,37 +35,40 @@ public void Delete(WriteOptions options, byte[] key)
         public byte[] Get(ReadOptions options, byte[] key)
         {
             IntPtr value = Native.leveldb_get(handle, options.handle, key, (UIntPtr)key.Length, out UIntPtr length, out IntPtr error);
-            try
+
```
erikzhang: Why remove try block?

shargon: If something was wrong here we should stop the node, otherwise we return an empty array.

erikzhang: But it used the `finally` block only. There was no `catch` block.

src/RpcClient/TransactionManager.cs
```diff
@@ -197,7 +183,8 @@ public TransactionManager AddWitness(UInt160 scriptHash, params object[] paramet
         public TransactionManager Sign()
         {
             // Calculate NetworkFee
-            Tx.NetworkFee = CalculateNetworkFee();
+            Tx.NetworkFee = new DummyWallet().CalculateNetworkFee(Blockchain.Singleton.GetSnapshot(), Tx);
```
cloud8little, line 186: it will cause timeout because Blockchain can not be initialized in rpcclient if without full-ledger data.

src/LevelDBStore/IO/Data/LevelDB/DB.cs
shargon: I will revert it.

src/RpcClient/TransactionManager.cs
shargon: Without the ledger data we can't compute the fee....

erikzhang: We need snapshot to run the contracts. Otherwise it is impossible to compute the fees.

cloud8little: > it will cause timeout because Blockchain can not be initialized in rpcclient if without full-ledger data.

@shargon  for rpcclient, the Blockchain.Singleton is null. 

```cs
        public static LocalNode Singleton
        {
            get
            {
                while (singleton == null) Thread.Sleep(10);
                return singleton;
            }
        }
```

src/RpcClient/TransactionManager.cs
```diff
@@ -194,13 +180,18 @@ public TransactionManager AddWitness(UInt160 scriptHash, params object[] paramet
...
-        public TransactionManager Sign()
+        /// <param name="snapshot">Snapshot required for Compute NetworkFee</param>
+        public TransactionManager Sign(SnapshotView snapshot)
         {
-            // Calculate NetworkFee
-            Tx.NetworkFee = CalculateNetworkFee();
-            var gasBalance = nep5API.BalanceOf(NativeContract.GAS.Hash, Tx.Sender);
-            if (gasBalance < Tx.SystemFee + Tx.NetworkFee)
-                throw new InvalidOperationException($"Insufficient GAS in address: {Tx.Sender.ToAddress()}");
+            if (snapshot != null)
```
shargon, line 186: If we use `snapshot == null` we can set the networkFee manually with the `Tx` property

src/RpcClient/TransactionManager.cs
cloud8little: Can we calculate the network fee in RpcServer?

shargon: We can send the tx in raw and compute the fee

shargon: @cloud8little please check it again

cloud8little:
Test Fail:

sendrawtransaction return invalid.  ("00da5bbe725873890000000000580f0200000000000717200001b13beae9e2edc2d1055d8116b37ab771c5c5467601005500640c149c09b202591267ecf38d953ae3b5b8d260ee693e0c14b13beae9e2edc2d1055d8116b37ab771c5c5467613c00c087472616e736665720c1425059ecb4878d3a875f91c51ceded330d4575fde41627d5b5201420c40ef801652df348e45270448c1a7abbef97b09b92bd7f858744b61653d96e0500ca52a78e7e45071e8923290284c3aad70f8b73151cdf514b9e29cc2b10c7e11ee290c21026a59b03f981a5dfded185472f4f19409cd229b5cfd11721e70f2b1d16079376a0b4195440d78")
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -500,
    "message": "Invalid",
    "data": "   at Neo.Plugins.RpcServer.GetRelayResult(VerifyResult reason, UInt256 hash)\r\n   at Neo.Plugins.RpcServer.SendRawTransaction(JArray _params)\r\n   at Neo.Plugins.RpcServer.ProcessRequest(HttpContext context, JObject request)"
  }
}
```

while invokescript return HALT.

```
{
  "jsonrpc": "2.0",
  "method": "invokescript",
  "params": ["110c149c09b202591267ecf38d953ae3b5b8d260ee693e0c14b13beae9e2edc2d1055d8116b37ab771c5c5467613c00c087472616e736665720c1425059ecb4878d3a875f91c51ceded330d4575fde41627d5b52", [{"account":"0x7646c5c571b77ab316815d05d1c2ede2e9ea3bb1", "scopes":"CalledByEntry"}]],
  "id": 1
}
```

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "script": "110c149c09b202591267ecf38d953ae3b5b8d260ee693e0c14b13beae9e2edc2d1055d8116b37ab771c5c5467613c00c087472616e736665720c1425059ecb4878d3a875f91c51ceded330d4575fde41627d5b52",
    "state": "HALT",
    "gasconsumed": "9007960",
    "stack": [
      {
        "type": "Boolean",
        "value": true
      }
    ],
    "tx": null
  }
}
```

shargon:
@cloud8little could you test the new api `calculatenetworkfee` ?

cloud8little:
@shargon calculatenetworkfee works. Test with invoke a nep5 contract deploy and name method. networkfee is same as sendrawtransaction.

cloud8little:
Please hold on merge since Rpcclient still need to be tested, calculatenetworkfee works as expected when integrate with neo.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -97,6 +114,16 @@ private JObject ImportPrivKey(JArray _params)
             };
         }
 
+        [RpcMethod]
+        private JObject CalculateNetworkFee(JArray _params)
+        {
+            byte[] tx = Convert.FromBase64String(_params[0].AsString());
+
+            JObject account = new JObject();
+            account["networkfee"] = (wallet ?? new DummyWallet()).CalculateNetworkFee(Blockchain.Singleton.GetSnapshot(), tx.AsSerializable<Transaction>());
```
cloud8little, line 123: This requires the user to open wallet? then rpcclient need to call openwallet first to get the right networkfee.

shargon: no, if it's not open, it will create a dummy wallet

cloud8little: then the networkfee is incorrect, and the tx will be invalid.

shargon: Cand you detail the error?

Tommo-L: https://github.com/neo-project/neo/blob/4066a4f3986a71612dcdeca33a4ecf8934511247/src/neo/Wallets/Wallet.cs#L373-L374

`GetAccount(hash)` returns empty.

cloud8little:
@shargon  how about opening a new PR relate with calculatenetworkfee? since it is not relate to this PR's title.

src/RpcServer/RpcServer.Wallet.cs
shargon: mmm... then is not a good solution, because you need to use the adresses of the rpc server...

Tommo-L: Maybe we should rename RpcClient to CSharpSDK, then we can open wallet and create transaction and relay it.

erikzhang: This way you can use Neo.dll directly.

shargon:
@cloud8little i will, give me three days.

shargon:
> how about opening a new PR relate with calculatenetworkfee? since it is not relate to this PR's title.

@cloud8little it's more or less the same because the problem appear upgrading the nuget

src/RpcServer/RpcServer.Wallet.cs
shargon: I think that the unique way is create the witnesses using the client wallet, and send them in the transaction, then with https://github.com/neo-project/neo/pull/1931 it could be easily computed.

shargon:
Please @cloud8little test it again

cloud8little:
> Please @cloud8little test it again

under test.

Tommo-L (approved):

PR #340, opened 2020-09-03
Allow querying contracts by name

### LINKED ISSUE #339: Add support for querying contracts by ID
The initial problem that led to this issue was that I wanted to make native contracts be discoverable via RPC. At the moment clients have to have GAS/NEO/Policy hashes to make any queries/transactions using them. Usually of course they don't change and these values can be hardcoded into the client. Still, if we're to make some changes (and to this date there is no preview version of Neo 3 that didn't change them) or if we're to deploy some private network with additional contract there is a question of client updates. The actual data is known to the RPC server, so if it could somehow convey this knowledge to the client that would be solved.

So I thought about some new API like `listnativecontracts`, but then remembered that we actually have now way more stable contract identifies, their ID numbers. And there is a simple scheme of numbering, going down from zero for native contracts and going up for regular ones.

If we're to enable `getcontractstate` RPC call to accept IDs (along with hashes) then we can make all our contracts easily enumerable and discoverable by clients. So the client may no longer store NEO hash, it can query contract `-1` and get it. Same thing for other contracts, native or regular ones.

- Plugins

shargon:
Close #339


src/RpcClient/Utility.cs
```diff
@@ -27,6 +28,21 @@ private static (BigInteger numerator, BigInteger denominator) Fraction(decimal d
...
         }
 
+        public static UInt160 ToScriptHash(this JObject value)
+        {
+            var addressOrScriptHash = value.AsString();
+
+            foreach (var native in NativeContract.Contracts)
+            {
+                if (addressOrScriptHash.Equals(native.Name, StringComparison.InvariantCultureIgnoreCase) ||
+                    addressOrScriptHash == native.Id.ToString())
```
roman-khimov, line 38: What about non-native contracts?

shargon: we can't fetch all the smartcontracts

roman-khimov: Ah, you don't keep mapping of IDs to hashes... OK then. We actually have that inside of neo-go, so I thought it'd be trivial to do the same thing here. But if at least it works for native ones that's good enough for now.

roman-khimov (dismissed):

cloud8little:
under test

erikzhang:
I think it should be in SDK, not in RPC.

shargon:
> I think it should be in SDK, not in RPC.

Done.

erikzhang:
@cloud8little Test.

cloud8little:
Regression test passed for getapplicationlog/getnep5transfers/getnep5balances/sendmany/sendfrom/sendtoaddress/invokefunction/invokescript/getunclaimedgas and getwalletbalance. 
for example:
```cs
        public static void TestSimpleToScriptHash()
        {
            string result = Utility.AsScriptHash("neo");
            Console.WriteLine(result);

            UInt160 neo_hash = Utility.ToScriptHash(result);
            Console.WriteLine(neo_hash.ToString());

            string gas_str = Utility.AsScriptHash("gas");
            Console.WriteLine(gas_str);

            UInt160 gas_hash = Utility.ToScriptHash(gas_str);
            Console.WriteLine(gas_hash.ToString());

            string policy = Utility.AsScriptHash("policy");
            Console.WriteLine(policy);

            UInt160 policy_result = Utility.ToScriptHash(policy);
            Console.WriteLine(policy_result.ToString());
        }
```

PR #342, opened 2020-09-07
Add Getcommittee

### LINKED ISSUE #341: Ability to get committee members via RPC
After the governance changes we now have a new entity, committee. It's an important entity in Neo but there is currently no easy way to get data about its members via RPC (counting candidates votes is not considered).

I see two main options here:
1) Add an explicit `getcommittee` RPC call that would return committee members (probably with some flag signifying validators).
2) Modify `getvalidators` RPC call, remove "active" field and add "incommittee" and "isvalidator" fields where `isvalidator` would have the same meaning as `active` previously and `incommittee` would be `true` for the members of committee.

But maybe there is some better way.

- RPC (HTTP)

cloud8little:
close #341

### DIFF
--- a/src/RpcClient/RpcClient.cs
+++ b/src/RpcClient/RpcClient.cs
@@ -256,6 +256,14 @@ namespace Neo.Network.RPC
             return ((JArray)RpcSend("getvalidators")).Select(p => RpcValidator.FromJson(p)).ToArray();
         }
 
+        /// <summary>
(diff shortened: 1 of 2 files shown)


shargon (commented):
Could you add it to rpc client too?

shargon (approved):

vncoelho (approved):

Tommo-L:
Merge?

Issue #361, opened 2020-10-08
[enhancement req] allow overriding RPC methods of RpcServer


ixje:
**Summary or problem description**
I want to suggest to make the RPC methods of the `RpcServer` plugin virtual so other plugins can extend them i.e. to collect usage metrics. Now one has to clone the whole plugin and do additions, and probably keep patching updates to the original RpcServer back into the clone.

**Where in the software does this update applies to?**
- RPC (HTTP)

ixje:
Thanks for the PR @Tommo-L and thanks for testing @superboyiii  🙌

PR #367, opened 2020-10-09
Allow rpc method can be overrided

Tommo-L:
Close https://github.com/neo-project/neo-modules/issues/361

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -98,7 +98,7 @@ namespace Neo.Plugins
         }
 
         [RpcMethod]
-        public virtual JObject InvokeFunction(JArray _params)
+        protected virtual JObject InvokeFunction(JArray _params)
         {
             UInt160 script_hash = UInt160.Parse(_params[0].AsString());
             string operation = _params[1].AsString();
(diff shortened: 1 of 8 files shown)


shargon (commented):
protected virtual?

Tommo-L:
> protected virtual?

Done

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -23,7 +23,7 @@ namespace Neo.Plugins
 {
     partial class RpcServer
     {
-        private class DummyWallet : Wallet
+        protected class DummyWallet : Wallet
```
shargon, line 26: I think that we can use private here

Tommo-L: Done

shargon (approved):

superboyiii:
Tested, works well on mine.

superboyiii (dismissed):

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -19,7 +19,7 @@ namespace Neo.Plugins
 {
     partial class RpcServer
     {
-        private class Signers : IVerifiable
+        protected class Signers : IVerifiable
```
shargon, line 22:
```suggestion
        private class Signers : IVerifiable
```

shargon (approved):

Issue #351, opened 2020-09-22
Remove sender from invoke* RPCs


roman-khimov:
**Summary or problem description**
neo-project/neo-modules#309 had introduced a `sender` parameter for `invokescript` and `invokefunction` RPCs, supposedly to solve the problem of `null` here:
```c#
tx = wallet.MakeTransaction(result["script"].AsString().HexToBytes(), null, witnessCosigners);
```

But it's not really needed as the sender is the first signer and `signers` are perfectly ordered to rely on this. At the moment having both an explicit `sender` and `signers` creates some duplication along with additional error modes that we need to check for. I don't see anything we couldn't do without an explicit `sender` (just with `signers`).

**Do you have any solution you want to propose?**
Something like this should solve the problem:
```c#
tx = wallet.MakeTransaction(result["script"].AsString().HexToBytes(), witnessSigners[0], witnessSigners);
```
And allow to simplify the RPC API (returning to the parameter set we had for preview3).

**Where in the software does this update applies to?**
- RPC (HTTP)

ixje:
Agreed. https://github.com/neo-project/neo/pull/1752 fixed the sender to the first signer.

Tommo-L:
I think it's clear enough.
https://github.com/neo-project/neo-modules/blob/fed3f7a87d65d91bbe4e42bce94734cc76571d5c/src/RpcServer/RpcServer.Wallet.cs#L168

ixje:
I don't understand your comment? 

The discussion is about removing the mandatory `sender` param for the `invokescript` and `invokefunction` RPC calls when you want to add signers, because `signer[0]` is already always the sender.

https://github.com/neo-project/neo-modules/blob/fed3f7a87d65d91bbe4e42bce94734cc76571d5c/src/RpcServer/RpcServer.SmartContract.cs#L120-L121

Tommo-L:
You are right, we should remove it. Plz see https://github.com/neo-project/neo-modules/pull/368

superboyiii:
> **Summary or problem description**
> #309 had introduced a `sender` parameter for `invokescript` and `invokefunction` RPCs, supposedly to solve the problem of `null` here:
> 
> ```cs
> tx = wallet.MakeTransaction(result["script"].AsString().HexToBytes(), null, witnessCosigners);
> ```
> 
[...]

PR #362, opened 2020-10-09
invoke* use base64 script

Tommo-L:
Fix https://github.com/neo-project/neo-modules/issues/337

PR #335, opened 2020-08-29
Pure async RpcClient

devhawk:
This PR plumbs removes all sync RpcClient calls and plumbs async thru end to end. 

If someone wants to block on an RpcClient call, they can use `.Result` (or `.GetAwaiter().GetResult()`) on the top level Task object.

~~Note, there are a few tests that are failing, so this PR isn't quite ready to merge yet. But I wanted to open it for review feedback since the core design is complete.~~ All the tests are passing now.

fixes #209

PR #368, opened 2020-10-09
Remove sender from invoke method

Tommo-L:
Fix https://github.com/neo-project/neo-modules/issues/351

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -104,7 +104,7 @@ namespace Neo.Plugins
             string operation = _params[1].AsString();
             ContractParameter[] args = _params.Count >= 3 ? ((JArray)_params[2]).Select(p => ContractParameter.FromJson(p)).ToArray() : new ContractParameter[0];
             Signers signers = _params.Count >= 4 ? SignersFromJson((JArray)_params[3]) : null;
-            
+
             byte[] script;
             using (ScriptBuilder sb = new ScriptBuilder())
             {
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -103,7 +103,7 @@ namespace Neo.Plugins
             UInt160 script_hash = UInt160.Parse(_params[0].AsString());
             string operation = _params[1].AsString();
             ContractParameter[] args = _params.Count >= 3 ? ((JArray)_params[2]).Select(p => ContractParameter.FromJson(p)).ToArray() : new ContractParameter[0];
-            Signers signers = _params.Count >= 4 ? SignersFromJson((JArray)_params[4]) : null;
+            Signers signers = _params.Count >= 4 ? SignersFromJson((JArray)_params[3]) : null;
             
             byte[] script;
             using (ScriptBuilder sb = new ScriptBuilder())
@@ -117,7 +117,7 @@ namespace Neo.Plugins
         private JObject InvokeScript(JArray _params)
         {
             byte[] script = _params[0].AsString().HexToBytes();
-            Signers signers = _params.Count > 1 ? SignersFromJson((JArray)_params[1]) : null;
+            Signers signers = _params.Count >= 2 ? SignersFromJson((JArray)_params[1]) : null;
             return GetInvokeResult(script, signers);
         }
 
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -61,7 +61,7 @@ namespace Neo.Plugins
             }
         }
 
-        private JObject GetInvokeResult(byte[] script, UInt160 sender = null, Signers signers = null)
+        private JObject GetInvokeResult(byte[] script, Signers signers = null)
         {
             using ApplicationEngine engine = ApplicationEngine.Run(script, container: signers, gas: settings.MaxGasInvoke);
             JObject json = new JObject();
@@ -76,7 +76,7 @@ namespace Neo.Plugins
             {
                 json["stack"] = "error: recursive reference";
             }
-            ProcessInvokeWithWallet(json, sender, signers);
+            ProcessInvokeWithWallet(json, signers);
             return json;
         }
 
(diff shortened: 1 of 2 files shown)


shargon (approved):

devhawk (changes requested):
Can we get our story straight on the sender parameter? @cloud8little just asked me to add the parameter to the new async RpcCLient PR: https://github.com/neo-project/neo-modules/pull/335#discussion_r502257495

Tommo-L:
> Can we get our story straight on the sender parameter? @cloud8little just asked me to add the parameter to the new async RpcCLient PR: [#335 (comment)](https://github.com/neo-project/neo-modules/pull/335#discussion_r502257495)

`Signers[0]` is the sender.

cloud8little:
why we make this change https://github.com/neo-project/neo-modules/pull/309 before?

cloud8little:
> Can we get our story straight on the sender parameter? @cloud8little just asked me to add the parameter to the new async RpcCLient PR: [#335 (comment)](https://github.com/neo-project/neo-modules/pull/335#discussion_r502257495)

@devhawk  I got it clear from https://github.com/neo-project/neo-modules/pull/368#pullrequestreview-505857571,  no need to add sender.

superboyiii:
> > Can we get our story straight on the sender parameter? @cloud8little just asked me to add the parameter to the new async RpcCLient PR: [#335 (comment)](https://github.com/neo-project/neo-modules/pull/335#discussion_r502257495)
> 
> `Signers[0]` is the sender.

Why `Signers[0]` should be the sender? `Signers[1]`, `Signers[2]` could also be sender, that's why we have to confirm the sender in params.

superboyiii:
Due to my test for this PR:
1. if sender is `Signers[0]` and `Signers[0]` doesn't have any `GAS`, then it will choose `Signers[1]` as sender(if `Signers[1]` has enough GAS) when `MakeTransaction`. 
2. if `Signers[0]` and `Signers[1]` both have enough GAS, sender will be `Signers[0]` even if I set `Signers[1]` as `Signer`.
***I think this change would make `sender` more confusion because we want `Signers[0]` to be sender, but in fact it still can't always be like this. Then why we have to set `Signers[0]` to be sender? Developer has their authority to choose which is the sender.***

roman-khimov:
> Developers have their authority to choose which is the sender.

Absolutely if you're to transfer signers given to "invoke*" into transaction 1:1. They'd get exactly the sender/signers they want this way.

superboyiii:
> > Developers have their authority to choose which is the sender.
> 
> Absolutely if you're to transfer signers given to "invoke*" into transaction 1:1. They'd get exactly the sender/signers they want this way.

Now I agree with you, just wait for solving https://github.com/neo-project/neo-modules/issues/370

superboyiii:
Let's make a summary:
1. `sender` should be from the input signer[].
2. If the input `signer[0]` hasn't enough GAS for make Tx, it could use `signer[1]` to be sender if it exist in the input signer[] and it has enough GAS as well. Or an error msg should be returned such as "Insufficient GAS".
3. `Signer[]` and `sender` should be from the input params but not from `wallet` class, should never automatically choosing an address from `wallet` even if the input `Signer[]` doesn't have enough GAS, the response should always be from what we input.

roman-khimov:
Just leave the first point and that's it. Second one is not necessary, it's up to the user of this API to order signers correctly, so if `signer[0]` doesn't have enough GAS --- it's not a problem of RPC server. Think also of RPC servers that don't manage wallets like C# one does, they just do 1:1 mapping from `invoke*` signers into transaction signers and don't care about balances at all, it's not their problem.

superboyiii:
> Just leave the first point and that's it. Second one is not necessary, it's up to the user of this API to order signers correctly, so if `signer[0]` doesn't have enough GAS --- it's not a problem of RPC server. Think also of RPC servers that don't manage wallets like C# one does, they just do 1:1 mapping from `invoke*` signers into transaction signers and don't care about balances at all, it's not their problem.

But current neo-cli is designed to care, we should solve this conflict in `wallet.cs`. Current `MakeTransaction` will check balance when invoking invokefunction, I think a new `MakeTransaction` should be created for this.

superboyiii:
Merged with latest, find it work well when signers[0] doesn't have enough gas.
```
{
  "jsonrpc": "2.0",
  "method": "invokefunction",
  "params": [
    "0x254b9decd76080ef368e7a6b0a065938dfbc31cf",
    "balanceOf",
    [
      {
        "type": "Hash160",
        "value": "0xf642a0401fdd56a03114639a98b7963eb587a30a"
      }
    ],
    [
      {
        "account": "0x7170793e96092211675c8333395d79c465a4f56f",
        "scopes": "CalledByEntry"
      },
      {
        "account": "0xf642a0401fdd56a03114639a98b7963eb587a30a",
        "scopes": "CalledByEntry"
      }
    ]
  ],
  "id": 3
}
```

```
{
    "jsonrpc": "2.0",
    "id": 3,
    "result": {
        "script": "DBQKo4e1Ppa3mJpjFDGgVt0fQKBC9hHADAliYWxhbmNlT2YMFM8xvN84WQYKa3qONu+AYNfsnUslQWJ9W1I=",
        "state": "HALT",
        "gasconsumed": "3840690",
        "exception": "Insufficient GAS",
        "stack": [
            {
                "type": "Integer",
                "value": "2000000000000000"
            }
        ]
    }
}
```

superboyiii (approved):

devhawk (approved):
👍

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -103,23 +103,22 @@ protected virtual JObject InvokeFunction(JArray _params)
...
         }
 
         [RpcMethod]
         protected virtual JObject InvokeScript(JArray _params)
         {
-            byte[] script = Convert.FromBase64String(_params[0].AsString());
-            UInt160 sender = _params.Count >= 2 ? AddressToScriptHash(_params[1].AsString()) : null;
-            Signers signers = _params.Count >= 3 ? SignersFromJson((JArray)_params[2]) : null;
-            return GetInvokeResult(script, sender, signers);
+            byte[] script = _params[0].AsString().HexToBytes();
```
ProDog, line 119: The modification of https://github.com/neo-project/neo-modules/pull/362 goes back here 🤔

Issue #355, opened 2020-09-27
Querying contracts by ID/name, server side


roman-khimov:
**Summary or problem description**
This is a follow-up to #339, I've overlooked somehow that #340 changed the client, but didn't change the server in any way, while the intention was really to have server answer the requests with ids/names. We want clients to not have any hashes to remember in their code, just point the client to the RPC server and it'll get all the data needed from it.

**Do you have any solution you want to propose?**
Implement support for #339 on the server side. We have a draft implementation of that in nspcc-dev/neo-go#1426.

**Where in the software does this update applies to?**
- RPC (HTTP)

PR #378, opened 2020-10-22
[RpcServer] Querying contracts by ID/name

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/355

- Fixed bug with decimal of GAS consumed in all RpcClient and RpcServer
- Querying contracts by ID/name, server side
- Fixed exception message error in ApplicationLog.


src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -132,7 +132,7 @@ private static Signers SignersFromJson(JArray _params)
         [RpcMethod]
         protected virtual JObject InvokeFunction(JArray _params)
         {
-            UInt160 script_hash = UInt160.Parse(_params[0].AsString());
+            UInt160 script_hash = UInt160.Parse(ToScriptHash(_params[0].AsString()));
```
roman-khimov, line 135: This is not #355, it's something different and probably not needed. #355 is about `getcontractstate`.

chenzhitong: ok

Tommo-L (approved):

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -103,11 +103,24 @@ protected virtual JObject GetBlockHeader(JArray _params)
...
             ContractState contract = Blockchain.Singleton.View.Contracts.TryGet(script_hash);
             return contract?.ToJson() ?? throw new RpcException(-100, "Unknown contract");
         }
 
+        public static string ToScriptHash(string addressOrScriptHash)
+        {
+            foreach (var native in NativeContract.Contracts)
+            {
+                if (addressOrScriptHash.Equals(native.Name, StringComparison.InvariantCultureIgnoreCase) ||
+                    addressOrScriptHash == native.Id.ToString())
```
shargon, line 116: https://github.com/neo-project/neo-modules/pull/340#issuecomment-687749790

chenzhitong: We only searched for native contracts, not other contracts. There's no harm in doing this, right?

roman-khimov: SDK is not a solution, RPC client is used in many scenarios, not just for software development. And we'd like to have it as little data as possible to function, that's why it's important to have the server replying to such requests. This particular change allows to remove native contract hashes from the client.

AnnaShaleva: Can we also add an ability to get non-native contracts hashes by ID here?

chenzhitong: Only native contracts can be searched by ID, not other contracts.  Because other contracts are stored as key-value pairs, they cannot be traversed. @annashaleva

Tommo-L (dismissed):

superboyiii:
Tested, it works well on mine.

superboyiii (approved):

roman-khimov:
This doesn't allow to run `invokescript` with `Verification` trigger which might be useful for non-contract verification scripts.

superboyiii:
> This doesn't allow to run `invokescript` with `Verification` trigger which might be useful for non-contract verification scripts.

Any suggestion or maybe we could merge it first?

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -67,7 +67,35 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
...
+            }
+            return json;
+        }
+
+        private JObject GetVerificationResult(UInt160 scriptHash, Signers signers = null)
+        {
+            var snapshot = Blockchain.Singleton.GetSnapshot();
+            var contract = snapshot.Contracts.TryGet(scriptHash);
+            var engine = ApplicationEngine.Create(TriggerType.Verification, new Transaction(), snapshot);
+            var context = engine.LoadScript(contract.Script, CallFlags.None, contract.Manifest.Abi.GetMethod("verify").Offset);
```
roman-khimov, line 92: To properly run `verify` we also need to execute `_initialize` as in `VerifyWitnesses`.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -67,7 +67,35 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
...
+            engine.LoadScript(Array.Empty<byte>(), CallFlags.None);
```
roman-khimov, line 93: What's this for? It's a no-op effectively.

roman-khimov: I think instead it should respect and properly pass `args` to the `verify` method.

roman-khimov (commented):
As for `invokescript`, maybe a separate `invokeverificationscript` is more appropriate here. A mandatory trigger parameter to `invokescript` seems to be excessive (as most invocations are expected to use `Application` trigger) and passing trigger among signers is not convenient to handle server-side. @ixje?

src/RpcServer/RpcServer.SmartContract.cs
ixje: seems like it was trying todo the `_init` but forgot a piece. Looks very similar to
https://github.com/neo-project/neo/blob/22f80a7ccfb00af15006569921ed87536cf3f857/src/neo/Wallets/Wallet.cs#L400-L404

ixje:
> As for `invokescript`, maybe a separate `invokeverificationscript` is more appropriate here. A mandatory trigger parameter to `invokescript` seems to be excessive (as most invocations are expected to use `Application` trigger) and passing trigger among signers is not convenient to handle server-side. @ixje?

Yeah that's what I was running into as well when I thought about it. If you want to keep the trigger optional while still part of `invokescript` then you have to traverse the JSON searching for the presence of a trigger type. I agree that a separate call sounds most appropriate here as I expect a `verification` trigger to be less than < 5% of the use-cases for calling an invoke script

chenzhitong:
`GetVerificationResult Method` I don't know if that's right.

roman-khimov:
@AnnaShaleva has experimented with it (and neo-project/neo#2052) a little: https://github.com/neo-project/neo-modules/commit/7f549c57b71afc601473dfb5d59d618fa3939838

But looking at it more and more I see that it's not that simple. The simplest possible signature checking verification script will fail on `invokescriptverify` because the execution environment lacks proper container:
```
        private bool VerifyWithECDsa(StackItem item, byte[] pubkey, byte[] signature, ECCurve curve)
        {
            ReadOnlySpan<byte> message = item switch
            {
                InteropInterface _interface => _interface.GetInterface<IVerifiable>().GetHashData(),
                Null _ => ScriptContainer.GetHashData(),

...

        public static byte[] GetHashData(this IVerifiable verifiable, uint magic)
        {
            using (MemoryStream ms = new MemoryStream())
            using (BinaryWriter writer = new BinaryWriter(ms))
            {
                writer.Write(magic);
                verifiable.SerializeUnsigned(writer);

...

        private class Signers : IVerifiable
        {
...
            public void SerializeUnsigned(BinaryWriter writer)
            {
                throw new NotImplementedException();
            }
```

And take a look at contract-based verification for Oracle contract, for example:
```
            Transaction tx = (Transaction)engine.ScriptContainer;
            return tx?.GetAttribute<OracleResponse>() != null;
```
It will fail also. So while we can run some scripts with verification trigger after this, it doesn't really solve the problem of calculating GAS for various verification scripts because they can easily fail without proper context.

Therefore, I have a bit different proposal. Instead of modifying `invokescript/invokefunction` we can make a completely separate method like `verifywitness(transaction, signer_index, witness) -> [state, gasconsumed, exception, stack]`, where `transaction` is a serialized transaction (it can be just `GetHashData()` of it, without witnesses), `signer_index` is an integer index of signer that we want to check (duplicating a hash doesn't make much sense, transaction has them all anyway) and `witness` is our regular witness with invocation and verification scripts (or without verification script for contract-based witnesses). It's just enough data for `TryCreateVerifyEngine()` to do its job. The output is basically the same as for `invoke*`, but without a `script` field as it doesn't make much sense in this context.

@ixje?

ixje:
Nice catch 👍 

> Therefore, I have a bit different proposal. Instead of modifying `invokescript/invokefunction` we can make a completely separate method like `verifywitness(transaction, signer_index, witness) -> [state, gasconsumed, exception, stack]`

What is the purpose of having both `transaction` and `signer_index`? Can't we make the signature `verifywitness(iverifiable, witness)` and then do `verifywitness(signer, witness)? From a quick glance that seems to provide sufficient information for `TryCreateVerifyEngine()`, no?

roman-khimov:
> What is the purpose of having both transaction and signer_index?

Transaction can have multiple signers and we're checking only one of them in a single run. Essentially, that's `UInt160 hash` parameter for `TryCreateVerifyEngine()`. It can of course be derived from verification script for ordinary witnesses, but this doesn't work for contract-based, because their `VerificationScript` is empty.

shargon:
Wait for https://github.com/neo-project/neo/pull/2052

roman-khimov:
According to neo-project/neo#1936 update, neo-project/neo#1977 is to be solved after preview4, can we move (small and simple) #355 fix into a separate PR to merge it before preview4?

shargon (approved):
The code looks good, but wait for @erikzhang's review because he was against moving this logic to the server.

superboyiii:
@erikzhang Could you review it again? It's useful.

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -103,11 +103,24 @@ protected virtual JObject GetBlockHeader(JArray _params)
         [RpcMethod]
         protected virtual JObject GetContractState(JArray _params)
         {
-            UInt160 script_hash = UInt160.Parse(_params[0].AsString());
+            UInt160 script_hash = UInt160.Parse(ToScriptHash(_params[0].AsString()));
             ContractState contract = Blockchain.Singleton.View.Contracts.TryGet(script_hash);
             return contract?.ToJson() ?? throw new RpcException(-100, "Unknown contract");
         }
 
+        public static string ToScriptHash(string keyword)
```
erikzhang, line 111:
```suggestion
        private static string ToScriptHash(string keyword)
```

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -103,11 +103,24 @@ protected virtual JObject GetBlockHeader(JArray _params)
...
+        {
+            foreach (var native in NativeContract.Contracts)
+            {
+                if (keyword.Equals(native.Name, StringComparison.InvariantCultureIgnoreCase) ||
```
erikzhang, line 115: https://github.com/neo-project/neo/blob/a1590b0995b201100b2153d2c0e638732eda0cee/src/neo/SmartContract/Native/NativeContract.cs#L101-L105

You can get native contract by name directly.

src/RpcServer/RpcServer.Blockchain.cs
erikzhang: And, why not return `UInt160` directly?

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -103,11 +103,24 @@ protected virtual JObject GetBlockHeader(JArray _params)
...
+                    keyword == native.Id.ToString())
+                    return native.Hash.ToString();
+            }
+
+            return keyword.Length < 40 ?
```
erikzhang, line 120: `== 40`?

src/RpcServer/RpcServer.Blockchain.cs
chenzhitong: I referenced this code.
https://github.com/neo-project/neo-modules/blob/master/src/RpcClient/Utility.cs#L39-L50

`NativeContract.GetContract(string name)` This method doesn't support searching by ID.

chenzhitong:
I modified the code, and the previous code referenced this
https://github.com/neo-project/neo-modules/blob/master/src/RpcClient/Utility.cs#L39-L50

src/RpcServer/RpcServer.Blockchain.cs
erikzhang: You can use `int.TryParse()` to check whether it is an ID or a name.

chenzhitong: If it's an ID, you also need to iterate through all NativeContracts. The code doesn't get any simpler.

src/ApplicationLogs/LogReader.cs
```diff
@@ -68,7 +68,9 @@ public void OnPersist(StoreView snapshot, IReadOnlyList<Blockchain.ApplicationEx
                 JObject trigger = new JObject();
                 trigger["trigger"] = appExec.Trigger;
                 trigger["vmstate"] = appExec.VMState;
-                trigger["exception"] = GetExceptionMessage(appExec.Exception);
+                var exceptionMessage = GetExceptionMessage(appExec.Exception);
+                if (exceptionMessage != null)
+                    trigger["exception"] = exceptionMessage;
```
Tommo-L, line 73:
```suggestion
                    trigger["exception"] = GetExceptionMessage(appExec.Exception);;
```

Tommo-L (dismissed):

superboyiii:
Retest: **PASS**

Invoke deploy method on nep17 SC:
```
neo> invoke 0x5ac8d09e4039571141b7b923e11fac46f0eddd81 deploy [] NLtDqwnj9s7wQVyaiD5ohjV3e9fUVkZxDp NLtDqwnj9s7wQVyaiD5ohjV3e9fUVkZxDp
Invoking script with: 'EMAMBmRlcGxveQwUgd3t8EasH+EjubdBEVc5QJ7QyFpBYn1bUg=='
VM State: HALT
Gas Consumed: 0.1869305
Result Stack: [{"type":"Integer","value":"1"}]
Relay tx(no|yes): y
Signed and relayed transaction with hash=0xa40cd1e197dbc4b588843b55412075872382e9a21ed4fb1c995df303eb586b05
```
Get ApplicationLogs

request:
```
{
  "jsonrpc": "2.0",
  "method": "getapplicationlog",
  "params": ["0xa40cd1e197dbc4b588843b55412075872382e9a21ed4fb1c995df303eb586b05"],
  "id": 1
} 
```
response:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xa40cd1e197dbc4b588843b55412075872382e9a21ed4fb1c995df303eb586b05",
        "executions": [
            {
                "trigger": "Application",
                "vmstate": "HALT",
                "exception": null,
                "gasconsumed": "18693050",
                "stack": [
                    {
                        "type": "Integer",
                        "value": "1"
                    }
                ],
                "notifications": [
                    {
                        "contract": "0x5ac8d09e4039571141b7b923e11fac46f0eddd81",
                        "eventname": "Transfer",
                        "state": {
                            "type": "Array",
                            "value": [
                                {
                                    "type": "Any"
                                },
                                {
                                    "type": "ByteString",
                                    "value": "CqOHtT6Wt5iaYxQxoFbdH0CgQvY="
                                },
                                {
                                    "type": "Integer",
                                    "value": "2000000000000000"
                                }
                            ]
                        }
                    }
                ]
            }
        ]
    }
}
```

superboyiii:
Retest: **PASS**

`gasconsumed` are all with decimals and RpcClinet also works well.

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xa40cd1e197dbc4b588843b55412075872382e9a21ed4fb1c995df303eb586b05",
        "executions": [
            {
                "trigger": "Application",
                "vmstate": "HALT",
                "exception": null,
                "gasconsumed": "0.1869305",
                "stack": [
                    {
                        "type": "Integer",
                        "value": "1"
                    }
                ],
                "notifications": [
                    {
                        "contract": "0x5ac8d09e4039571141b7b923e11fac46f0eddd81",
                        "eventname": "Transfer",
                        "state": {
                            "type": "Array",
                            "value": [
                                {
                                    "type": "Any"
                                },
                                {
                                    "type": "ByteString",
                                    "value": "CqOHtT6Wt5iaYxQxoFbdH0CgQvY="
                                },
                                {
                                    "type": "Integer",
                                    "value": "2000000000000000"
                                }
                            ]
                        }
                    }
                ]
            }
        ]
    }
}
```

ixje:
I haven't checked the latest code, but is the "gasconsumed" output format now still aligned across all functions? According to the examples in the latest online documentation
* [invokefunction](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokefunction.html) returns gas without decimals
* [invokescript](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokescript.html) returns gas without decimals

It would be good to make sure all functions return in the same format for a better developer experience.

superboyiii:
> I haven't checked the latest code, but is the "gasconsumed" output format now still aligned across all functions? According to the examples in the latest online documentation
> 
> * [invokefunction](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokefunction.html) returns gas without decimals
> * [invokescript](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokescript.html) returns gas without decimals
> 
> It would be good to make sure all functions return in the same format for a better developer experience.

Good question. We make "gasconsumed" output format with decimals in this PR. These documents are still for Preview3, we'll update it after Preview4 being ready.

cloud8little:
> > I haven't checked the latest code, but is the "gasconsumed" output format now still aligned across all functions? According to the examples in the latest online documentation
> > 
> > * [invokefunction](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokefunction.html) returns gas without decimals
> > * [invokescript](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/invokescript.html) returns gas without decimals
> > 
> > It would be good to make sure all functions return in the same format for a better developer experience.
> 
> Good question. We make "gasconsumed" output format with decimals in this PR. These documents are still for Preview3, we'll update it after Preview4 being ready.

It's better to update [getwalletunclaimedgas](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/getwalletunclaimedgas.html) and getunclaimedgas too.

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "9500000000"
}
```

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "unclaimed": "10150000000",
    "address": "Ncu47NyTJLJPEFi859gmuqQ8fPPt1AVwHX"
  }
}

```

src/ApplicationLogs/LogReader.cs
```diff
@@ -173,7 +174,7 @@ public bool ShouldThrowExceptionFromCommit(Exception ex)
 
         static string GetExceptionMessage(Exception exception)
         {
-            if (exception == null) return "Engine faulted.";
+            if (exception == null) return null;
```
shargon, line 177: We want `{ "exception": null } ` or `{}`?

chenzhitong: @Tommo-L said `{ "exception": null }` ✅ `{}` ❌

superboyiii: > We want `{ "exception": null } ` or `{}`?

How about:
` if (GetExceptionMessage(appExec.Exception)!= null)
            {
                trigger["exception"] = GetExceptionMessage(appExec.Exception);
            }
`

shargon (approved):

superboyiii (approved):

src/RpcServer/Utility.cs
```diff
@@ -0,0 +1,28 @@
...
+using Neo.Network.P2P.Payloads;
+using Neo.SmartContract.Native;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Neo.Plugins
+{
+    public static class Utility
```
shargon, line 11: Helper?

chenzhitong: It's called Utility in RpcClient, and I've just adopted it.

shargon (approved):

PR #383, opened 2020-10-30
Use Base64String on response of getrawtransaction

superboyiii:
Use Base64String on response of getrawtransaction

### DIFF
--- a/src/RpcServer/RpcServer.Blockchain.cs
+++ b/src/RpcServer/RpcServer.Blockchain.cs
@@ -147,7 +147,7 @@ namespace Neo.Plugins
                 }
                 return json;
             }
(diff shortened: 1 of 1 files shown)


shargon (approved):

roman-khimov:
If we're going this way, I think adjusting `getblock`, `getheader` and `getstorage` would also make sense.

superboyiii:
> If we're going this way, I think adjusting `getblock`, `getheader` and `getstorage` would also make sense.

I fix this because `calculatenetworkfee` is already using base64 tx raw as input param but still need manual base64 conversion from the returned hexstring of `getrawtransaction`. But you're right, I will unify these soon.

Issue #398, opened 2020-11-13
RPCServer: `invoke*` RPC requests failures


AnnaShaleva:
**Describe the bug**
https://github.com/neo-project/neo-modules/pull/390 broke `invokefunction` and `invokescript` RPC requests handlers. In the case of missing signers, the response to these requests is the following:
```
{
   "jsonrpc" : "2.0",
   "id" : 1,
   "error" : {
      "message" : "Object reference not set to an instance of an object.",
      "code" : -2147467261
   }
}
```

The reason is that the default value for `signers` is `null` ([here](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.SmartContract.cs#L116), [here](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.SmartContract.cs#L130) and [here](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.SmartContract.cs#L64)) which results in `NullReferenceException` right [here](https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.SmartContract.cs#L68).

**To Reproduce**
Steps to reproduce the behaviour:
1. Send the following RPC request to the Neo node with RPCServer plugin installed from the current [master](https://github.com/neo-project/neo-modules/commit/f43c5abfeafae49d8ed84f4112a8f3fdd0c69c10):
```
curl -d '{ "jsonrpc": "2.0", "id": 1, "method": "invokefunction", "params": ["0x668e0c1f9d7b70a99dd9e06eadd4c784d641afbc", "name"]}' localhost:20331 | json_pp
```
2. The result response is the following:
```
{
   "jsonrpc" : "2.0",
   "id" : 1,
   "error" : {
      "message" : "Object reference not set to an instance of an object.",
      "code" : -2147467261
   }
}
```
**Expected behaviour**
Request should be correctly processed and GAS native contract name should be returned:
```
{
   "result" : {
      "stack" : [
         {
            "value" : "R0FT",
            "type" : "ByteString"
         }
      ],
      "gasconsumed" : "1007570",
      "state" : "HALT",
      "script" : "EMAMBG5hbWUMFLyvQdaEx9StbuDZnalwe50fDI5mQWJ9W1I=",
      "exception" : null
   },
   "jsonrpc" : "2.0",
   "id" : 1
}
```

**Additional context**
[...]

PR #400, opened 2020-11-16
[RpcServer] Update RpcServer.SmartContract.cs

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/398

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -63,7 +63,7 @@ namespace Neo.Plugins
 
         private JObject GetInvokeResult(byte[] script, Signers signers = null)
         {
-            Transaction tx = new Transaction
+            Transaction tx = signers == null ? null : new Transaction
             {
                 Signers = signers?.GetSigners(),
                 Attributes = Array.Empty<TransactionAttribute>(),
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -65,9 +65,9 @@ namespace Neo.Plugins
         {
             Transaction tx = signers == null ? null : new Transaction
             {
-                Signers = signers?.GetSigners(),
+                Signers = signers.GetSigners(),
                 Attributes = Array.Empty<TransactionAttribute>(),
-                Witnesses = signers?.Witnesses,
+                Witnesses = signers.Witnesses,
             };
             using ApplicationEngine engine = ApplicationEngine.Run(script, container: tx, gas: settings.MaxGasInvoke);
             JObject json = new JObject();
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -65,9 +65,9 @@ namespace Neo.Plugins
         {
             Transaction tx = new Transaction
             {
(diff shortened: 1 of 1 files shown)


shargon (changes requested):
I think that this properties can't be null or compute the `Hash` will fault.

chenzhitong:
> I think that this properties can't be null or compute the `Hash` will fault.

So we should make `signers` a normal parameter instead of an optional parameter?
``` c#
GetInvokeResult(byte[] script, Signers signers = null)
```

superboyiii:
> I think that this properties can't be null or compute the `Hash` will fault.

It could be null. If it's null, just do nothing in `ProcessInvokeWithWallet`.
https://github.com/neo-project/neo-modules/blob/2ae6a9722b7a6a031943f92866d53c89b18b3433/src/RpcServer/RpcServer.Wallet.cs#L168

cloud8little:
I think default  signers value should not be null when calling invokefunction and invokescript. It should be Array.Empty<Signer>().

https://github.com/neo-project/neo-modules/blob/2ae6a9722b7a6a031943f92866d53c89b18b3433/src/RpcServer/RpcServer.SmartContract.cs#L130

https://github.com/neo-project/neo-modules/blob/2ae6a9722b7a6a031943f92866d53c89b18b3433/src/RpcServer/RpcServer.SmartContract.cs#L116

shargon:
If `Hash` FAULT, It can FAULT here https://github.com/neo-project/neo/blob/2d0e6123a1f9e58c8069f0762d17c6e5eec2dde9/src/neo/Network/P2P/Payloads/Transaction.cs#L325

superboyiii:
> If `Hash` FAULT, It can FAULT here https://github.com/neo-project/neo/blob/2d0e6123a1f9e58c8069f0762d17c6e5eec2dde9/src/neo/Network/P2P/Payloads/Transaction.cs#L325

Hash will not be computed if `signer` is null. 
https://github.com/neo-project/neo-modules/blob/f13d7c305f02f77aa48cc579f678a96e4f319a25/src/RpcServer/RpcServer.Wallet.cs#L171

shargon:
Could you test it calling `InvokeFunction` to a method that return the hash of the current transaction?
https://github.com/neo-project/neo-modules/blob/8f03a7f78fa0de02385574b4aedc28b21c0d81c3/src/RpcServer/RpcServer.SmartContract.cs#L116

superboyiii:
> Could you test it calling `InvokeFunction` to a method that return the hash of the current transaction (without signer)?
> https://github.com/neo-project/neo-modules/blob/8f03a7f78fa0de02385574b4aedc28b21c0d81c3/src/RpcServer/RpcServer.SmartContract.cs#L116

Do you mean return hash in the internal of SC? If in this situation, it will truely return fault.

Tommo-L:
In some cases, we don't need to set `Signer`, like invoke `balanceOf` method.  And if the invoke return faults, the user need to check it.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -65,9 +65,9 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
         {
             Transaction tx = new Transaction
```
shargon, line 66:
```suggestion
            Transaction tx = signers == null ? null : new Transaction
```

shargon (approved):

PR #414, opened 2020-11-27
MPT support latest state

ZhangTao1596:
Move https://github.com/neo-project/neo/pull/1996 here.

Changes:
 * Add `full_state` parameter when construct `MPTTrie`. If  it sets `false`, old state tree nodes will be removed from database when update tree. Before, We keep all tree nodes to retain the ability to provide all historical state.
 * Add `Reference` in `MPTNode` as reference count to  avoid deleting nodes used by other branches. We only store one copy of node with same hash in database. For example, if  A and B both have 1 gas, we only store `00e1f505` once.
 * Add `cache` when mpt read and write `ISnapshot`.


Tommo-L:
Conflicts

ZhangTao1596:
Merge?

erikzhang:
If it is well tested, then merge it.

ZhangTao1596:
UnitTests are nearly 100%. Will ask @superboyiii to have a systematic test.

shargon (approved):

shargon (commented):
Could you include the tests in github actions?
It wasn't added in coverage report https://coveralls.io/builds/35523583

ZhangTao1596:
> Could you include the tests in github actions?
> It wasn't added in coverage report https://coveralls.io/builds/35523583

Done.

superboyiii:
> If it is well tested, then merge it.

We could merge it first and wait for another PR implementing it on functions, then we could make a systematic test. I'll not release it on Preview4, but since there're compile issues on stateservice and wait for this to fix(https://github.com/neo-project/neo-modules/issues/424), we'd better merge it before Preview4.

shargon (approved):
For me, we can merge it and test it in deep

Issue #424, opened 2020-12-11
RpcServer doesn't compile


devhawk:
Note, the compile issues in StateService are tracked by #415.  RpcServer compile was broken by #409 

```
C:\Users\harry\Source\neo\official\modules\src\StateService\MPT\MPTNodeType.cs(7,10): error CS0404: Cannot apply attribute class 'ReflectionCache<T>' because it is generic [C:\Users\harry\Source\neo\official\modules\src\StateService\StateService.csproj]
C:\Users\harry\Source\neo\official\modules\src\StateService\MPT\MPTNodeType.cs(9,10): error CS0404: Cannot apply attribute class 'ReflectionCache<T>' because it is generic [C:\Users\harry\Source\neo\official\modules\src\StateService\StateService.csproj]
C:\Users\harry\Source\neo\official\modules\src\StateService\MPT\MPTNodeType.cs(11,10): error CS0404: Cannot apply attribute class 'ReflectionCache<T>' because it is generic [C:\Users\harry\Source\neo\official\modules\src\StateService\StateService.csproj]
C:\Users\harry\Source\neo\official\modules\src\StateService\MPT\MPTNodeType.cs(13,10): error CS0404: Cannot apply attribute class 'ReflectionCache<T>' because it is generic [C:\Users\harry\Source\neo\official\modules\src\StateService\StateService.csproj]
C:\Users\harry\Source\neo\official\modules\src\RpcServer\RpcServer.SmartContract.cs(113,37): error CS1061: 'SnapshotView' does not contain a definition for 'Contracts' and no accessible extension method 'Contracts' accepting a first argument of type 'SnapshotView' could be found (are you missing a using directive or an assembly reference?) [C:\Users\harry\Source\neo\official\modules\src\RpcServer\RpcServer.csproj]
```

superboyiii:
> Note, the compile issues in StateService are tracked by #415. RpcServer compile was broken by #409
> 
> ```
> C:\Users\harry\Source\neo\official\modules\src\StateService\MPT\MPTNodeType.cs(7,10): error CS0404: Cannot apply attribute class 'ReflectionCache<T>' because it is generic [C:\Users\harry\Source\neo\official\modules\src\StateService\StateService.csproj]
[...]

PR #420, opened 2020-12-10
Fixed UT

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/419

PR #425, opened 2020-12-14
Fix GetContract

superboyiii:
Close https://github.com/neo-project/neo-modules/issues/424


superboyiii:
Close this, because https://github.com/neo-project/neo-modules/pull/420 is already there.

Issue #437, opened 2020-12-22
Add CI to test build of each module


vncoelho:
Add a command in the workflow to automatically check the build of each module.

`dotnet publish /src/$PLUGIN_TO_INCLUDE/$PLUGIN_TO_INCLUDE.csproj -c Release -o /src/$PLUGIN_TO_INCLUDE/app`

chenzhitong:
![image](https://user-images.githubusercontent.com/6295602/102965592-19936d80-4529-11eb-8a28-cdfab41dcd4a.png)

PR #439, opened 2020-12-23
Add CI to test build of each module

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/437


.github/workflows/main.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       run: |
         dotnet tool install --version 5.0.142902 --tool-path ./ dotnet-format --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
         ./dotnet-format . --check -v diagnostic
+    - name: Publish
+      run: |
+        dotnet publish neo-modules.sln
```
erikzhang, line 28: I think it is not necessary. We have used `dotnet test` already. It will fail if build fails.

vncoelho: Does it test all src or just those linked to the test?

vncoelho:
I will double check if dotnet test build all src.

Issue #426, opened 2020-12-14
deployed contract name support in getContractState


devhawk:
**Summary or problem description**
[getcontractstate](https://docs.neo.org/v3/docs/en-us/reference/rpc/latest-version/api/getcontractstate.html) supports retrieving contract state by invocation hash (i.e. the new script hash that doesn't change when the contract is updated), or by native contract name or native contract id. There is no way to retrieve the contract state of a non-native deployed contract except by invocation hash.

**Do you have any solution you want to propose?**
Update GetContractState to support query of deployed contracts by name. Note, since it's possible for multiple contracts to have the same name, this means returning an array of contract states when querying by name.

Neo-express is creating a custom express-specific RPC endpoint with this behavior (https://github.com/neo-project/neo-express/pull/92). But it would be best for neo visual devtracker for this behavior to be standard on all RPC nodes.

**Where in the software does this update applies to?**
- RPC (HTTP)

Tommo-L:
> Update GetContractState to support query of deployed contracts by name. 

It seems we can not do that, as there is no `name -> contractstate` table in neo-core.

cloud8little:
> > Update GetContractState to support query of deployed contracts by name.
> 
> It seems we can not do that, as there is no `name -> contractstate` table in neo-core.

we have list all contracts function, can filter the result by name.
```cs
        public IEnumerable<ContractState> ListContracts(StoreView snapshot)
        {
            byte[] listContractsPrefix = CreateStorageKey(Prefix_Contract).ToArray();
            return snapshot.Storages.Find(listContractsPrefix).Select(kvp => kvp.Value.GetInteroperable<ContractState>());
        }
```

devhawk:
Sorry, clicked the close button by accident

PR #441, opened 2020-12-23
Deployed contract name support in getContractState

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/426


src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -105,20 +105,16 @@ protected virtual JObject GetBlockHeader(JArray _params)
...
         {
-            foreach (var native in NativeContract.Contracts)
-            {
-                if (keyword.Equals(native.Name, StringComparison.InvariantCultureIgnoreCase) || keyword == native.Id.ToString())
-                    return native.Hash;
-            }
-
-            return UInt160.Parse(keyword);
+            return NativeContract.Contracts.FirstOrDefault(p => keyword.Equals(p.Name, StringComparison.InvariantCultureIgnoreCase) || keyword == p.Id.ToString())?.Hash ??
+                   NativeContract.ContractManagement.ListContracts(snapshot).FirstOrDefault(p => keyword.Equals(p.Manifest.Name, StringComparison.InvariantCultureIgnoreCase) || keyword == p.Id.ToString())?.Hash ??
```
shargon, line 116: It will fetch all the smart contracts, we should test it with at least 100k smart contracts. In order to avoid a possible Denial of Service

superboyiii: > It will fetch all the smart contracts, we should test it with at least 100k smart contracts. In order to avoid a possible Denial of Service

Maybe we could move `getcontractstate` to somewhere like ApplicationLogs, then it could get contract state indenpently from the beginning just with a little adjustment on that plugin?

cloud8little: @devhawk  what do you think? if it is possible to cause Denial of Service, we need to reconsider what the API benefits. Otherwise there is no conclusion on this.

erikzhang:
I don't think it is possible to map name to contract. Because the name is not unique.

src/RpcServer/RpcServer.Blockchain.cs
devhawk: I understand the need for a stable contract identifier, but the choice of generating it from a combination of the original deployed contract hash + the original contract deployer address is not very developer friendly. I think there needs to be a mechanism for developers to be able to retrieve information about their deployed contracts w/o having to understand the algorithm that generates the identification hash. 

If we don't make this change to getContractState, another option could be to introduce a new contract search RPC method that returns an array of all contacts identification hashes (along with associated metadata) that matches a given name.

chenzhitong:
Is it okay to do what he said? @devhawk 

request:
```
{
  "jsonrpc": "2.0",
  "method": "getcontractstate",
  "params": ["hello"],
  "id": 1
}
```

response:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": [
        "0x99042d380f2b754175717bb932a911bc0bb0ad7d",
        "0x1bc09179180b0ad7bb932540f2b7717bd542d39a"
    ]
}
```

devhawk:
I don't think `getcontractstate` should return different data depending on the input parameters as you've mocked up @chenzhitong. I'd rather introduce a new RPC method. 

Does the thing we used to call "script hash" have a new name? I called it "invocation hash" in #426 but I just made that up on the fly. I'd suggest `getinvocationhashesbyname` (or whatever we end up calling it) as the name of the new RPC method.

ProDog:
For native contract, we supports get contract state by hash, name or id, so I think for the other contracts should also be the same, or support, or don't support name and id.

chenzhitong:
Related PR https://github.com/neo-project/neo-modules/pull/462

bettybao1209:
> Related PR #462

It seems the issue https://github.com/neo-project/neo-modules/issues/426 has not been solved.

devhawk:
> It seems the issue #426 has not been solved.

Agreed @bettybao1209

devhawk:
(sorry, reopened by accident)

Issue #443, opened 2020-12-28
RpcClient and RpcServer don't compile against neo master


devhawk:
RpcClient errors:

```
modules\src\RpcClient\ContractClient.cs(44,62): error CS1503: Argument 3: cannot convert from 'object[]' to 'bool' [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\ContractClient.cs(60,20): error CS1061: 'ScriptBuilder' does not contain a definition for 'EmitAppCall' and no accessible extension method 'EmitAppCall' accepting a first argument of type 'ScriptBuilder' could be found (are you missing a using directive or an assembly reference?) [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\Nep17API.cs(80,28): error CS7036: There is no argument given that corresponds to the required formal parameter 'hasReturnValue' of 'Helper.MakeScript(UInt160, string, bool, params object[])' [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\Nep17API.cs(81,28): error CS7036: There is no argument given that corresponds to the required formal parameter 'hasReturnValue' of 'Helper.MakeScript(UInt160, string, bool, params object[])' [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\Nep17API.cs(82,28): error CS7036: There is no argument given that corresponds to the required formal parameter 'hasReturnValue' of 'Helper.MakeScript(UInt160, string, bool, params object[])' [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\Nep17API.cs(113,63): error CS1503: Argument 3: cannot convert from 'Neo.UInt160' to 'bool' [modules\src\RpcClient\RpcClient.csproj]
modules\src\RpcClient\Nep17API.cs(140,63): error CS1503: Argument 3: cannot convert from 'Neo.UInt160' to 'bool' [modules\src\RpcClient\RpcClient.csproj]
```

RpcServer
```
modules\src\RpcServer\RpcServer.SmartContract.cs(128,20): error CS7036: There is no argument given that corresponds to the required formal parameter 'hasReturnValue' of 'ApplicationEngine.LoadContract(ContractState, string, CallFlags, bool, ushort)' [modules\src\RpcServer\RpcServer.csproj]
[...]

PR #444, opened 2020-12-29
Fix compilation bug

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/443


chenzhitong:
Only one UT left to fix

shargon (approved):

Issue #433, opened 2020-12-21
Preview4 GetNextBlockValidators returning empty


vncoelho:
**Describe the bug**
result of RPC call `GetNextBlockValidators` is `[]` with test on NeoCompiler Eco V3 network

cloud8little:
It is because in https://github.com/neo-project/neo-modules/pull/338 getvalidators API was renamed to getnextblockvalidators, but the meaning of the API was unchanged.

vncoelho:
Hi @cloud8little, I did not understand.
I got that it was renamed. 
It looks like we still need to fix its call, right?

cloud8little:
> Hi @cloud8little, I did not understand.
> I got that it was renamed.
> It looks like we still need to fix its call, right?

yes, we should fix it. @Tommo-L

chenzhitong:
Ready to fix, blocked by https://github.com/neo-project/neo-modules/issues/440

PR #445, opened 2020-12-29
Fix the bug that getnextblockvalidators return empty when there is no votes.

chenzhitong:
close https://github.com/neo-project/neo-modules/issues/433


src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -197,14 +197,29 @@ protected virtual JObject GetNextBlockValidators(JArray _params)
...
+                    return validator;
+                }).ToArray();
+            }
+            else
+            {
+                return validators.Select(p =>
+                {
+                    JObject validator = new JObject();
+                    validator["publickey"] = p.ToString();
+                    validator["votes"] = 0;
```
chenzhitong, line 218: 0 or 100000000 whis is better?

shargon: 0 it's fine for me

shargon (approved):

vncoelho (commented):
I think it should return the Default Commit instead.
Otherwise it will be strange to have "0" and, in fact, we actually have the next validators.

vncoelho (commented):
Sorry, I think it is good like this, perhaps the "active" will contain the current validators.

src/RpcServer/RpcServer.Blockchain.cs
cloud8little: I think this API should be sync with the result that neo-cli client returns. getnextblockvalidators should return nextblockvalidators, and we can have GetCandidates to return the candidates information.

Tommo-L (approved):

Issue #450, opened 2021-01-08
Default RpcServer gas limit too low for light clients to request deploy system fee


ixje:
**Describe the bug**
Lights wallet SDKs depend on RPC servers for obtaining the system fee for building their transactions. There is a 10 GAS limit on the `invokescript` method as set by https://github.com/neo-project/neo-modules/blob/a1bdf84079a2d413a15388dff51b1b1203fdbfdb/src/RpcServer/RpcServer/config.json#L10 

An attempt to request the system fee to deploy this very small contract

```python
from boa3.builtin import public

def main() -> int:
    return 1

@public
def test_func() -> int:
    return 2

@public
def test_func2(value: int) -> int:
    return 1 + value
``` 

[contract_nef_and_manifest.zip](https://github.com/neo-project/neo-modules/files/5788869/contract_nef_and_manifest.zip)

already caused `invokescript` to run out of gas. The required gas is just above 10 (to be exact; 10.01014750). `neo-cli` doesn't suffer from this issue, because it has the full chain and invokes the engine with 20 gas. 

**To Reproduce**
call `invokescript` RPC function with a script that attemps to deploy a contract.

**Expected behavior**
I expect light SDKs to be able to obtain the transaction system fee via invokescript. I think perhaps 15 or 20 default gas limit is more reasonable, although it could be calculated exactly by taking the Nef and manifest max file sizes and calculating the system fee required in that such an extreme case. 

I propose to increase the default gas limit because the list of RPC servers already isn't that big and if light wallets have to rely on custom configurations it will limit the available servers even more leading to a worse user experience. 

**Platform:**
n/a

ixje:
@shargon thanks for the quick response ❤️

PR #452, opened 2021-01-10
Increase RPC.MaxGasInvoke

shargon:
Close https://github.com/neo-project/neo-modules/issues/450


erikzhang:
Why do you change the default value?

shargon:
@erikzhang @ixje has troubles with this limit on his test https://github.com/neo-project/neo-modules/issues/450

erikzhang (approved):

PR #454, opened 2021-01-10
Fix UT

erikzhang:


erikzhang:
Please fix the UT. 😅

erikzhang:
It seems that the RPC UT depends on the `http://seed1.neo.org:10331`. I think this cannot be applied to the latest update.

superboyiii:
> It seems that the RPC UT depends on the `http://seed1.neo.org:10331`. I think this cannot be applied to the latest update.

Never mind. It's just a mock data, will not create actual request to `http://seed1.neo.org:10331`.

Issue #458, opened 2021-01-12
Error on `invokescript` with a contract deployment script


csmuller:
**Describe the bug**
When calling `invokescript` with a script that deploys a new contract, I get the error `Exception has been thrown by the target of an invocation.`. When sending the script via `sendrawtransaction`, the deployment is successful.
This issue only arises when i use a `_deploy` method in the contract.

**To Reproduce**

1. Compile a dummy contract with a `_deploy()` method. E.g.:

```
using Neo.SmartContract.Framework;

public class Contract : SmartContract
{

    public static void _deploy(bool update)
    {
    }
}
```

2. Build the deployment script.  (Or deploy it with neo-cli and get the deployment script with getrawtransaction.)

3. Call `invokescript` with the deployment script as an argument (and a signer).

**Expected behavior**
Success response with the correct GAS amount consumed by the deployment.

**Platform:**
 - OS: MacOS BigSur
 - Neo Version: neo3-preview4

vncoelho:
The same here,@csmuller. It is good that you reported.
I have been discussing with @superboyiii.

Let's keep this in mind because this is important for us to test the current master.

bettybao1209:
Hi, I reproduced your case, but I can get the expected result. Here is my use case:

```c#
{
  "jsonrpc": "2.0",
  "method": "invokescript",
[...]

PR #463, opened 2021-01-13
get message of base exception

bettybao1209:
fix https://github.com/neo-project/neo-modules/issues/458

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -206,8 +206,7 @@ namespace Neo.Plugins
 
         static string GetExceptionMessage(Exception exception)
         {
-            if (exception == null) return null;
-            return exception.GetBaseException().Message;
+            return exception?.GetBaseException().Message;
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -207,13 +207,7 @@ protected virtual JObject GetUnclaimedGas(JArray _params)
...
         {
             if (exception == null) return null;
-
-            if (exception.InnerException != null)
-            {
-                return exception.InnerException.Message;
-            }
-
-            return exception.Message;
+            return exception.GetBaseException().Message;
```
shargon, line 210: It's recursive?

bettybao1209: https://docs.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-5.0

It's used to get the root cause of the exception, which may  tell the more detailed information.

chenzhitong (dismissed):

chenzhitong:
And here also need to change https://github.com/neo-project/neo-modules/blob/3fdbcf8bd4acfe2c15cf5e447c9e99995ca1bf8e/src/ApplicationLogs/LogReader.cs#L166-L176

chenzhitong (approved):

shargon (approved):

PR #467, opened 2021-01-16
Fix FormatException for systems with European culture (comma decimal separator)

### LINKED ISSUE #466: Decimal toString() calls should include InvariantCulture
**Describe the bug**
Currently on systems with an European culture is set not using the InvariantCulture might result in Decimal.ToString() to result in something like: "123,45" instead of "123.45". This then results in a parsing error when using in BigDecimal.Parse() for example.

**To Reproduce**
1. Set your system's culture to something like 'nl-NL' (so your system's decimal becomes a comma).
2. Run a project which uses the RPC Server (for example: [neo-express](https://github.com/neo-project/neo-express)).
3. See error logging for format exception.

**Expected behavior**
Independent of the culture of the system where NEO projects are ran, parsing should work as expected.

**Screenshots**

**Platform:**
 - OS: Windows 10 x64
 - Version: v3.0.0-preview4

robinabganpat:
Fix #466

### DIFF
--- a/tests/Neo.Network.RPC.Tests/UT_RpcClient.cs
+++ b/tests/Neo.Network.RPC.Tests/UT_RpcClient.cs
@@ -13,6 +13,7 @@ using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using Neo.Plugins;
 
 namespace Neo.Network.RPC.Tests
 {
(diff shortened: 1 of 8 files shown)


tests/Neo.Network.RPC.Tests/UT_RpcClient.cs
```diff
@@ -78,6 +79,13 @@ public void TestConstructorByUrlAndDispose()
             action.Should().NotThrow<Exception>();
         }
 
+        [TestMethod]
+        public void TestRpcServerSettingsConstructorForFormatException()
```
erikzhang, line 83: I think this should not be in `UT_RpcClient.cs`.

robinabganpat: I thought so. Mind if I create a `UT_RpcServer.cs` and move the test there?

robinabganpat: I've moved the test to a separate test project. Given the current setup of the unit tests projects, I've decided to move it to a separate Neo.Plugins.RpcServer.Tests folder.

shargon (approved):

erikzhang (approved):

PR #470, opened 2021-01-18
[StateService] Verification

ZhangTao1596:
### Step 2/3 StateService
Since there are too many changes in StateService, I will split it into 3 parts to commit pull request, and this is the 2nd.
  - [x] Storage and P2P message (Merged)
  - [x] Verification
  - [x] More command & RPC
  
Changes:
* Add Verification service to get consensus of state root from verifiers.
     * Cache `MaxCachedValidationProcess· Verification processes because block grows without state root.
     * Node use latest state root to compare state.
     * Not consensused state root of height will be dropped when block grows.
 * Add RPC method to exchange consensus message.
 * Add command to start Verification service.


src/StateService/Settings.cs
```diff
@@ -1,18 +1,21 @@
 using Microsoft.Extensions.Configuration;
+using System.Linq;
 
 namespace Neo.Plugins.StateService
 {
     internal class Settings
     {
         public string Path { get; }
         public bool FullState { get; }
+        public string[] Validators { get; }
```
shargon, line 10: `Uri` ?

src/StateService/StatePlugin.cs
```diff
@@ -15,27 +19,52 @@ public class StatePlugin : Plugin, IPersistencePlugin
...
         {
             StateStore.Singleton.UpdateLocalStateRoot(block.Index, snapshot.Storages.GetChangeSet().Where(p => p.State != TrackState.None).ToList());
         }
+
+        [RpcMethod]
+        public JObject VoteStateRoot(JArray _params)
+        {
+            uint height = uint.Parse(_params[0].AsString());
+            int validator_index = int.Parse(_params[1].AsString());
+            byte[] sig = _params[2].AsString().HexToBytes();
```
shargon, line 53: Base64 for binary

src/StateService/StatePlugin.cs
```diff
@@ -15,27 +19,52 @@ public class StatePlugin : Plugin, IPersistencePlugin
...
+            Validator?.Tell(new Vote(height, validator_index, sig));
+            return true;
+        }
+
+        [ConsoleCommand("start validate", Category = "StateService", Description = "Start as a state validator if wallet is open")]
+        private void OnStartValidate()
+        {
+            var wallet_provider = GetService<IWalletProvider>();
+            if (wallet_provider is null)
```
shargon, line 62: It can't be null, but when you call `GetWallet` it can be

src/StateService/Validation/ValidationProcess.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (signatures.ContainsKey(index)) return false;
+            Utility.Log(nameof(ValidationService), LogLevel.Info, $"vote received, index={index}");
+            ECPoint validator = validators[index];
+            byte[] hash_data = StateRoot?.GetHashData();
+            if (hash_data != null && !Crypto.VerifySignature(hash_data, sig, validator))
+            {
+                Utility.Log(nameof(ValidationService), LogLevel.Info, "incorrect vote, invalid signature");
+                return false;
+            }
+            signatures.Add(index, sig);
```
shargon, line 96: If hash_data it's null we should return `false`?

src/StateService/Validation/ValidationProcess.cs
```diff
@@ -0,0 +1,147 @@
...
+
+        public bool AddSignature(int index, byte[] sig)
+        {
+            if (M <= signatures.Count) return false;
+            if (index < 0 || validators.Length <= index) return false;
+            if (signatures.ContainsKey(index)) return false;
+            Utility.Log(nameof(ValidationService), LogLevel.Info, $"vote received, index={index}");
+            ECPoint validator = validators[index];
+            byte[] hash_data = StateRoot?.GetHashData();
+            if (hash_data != null && !Crypto.VerifySignature(hash_data, sig, validator))
```
shargon, line 91:
```suggestion
            if (hash_data == null || !Crypto.VerifySignature(hash_data, sig, validator))
```

src/StateService/Settings.cs
vncoelho: `MPTValidators` `MPTDispachers` `MPTVerifiers`
I think it would be better to chose a name that does not cross with dBFT `Validators`

src/StateService/StatePlugin.cs
```diff
@@ -15,27 +19,52 @@ public class StatePlugin : Plugin, IPersistencePlugin
         public override string Name => "StateService";
         public override string Description => "Enables MPT for the node";
 
-        private IActorRef store;
+        internal IActorRef Store;
+        internal IActorRef Validator;
```
vncoelho, line 23: I think this also should have another name instead of `Validator`
It is acting like a `MPTDispacher`, or  `MPTAgent`

src/StateService/Settings.cs
ZhangTao1596: > `Uri` ?

Yes. Like Oracle.

src/StateService/StatePlugin.cs
ZhangTao1596: Will do.

src/StateService/StatePlugin.cs
ZhangTao1596: Let's call it `Dispacher`.

src/StateService/Settings.cs
ZhangTao1596: > `MPTValidators` `MPTDispachers` `MPTVerifiers`
> I think it would be better to chose a name that does not cross with dBFT `Validators`

Let's take `MPTVerifier`.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/Validation/ValidationProcess.cs
ZhangTao1596: Done.

src/StateService/Settings.cs
ZhangTao1596: Done.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/Settings.cs
shargon: I mean change the type https://github.com/neo-project/neo-modules/pull/470/commits/70e7201a241abbca0492a261520f339fe85da1e7

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,140 @@
...
+        }
+        public ExtensiblePayload Message => payload;
+
+        public VerificationContext(Wallet wallet, uint index)
+        {
+            this.wallet = wallet;
+            Initialize(index);
+        }
+
+        public void Initialize(uint index)
```
Qiao-Jin, line 53: private?

ZhangTao1596: Done.

shargon (commented):
Tested?

ZhangTao1596:
@superboyiii Fixed https://github.com/neo-project/neo-modules/pull/475#issuecomment-766368446

superboyiii:
> Tested?

Give me 2 days, I'm testing these two intergratively.

superboyiii:
Mark a persist issue.
I set a single CN environment and I set a verification node and an ordinary node for state root as well.
The issue is when verification node clear its chain data, resync and start verifying state root at the beginning, ongoing verification tasks will be stashed until the max amount -> 10. Then when verifying node sync to the latest height, the last 10 state root will be relay, vote and persist. But at the same time, new state root will come to tasks so there will generate tasks continuously until the ongoing is 10. So the problem is persist cost some time which will make the other ordinary nodes' state lost behind for a while but not increase linearly. I'm setting 15 seconds for consensus, a little worry about if multi CNs and less consensus time will make this result worse. Next, I'll try multi CNs to see if it influence a lot. If not, we could improve it later after Preview5.
image
For this result, I show state height every 2 seconds.

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,147 @@
...
+        {
+            var vote = context.CreateVote();
+            if (vote is null) return;
+            if (context.Verifiers.Length == 1) CheckVotes(context);
+            Utility.Log(nameof(VerificationService), LogLevel.Info, $"relay vote");
+            Parallel.ForEach(Settings.Default.VerifierUrls, (url, state, i) =>
+            {
+                try
+                {
+                    var client = new RpcClient(url);
```
vncoelho, line 41: This should be here or on the RpcClient RPC?

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,147 @@
...
+                    value.Timer.CancelIfNotNull();
+                }
+            }
+        }
+
+        private void OnTimer(uint index)
+        {
+            if (contexts.TryGetValue(index, out VerificationContext context))
+            {
+                SendVote(context);
```
vncoelho, line 112: Any node can send vote?

src/StateService/Settings.cs
```diff
@@ -1,18 +1,22 @@
...
         public bool FullState { get; }
+        public Uri[] VerifierUrls { get; }
 
         public static Settings Default { get; private set; }
 
         private Settings(IConfigurationSection section)
         {
             Path = string.Format(section.GetValue("Path", "Data_MPT_{0}"), ProtocolSettings.Default.Magic.ToString("X8"));
             FullState = section.GetValue("FullState", false);
+            VerifierUrls = section.GetSection("VerifierUrls").GetChildren().Select(p => new Uri(p.Get<string>(), UriKind.Absolute)).ToArray();
```
vncoelho, line 19: Who are and who define this `VerifierUrls`? How to update it?

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,136 @@
...
+        public ExtensiblePayload Message => payload;
+
+        public VerificationContext(Wallet wallet, uint index)
+        {
+            this.wallet = wallet;
+            myIndex = -1;
+            root = null;
+            rootIndex = index;
+            using SnapshotCache snapshot = Blockchain.Singleton.GetSnapshot();
+            verifiers = NativeContract.RoleManagement.GetDesignatedByRole(snapshot, Role.StateValidator, index);
```
vncoelho, line 56: I got it now. It will be automatic with those that are published with the `Role.StateValidator`.
This is exactly what I was expecting.

Now, I need to understand how to update it.

vncoelho: Who are the first `StateValidator`, do the committee need to call `designateAsRole` or there will be default nodes such as `Validators`, `Committee` and `Oracles` as default MPT?

ZhangTao1596: I think committee need call `designateAsRole ` to designate validators.

vncoelho: In my initial tests here I got an error after the Genesis.

Perhaps,as default we should have some node to do that,even if it is a single node.

ZhangTao1596: > In my initial tests here I got an error after the Genesis.

What error? Is it in this pr?

> Perhaps,as default we should have some node to do that,even if it is a single node.
> I mean,before the committee vote.

We should discuss this in neo-core.

src/StateService/StatePlugin.cs
```diff
@@ -22,29 +23,59 @@ public class StatePlugin : Plugin, IPersistencePlugin
...
         {
             base.Dispose();
-            System.EnsureStoped(store);
+            System.EnsureStoped(Store);
+            if (Verifier != null) System.EnsureStoped(Verifier);
         }
 
         void IPersistencePlugin.OnPersist(Block block, DataCache snapshot, IReadOnlyList<ApplicationExecuted> applicationExecutedList)
         {
             StateStore.Singleton.UpdateLocalStateRoot(block.Index, snapshot.GetChangeSet().Where(p => p.State != TrackState.None).ToList());
```
ZhangTao1596, line 49: @shargon If we call in this way. It will affect the performance of block persisting.

src/StateService/StatePlugin.cs
```diff
@@ -22,29 +23,59 @@ public class StatePlugin : Plugin, IPersistencePlugin
...
+        public JObject VoteStateRoot(JArray _params)
+        {
+            uint height = uint.Parse(_params[0].AsString());
+            int validator_index = int.Parse(_params[1].AsString());
+            byte[] sig = Convert.FromBase64String(_params[2].AsString());
+            Verifier?.Tell(new Vote(height, validator_index, sig));
+            return true;
+        }
+
+        [ConsoleCommand("start verifying state", Category = "StateService", Description = "Start as a state verifier if wallet is open")]
```
shargon, line 62:
```suggestion
        [ConsoleCommand("start states", Category = "StateService", Description = "Start as a state verifier if wallet is open")]
``` ?

File: src/StateService/StatePlugin.cs [hunk 14]
```diff
@@ -22,29 +23,59 @@ public class StatePlugin : Plugin, IPersistencePlugin
...
 
         void IPersistencePlugin.OnPersist(Block block, DataCache snapshot, IReadOnlyList<ApplicationExecuted> applicationExecutedList)
         {
             StateStore.Singleton.UpdateLocalStateRoot(block.Index, snapshot.GetChangeSet().Where(p => p.State != TrackState.None).ToList());
         }
 
+        [RpcMethod]
+        public JObject VoteStateRoot(JArray _params)
+        {
+            uint height = uint.Parse(_params[0].AsString());
```
shargon, line 55: User must know what's happend without debug. We should provide more error detail, if we have less than 3 params => error, if `Verifier` it's null => error.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,147 @@
...
+            var vote = context.CreateVote();
+            if (vote is null) return;
+            if (context.Verifiers.Length == 1) CheckVotes(context);
+            Utility.Log(nameof(VerificationService), LogLevel.Info, $"relay vote");
+            Parallel.ForEach(Settings.Default.VerifierUrls, (url, state, i) =>
+            {
+                try
+                {
+                    var client = new RpcClient(url);
+                    client?.RpcSendAsync("votestateroot", vote.RootIndex, vote.ValidatorIndex, vote.Signature.ToHexString())
```
shargon, line 42:
```suggestion
                    client.RpcSendAsync("votestateroot", vote.RootIndex, vote.ValidatorIndex, vote.Signature.ToHexString())
```

shargon (commented):
it's full tested?

Tommo-L:
Not yet, it's still under testing.

superboyiii:
image
I've tried 4 state validators and close one node manually. Finally fail to send vote, is it designed to need full signatures collection?  I supposed more than 2/3 is enough at the beginning.

ZhangTao1596:
> image
> I've tried 4 state validators and close one node manually. Finally fail to send vote, is it designed to need full signatures collection? I supposed more than 2/3 is enough at the beginning.

Fixed. You can continue.

src/StateService/Verification/VerificationService.cs
```diff
@@ -39,7 +39,7 @@ private void SendVote(VerificationContext context)
                 try
                 {
                     var client = new RpcClient(url);
-                    client.RpcSendAsync("votestateroot", vote.RootIndex, vote.ValidatorIndex, vote.Signature.ToHexString())
+                    client?.RpcSendAsync("votestateroot", vote.RootIndex, vote.ValidatorIndex, Convert.ToBase64String(vote.Signature))
```
shargon, line 42: client can't be null

ZhangTao1596: Recovered.

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,139 @@
...
+using Neo.SmartContract.Native;
+using Neo.Wallets;
+using System;
+using System.Collections.Generic;
+
+namespace Neo.Plugins.StateService.Verification
+{
+    public class VerificationContext
+    {
+        private const uint MaxValidUntilBlockIncrement = 100;
```
Tommo-L, line 20:
```suggestion
        private static const uint MaxValidUntilBlockIncrement = 100;
```

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,153 @@
...
+{
+    public class VerificationService : UntypedActor
+    {
+        public class ValidatedRootPersisted { public uint Index; }
+        public class BlockPersisted { public uint Index; }
+        private class Timer { public uint Index; }
+        private static readonly uint TimeoutMilliseconds = ProtocolSettings.Default.MillisecondsPerBlock;
+        private static readonly uint DelayMilliseconds = 3000;
+        private readonly NeoSystem core;
+        private const int MaxCachedVerificationProcessCount = 10;
```
Tommo-L, line 21:
```suggestion
        private static readonly int MaxCachedVerificationProcessCount = 10;
```

and reorder

src/StateService/Verification/VerificationContext.cs
erikzhang: Why?

src/StateService/Verification/VerificationService.cs
erikzhang: ?

src/StateService/Verification/VerificationContext.cs
Tommo-L: I misunderstood. `const` is enough. `static readonly` can be modified.

ProDog:
I tested it with multiple state nodes locally, it works fine, then I will create multiple nodes in the cloud server and test.

Tommo-L (dismissed):

ProDog:
I have tested it with 4 consensus nodes and 1-6 state nodes on the cloud server,  it works fine, I think we're ready to merge.

superboyiii (dismissed):
@shargon @erikzhang Passed test, need your final review.

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,139 @@
...
+        private StateRoot root;
+        private ExtensiblePayload payload;
+        private readonly Wallet wallet;
+        private readonly KeyPair keyPair;
+        private readonly int myIndex;
+        private readonly uint rootIndex;
+        private readonly ECPoint[] verifiers;
+        private int M => verifiers.Length - (verifiers.Length - 1) / 3;
+        private readonly Dictionary<int, byte[]> signatures = new Dictionary<int, byte[]>();
```
shargon, line 29:
```suggestion
using Akka.Actor;
using Neo.Cryptography;
using Neo.Cryptography.ECC;
using Neo.IO;
using Neo.Ledger;
using Neo.Network.P2P;
using Neo.Network.P2P.Payloads;
using Neo.Persistence;
using Neo.Plugins.StateService.Storage;
using Neo.SmartContract;
using Neo.SmartContract.Native;
using Neo.Wallets;
using System;
using System.Collections.Concurrent;

namespace Neo.Plugins.StateService.Verification
{
    public class VerificationContext
    {
        private const uint MaxValidUntilBlockIncrement = 100;
        private StateRoot root;
        private ExtensiblePayload payload;
        private readonly Wallet wallet;
        private readonly KeyPair keyPair;
        private readonly int myIndex;
        private readonly uint rootIndex;
        private readonly ECPoint[] verifiers;
        private int M => verifiers.Length - (verifiers.Length - 1) / 3;
        private readonly ConcurrentDictionary<int, byte[]> signatures = new ConcurrentDictionary<int, byte[]>();
```

Use concurrent?

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,139 @@
...
+            Utility.Log(nameof(VerificationContext), LogLevel.Info, $"vote received, height={rootIndex}, index={index}");
+            ECPoint validator = verifiers[index];
+            byte[] hash_data = StateRoot?.GetHashData();
+            if (hash_data is null || !Crypto.VerifySignature(hash_data, sig, validator))
+            {
+                Utility.Log(nameof(VerificationContext), LogLevel.Info, "incorrect vote, invalid signature");
+                return false;
+            }
+            signatures.Add(index, sig);
+            return true;
```
shargon, line 95:
```suggestion
            return signatures.TryAdd(index, sig);
```

src/StateService/Verification/VerificationContext.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        private void CreateExtensiblePayload()
```
shargon, line 98: Move this method to `CheckSignatures` in order to return false if sc it's not complete?

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,162 @@
...
+                CheckVotes(context);
+            }
+        }
+
+        private void CheckVotes(VerificationContext context)
+        {
+            if (context.CheckSignatures())
+            {
+                if (context.Message is null) return;
+                var state_root = context.Message.Data.AsSerializable<StateRoot>();
```
shargon, line 73: Maybe we can save this serialization caching it in `VerificationContext`

src/StateService/Verification/VerificationContext.cs
ZhangTao1596: Why use concurrent, it is only in one actor.

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Done.

src/StateService/Verification/VerificationContext.cs
ZhangTao1596: Done. I remove the cath. Here we should throw exception beacuse it is not expected to happen. If it does, there must be  a bug.

src/StateService/Verification/VerificationContext.cs
shargon: `VerificationService` use Concurrent for his dictionary and it use this class

ZhangTao1596: I don't think we need concurrent in `VerificationService`, do we?

ZhangTao1596: Done.

src/StateService/Verification/VerificationContext.cs
ZhangTao1596: Done.

superboyiii:
@erikzhang @shargon Please review again.

ZhangTao1596:
Merge?

vncoelho (commented):
There is something strange,
Once I activate this plugin the screen do not allow to be stopped on linux server with `ctrl+c`.
Multiple nodes are running together and the ones with this plugin activated suffers this.

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,161 @@
...
+                Task.Run(() =>
+                {
+                    try
+                    {
+                        var http_client = new HttpClient
+                        {
+                            Timeout = TimeSpan.FromMilliseconds(TimeoutMilliseconds),
+                        };
+                        var client = new RpcClient(http_client, url);
+                        client.RpcSendAsync("votestateroot", vote.RootIndex, vote.ValidatorIndex, Convert.ToBase64String(vote.Signature))
```
vncoelho, line 49: Who are able to send Votes? Which nodes can connect to the Verifies and send valid votes?

vncoelho: I also did not fully understand who will determine the Verifiers.

Tommo-L: > Initially its empty right until we vote and fill NativeContract.RoleManagement.GetDesignatedByRole.?

Yes


> I also did not fully understand who will determine the Verifiers.


The state verifier is designated by the committee.

vncoelho: The Default values on the  `config.json` are just used to start a communication between them, right?

vncoelho: In this sense, we would consider that the `VerifierUrls` is fixed to `4` and `Committee` would keep this number as `4` to be consistent with the Default `Config`.

Finally, thus, a normal user will just start the service with `start states` and it will sync normally and then the user will have a personalized RPC, right?

src/StateService/StateService/config.json
```diff
@@ -1,6 +1,12 @@
 {
   "PluginConfiguration": {
     "Path": "Data_MPT_{0}",
-    "FullState": false
+    "FullState": false,
+    "VerifierUrls": [
+      "http://state1.ngd.network:10332",
+      "http://state2.ngd.network:10332",
+      "http://state3.ngd.network:10332",
+      "http://state4.ngd.network:10332"
```
Tommo-L, line 9: Maybe we should set empty here @ZhangTao1596 @vncoelho

vncoelho: Maybe @Tommo-L, I still need to understand some aspects. 
Thanks for the previous explanations.

If you have some time to take a look at my last comment on our previous thread.

src/StateService/Verification/VerificationService.cs
ZhangTao1596: > Who are able to send Votes? Which nodes can connect to the Verifies and send valid votes?

Normally only verifiers send votes, but anyone can. Because it's a rpc service, anyone can connect and send once the verifiers' rpc port exposed. But only verifiers vote is valid.

> Initially its empty right until we vote and fill `NativeContract.RoleManagement.GetDesignatedByRole`.?

The `VerifierUrls` is not saved in `RoleManagement` .Committee only designate verifier public keys. Verifier urls should be manully changed in config.json of this service.

> Finally, thus, a normal user will just start the service with `start states` and it will sync normally and then the user will have a personalized RPC, right?

Normal user don't need to start states when sync. They can state states but there will be nothing because it's not verifier which are designated in `RoleManagement`.

ZhangTao1596:
> There is something strange,
> Once I activate this plugin the screen do not allow to be stopped on linux server with `ctrl+c`.
> Multiple nodes are running together and the ones with this plugin activated suffers this.

Did you install oracle plugin? As @ProDog tested, orcale plugin will cause this can't exit problem.

src/StateService/StateService/config.json
ZhangTao1596: Done.

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,161 @@
...
+                Utility.Log(nameof(VerificationService), LogLevel.Info, $"relay state root, height={context.StateRoot.Index}, root={context.StateRoot.RootHash}");
+                core.Blockchain.Tell(context.Message);
+            }
+        }
+
+        private void OnBlockPersisted(uint index)
+        {
+            if (MaxCachedVerificationProcessCount <= contexts.Count)
+            {
+                var indexes = contexts.Keys.OrderBy(i => i).ToArray();
```
shargon, line 82: Use list?

shargon (approved):
I think that the code it's ok

src/StateService/Verification/VerificationService.cs
erikzhang: I think `ReadOnlySpan<T>` is better.

vncoelho:
hi @ZhangTao1596, you are correct, I forgot to delete the `OraclePlugin` just for those clients, makes sense.
I did not test without it, but the analysis is precise.

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Done.

src/StateService/Verification/VerificationService.cs
vncoelho: @ZhangTao1596, thus, how the system can scale up?

I mean, a project running on Neo that wants to follow the `states` needs to start and sync, right?
Then it would have its personalized RPC for querying data from.
Otherwise, the officials RPC workload would be too much.

src/StateService/Verification/VerificationService.cs
```diff
@@ -79,13 +79,14 @@ private void OnBlockPersisted(uint index)
...
+                var indexes = contexts.Keys.OrderBy(i => i).ToList();
+                while (MaxCachedVerificationProcessCount <= indexes.Count)
                 {
-                    if (contexts.TryRemove(indexes[0], out var value))
+                    var i = indexes.First();
+                    if (contexts.TryRemove(i, out var value))
                     {
                         value.Timer.CancelIfNotNull();
-                        indexes = indexes[..1];
+                        indexes.Remove(i);
```
shargon, line 89: i?

shargon (changes requested):

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Yes. Problem?

src/StateService/Verification/VerificationService.cs
ZhangTao1596: > I think `ReadOnlySpan<T>` is better.

This `var indexes = new ReadOnlySpan<uint>(contexts.Keys.OrderBy(i => i).ToArray());`?

erikzhang: Done.

src/StateService/Verification/VerificationService.cs
erikzhang: Fixed.

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,160 @@
...
+            {
+                ReadOnlySpan<uint> indexes = contexts.Keys.OrderBy(i => i).ToArray();
+                while (MaxCachedVerificationProcessCount <= indexes.Length)
+                {
+                    if (contexts.TryRemove(indexes[0], out var value))
+                    {
+                        value.Timer.CancelIfNotNull();
+                        indexes = indexes[1..];
+                    }
+                }
```
shargon, line 89:
```suggestion
                int i = 0;
                while (MaxCachedVerificationProcessCount <= (indexes.Length - i))
                {
                    if (contexts.TryRemove(indexes[i], out var value))
                    {
                        value.Timer.CancelIfNotNull();
                        i++;
                    }
                }
```

src/StateService/Verification/VerificationService.cs
ZhangTao1596:
```C#
            if (MaxCachedVerificationProcessCount <= contexts.Count)
            {
                contexts.Keys.OrderBy(p => p).Take(contexts.Count - MaxCachedVerificationProcessCount + 1).ForEach(p =>
                {
                    if (contexts.TryRemove(p, out var value))
                    {
                        value.Timer.CancelIfNotNull();
                    }
                });
            }
```
@erikzhang  How about this？

src/StateService/Verification/VerificationService.cs
ZhangTao1596:
```C#
        if (MaxCachedVerificationProcessCount <= contexts.Count)
        {
            contexts.Keys.OrderBy(p => p).Take(contexts.Count - MaxCachedVerificationProcessCount + 1).ForEach(p =>
            {
                if (contexts.TryRemove(p, out var value))
                {
                    value.Timer.CancelIfNotNull();
                }
            });
        }
``` 
@shargon How about this？

src/StateService/Verification/VerificationService.cs
erikzhang: Good.

ZhangTao1596: Done.

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Done.

shargon: It's ok for me

src/StateService/Verification/VerificationService.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                Task.Run(() =>
+                {
+                    try
+                    {
+                        var http_client = new HttpClient
+                        {
+                            Timeout = TimeSpan.FromMilliseconds(TimeoutMilliseconds),
+                        };
+                        var client = new RpcClient(http_client, url);
```
shargon, line 47: We should use `ExtensiblePayload` with p2p, but it could be done in a different PR

shargon (approved):

src/StateService/Verification/VerificationService.cs
ZhangTao1596: > I mean, a project running on Neo that wants to follow the `states` needs to start and sync, right?
> Then it would have its personalized RPC for querying data from.
> Otherwise, the officials RPC workload would be too much.

Right. Normal node installed state service will have state releated rpc service. They don't need to query verifier or official rpc nodes.
Futhermore normal nodes don't need to `start states`. Once the plugin installed, sync and rpc service will work. Considering mpt calculating, node should install this plugin before block syncing.
Verifiers use `start states` to start verifying.

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Yes. Using p2p, We won't need url and avoid inconsistence between urls and public keys.
I have code here https://github.com/ZhangTao1596/neo-modules/tree/state-service-p2p.

@Tommo-L  worries that use p2p will put more pressure on p2p. And he propose add service discovery function.
Once we make agreement, I will commit another pr.

erikzhang: If we use service discovery, will it expose the IP address?

src/StateService/Verification/VerificationService.cs
vncoelho: @ZhangTao1596, sounds good.

In this sense, @ZhangTao1596.
Maybe there should be  warning if the plugin is installed in a node that already has synced some blocks.
Perhaps this could be done in another PR.

Thanks you and @Tommo-L for the clarifications.

Great work!

vncoelho (approved):

src/StateService/Verification/VerificationService.cs
ZhangTao1596: Can we merge this for now. I will make p2p improvement  pr later.

Tommo-L: >  use p2p will put more pressure on p2p. 

Yes


> will it expose the IP address?

Maybe we can encrypt part of the service information so that only the relevant node can decrypt it, like:

```
service type + service information

state: xxxxxxx 
oracle: yyyyyyy
rpc: 192.168.1.1
...
```

We can consider it later, after this pr.

ZhangTao1596: @Tommo-L Can you create a issue of your idea. We can discuss there.  Since oracle and state both need that.

Tommo-L: Oracle can't use p2p message for siging!

Tommo-L: We can discuss here https://github.com/neo-project/neo-modules/issues/493

ProDog:
Merge?

PR #477, opened 2021-01-21
Neo 3.0.0-CI01168

erikzhang:
Close https://github.com/neo-project/neo-modules/issues/479


erikzhang:
Please fix the UT.

cloud8little:
Fail to create genesis block:

a345bee63e5f38595ab73d8024a48f8

f3bc2668c9286350a877dd68fb2c0e6

src/StateService/MPT/MPTCache.cs
```diff
@@ -30,13 +31,25 @@ public MPTCache(ISnapshot store, byte prefix)
...
+        {
+            byte[] buffer = new byte[UInt256.Length + 1];
+            using (MemoryStream ms = new MemoryStream(buffer, true))
+            using (BinaryWriter writer = new BinaryWriter(ms))
+            {
+                writer.Write(prefix);
+                hash.Serialize(writer);
+            }
+            return buffer;
+        }
```
ZhangTao1596, line 44:
```suggestion
        private byte[] Key(UInt256 hash)
        {
            return Concat(new byte[] { prefix }, hash.ToArray());
        }
```
Like this?

src/StateService/Storage/Keys.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Buffers.Binary;
+
+namespace Neo.Plugins.StateService.Storage
+{
+    public static class Keys
+    {
+        public static byte[] StateRoot(uint index)
+        {
+            byte[] buffer = new byte[sizeof(uint) + 1];
+            buffer[0] = 1;
```
ZhangTao1596, line 11:
```suggestion
            buffer[0] = 0xf1;
```
Let's use `0xf1` otherwise it will conflict with prefixes in core.

src/StateService/Storage/Keys.cs
```diff
@@ -0,0 +1,19 @@
...
+            BinaryPrimitives.WriteUInt32BigEndian(buffer.AsSpan(1), index);
+            return buffer;
+        }
+
+        public static readonly byte[] CurrentLocalRootIndex = { 0x02 };
```
ZhangTao1596, line 16:
```suggestion
        public static readonly byte[] CurrentLocalRootIndex = { 0xf2 };
```

src/StateService/Storage/Keys.cs
```diff
@@ -0,0 +1,19 @@
...
+        public static readonly byte[] CurrentValidatedRootIndex = { 0x04 };
```
ZhangTao1596, line 17:
```suggestion
        public static readonly byte[] CurrentValidatedRootIndex = { 0xf4 };
```

src/StateService/MPT/MPTCache.cs
erikzhang: It's less efficient.

src/StateService/Storage/Keys.cs
erikzhang: There is no prefix in core. So it will conflict anyway. That's why we use a separate database.

erikzhang:
Ready for review.

chenzhitong (approved):

shargon (approved):

PR #496, opened 2021-02-01
fix insufficient networkfee in RpcClient

bettybao1209:
The decimal of network fee in RpcClient and RpcServer is different, resulting in `insuffient funds` exception.

### DIFF
--- a/src/RpcClient/RpcClient.cs
+++ b/src/RpcClient/RpcClient.cs
@@ -5,6 +5,7 @@ using Neo.Network.P2P.Payloads;
 using Neo.Network.RPC.Models;
 using Neo.SmartContract;
 using Neo.SmartContract.Manifest;
+using Neo.SmartContract.Native;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -264,7 +265,7 @@ namespace Neo.Network.RPC
         {
             var json = await RpcSendAsync(GetRpcName(), Convert.ToBase64String(tx.ToArray()))
                 .ConfigureAwait(false);
-            return (long)json["networkfee"].AsNumber();
+            return (long)BigDecimal.Parse(json["networkfee"].AsString(), NativeContract.GAS.Decimals).Value;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


erikzhang:
I think use floating-point numbers in JSON is not safe. Especially in the case of expressing prices.

bettybao1209:
> I think use floating-point numbers in JSON is not safe. Especially in the case of expressing prices.

If not, it may be knida confusing for users when invoking `calculatenetworkfee` in RPCServer.

erikzhang:
We should uniformly use `long` to ensure accuracy.

bettybao1209:
> We should uniformly use `long` to ensure accuracy.

But currently we use float value for json output like json["gasconsumed"].

erikzhang:
> But currently we use float value for json output like json["gasconsumed"].

I think we should use `long` for JSON output.

bettybao1209:
So should we uniform both the user input and the JSON output in `long` type?

erikzhang:
For JSON, I think we should use `long` for both requests and response.

chenzhitong:
> For JSON, I think we should use `long` for both requests and responses.

The asset amount(fees and NEP-17 tokens) is now in `decimal` format in both neo-cli and RpcServer.

**neo-cli command:**
deploy
invoke
balanceof
totalsupply
transfer
etc.

**RpcServer：**
 calculatenetworkfee
 getapplicationlog
 getblock
 getblockheader
 getrawtransaction
 getunclaimedgas
 getwalletunclaimedgas
 invokecontractverify
 invokefunction
 invokescript
 sendfrom
 sendmany
 sendtoaddress
 getnep17balances
 getnep17transfers
 getwalletbalance

roman-khimov:
> I think use floating-point numbers in JSON is not safe.

They really are, we had problems with rounding errors.

> For JSON, I think we should use long for both requests and responses.

But we can wrap them into strings, returning `"value": "0.12345678"` instead of `"value": 0.12345678`. This representation is much more convenient to me (especially for everything that is not GAS or NEO where the number of decimals is well-known, it's not for $RANDOM_TOKEN). And it avoids any losses when converting to/from JSON.

I mean `BigDecimal.ToString()` doesn't introduce any rounding errors and neither does `TryParse()`, so we just need to avoid going through floating point everywhere we're dealing with these values and we'll be fine.

erikzhang:
Wrapping `BigInteger` into `string` in JSON is okay for me.

chenzhitong (approved):

bettybao1209:
move to https://github.com/neo-project/neo-modules/pull/505

PR #498, opened 2021-02-01
RocksDbStore usable from Blockchain Toolkit

### LINKED ISSUE #497: Make RocksDb plugin usable from Neo Express / Debugger
With a few changes, neo express and debugger could use the RocksDb storage plugin directly rather than having a separate implementation.

* Must be able to create Store instance w/o using `IStorageProvider.GetStore` method. Easiest way to do that would be to make the Store class public
* Must be able to access `RocksDb.Checkpoint` method. Easiest way to do that would be to add a Checkpoint method to Store class 
* Provide mechanism to open a RocksDB instance for readonly
* Switch to [curiosity-ai fork of RocksDbSharp](https://www.nuget.org/packages/RocksDB/) which is [kept up to date](https://github.com/curiosity-ai/rocksdb-sharp) better than the original version
  * This switch is not required for use from Blockchain Toolkit, but it is the rocksdb package the Neo 3 previews of Blockchain Toolkit are using

Note, as part of this work, probably better NOT to use generic term "Store" for the IStore implementation in the RocksdDb plugin

devhawk:
fixes #497


shargon (approved):

src/RocksDBStore/Plugins/Storage/RocksDBStorageProvider.cs
```diff
@@ -0,0 +1,19 @@
+using Neo.Persistence;
+
+namespace Neo.Plugins.Storage
+{
+    public class RocksDBStorageProvider : Plugin, IStorageProvider
```
erikzhang, line 5: The plugin name should not be changed. Otherwise it will cause problems in the cli.

erikzhang (changes requested):

devhawk:
FYI, after working on migrating neo express to work against neo master, I don't think I'll be able to use the standard RocksDb plugin in express after all. Express stores additional information such as app logs in alternative RocksDB columns. I don't think we want to add that developer-tool specific functionality to this plugin

PR #503, opened 2021-02-02
Make GetExpectedBlockSize and GetExpectedBlockSystemFee public

devhawk:
needed for neo express


erikzhang:
Copy the code? I think these codes are simple.

superboyiii:
> Copy the code? I think these codes are simple.

Public does no harm to the current, right?

devhawk:
> > Copy the code? I think these codes are simple.
> 
> Public does no harm to the current, right?

I don't see how making these methods public has any negative impact on the plugin

devhawk:
> Copy the code? I think these codes are simple.

Can't. `GetExpectedBlockSize` depends on `GetExpectedBlockSizeWithoutTransactions` which depends on private member `_witnessSize`

shargon (approved):

steven1227 (approved):

PR #516, opened 2021-02-08
Add tx count check on prepare request

ZhangTao1596:
I think we should check this limit.


erikzhang:
Wait for neo-project/neo#2300

src/DBFTPlugin/ConsensusService.cs
```diff
@@ -417,6 +417,11 @@ private void OnPrepareRequestReceived(ExtensiblePayload payload, PrepareRequest
                 Log($"Timestamp incorrect: {message.Timestamp}", LogLevel.Warning);
                 return;
             }
+            if (message.TransactionHashes.Length > NativeContract.Policy.GetMaxTransactionsPerBlock(context.Snapshot))
```
vncoelho, line 420: `ProtocolSettings.Default.MaxTransactionsPerBlock` it may be now

erikzhang:
It's done in #517

PR #521, opened 2021-02-17
Exclude LedgerContract

### LINKED ISSUE #520: StatesDumper should exclude Ledger contract
As was discussed in neo-project/neo#2215 `Ledger` contract state can differ from node to node for at least two reasons:
 * different witness signatures
 * tail cutting (`MaxTraceableBlocks`)

For this reasons it's excluded from MPT and state root calculations in the `StateService`. And for the same reasons it should be excluded from `StatesDumper` output. This plugin is mostly useful for comparing state changes between different nodes and as `Ledger` state can't reliably be compared there is little value in dumping it.

Exclude `Ledger` contract state from `StatesDumper` plugin.

- Plugins

erikzhang:
Close #520

### DIFF
--- a/src/StatesDumper/Settings.cs
+++ b/src/StatesDumper/Settings.cs
@@ -1,5 +1,7 @@
 using Microsoft.Extensions.Configuration;
-using System;
+using Neo.SmartContract.Native;
+using System.Collections.Generic;
+using System.Linq;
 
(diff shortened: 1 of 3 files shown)


erikzhang:
Anyone can fix the UT?

shargon (approved):

Issue #523, opened 2021-02-18
AutoStart to OracleService plugin


gsmachado:
**Summary description**

For operations and also to turn the OracleService a bit easier to test (integration tests), it would be good to start it automatically upon CLI start.

**Do you have any solution you want to propose?**

I think that a flag `AutoStart` could be added to the `config.json` of the OracleService, exactly like we have in the `DBFTPlugin`: https://github.com/neo-project/neo-modules/blob/master/src/DBFTPlugin/DBFTPlugin/config.json

**Where in the software does this update applies to?**
- Plugins: more specifically to the `OracleService` plugin.

gsmachado:
Thinking a bit better, the same should be applied to other plugins, like `StateService`, `StateDumper`, etc etc.

shargon:
StateDumper has no Start method. But agree with `Oracle` and `StateService`

gsmachado:
@shargon check it out:

This is my `OracleService/config.json`:

```json
{
  "PluginConfiguration": {
    "Network": 5195086,
    "Nodes": [ ],
    "MaxTaskTimeout": 432000000,
    "AllowPrivateHost": true,
    "AllowedContentTypes": [ "application/json" ],
    "Https": {
      "Timeout": 15000
    },
    "NeoFS": {
      "EndPoint": "127.0.0.1:8080",
      "Timeout": 15000
    },
    "AutoStart": true
  }
}
```

Once the neo-node starts, this happens:

![image](https://user-images.githubusercontent.com/2124276/111197169-6cc40c00-85be-11eb-8619-8fa0d350cb9f.png)

If I start by the neo-node CLI, it's fine -- it starts.

If I add `"http://localhost:40332"` or `"localhost:40332"` or whatever to the `Nodes` attribute, it doesn't start either. Same behavior.

Is there anything I'm doing wrong here? Please, give me some light. ⚡  😄

superboyiii:
> @shargon check it out:
> 
> This is my `OracleService/config.json`:
> 
> ```json
> {
>   "PluginConfiguration": {
>     "Network": 5195086,
>     "Nodes": [ ],
>     "MaxTaskTimeout": 432000000,
>     "AllowPrivateHost": true,
>     "AllowedContentTypes": [ "application/json" ],
>     "Https": {
>       "Timeout": 15000
>     },
>     "NeoFS": {
>       "EndPoint": "127.0.0.1:8080",
>       "Timeout": 15000
>     },
[...]

PR #524, opened 2021-02-18
Add AutoStart

shargon:
Close https://github.com/neo-project/neo-modules/issues/523

### DIFF
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -32,6 +32,7 @@ namespace Neo.Plugins
     {
         private const int RefreshInterval = 1000 * 60 * 3;
 
+        private Wallet wallet;
         private readonly ConcurrentDictionary<ulong, OracleTask> pendingQueue = new ConcurrentDictionary<ulong, OracleTask>();
         private readonly ConcurrentDictionary<ulong, DateTime> finishedCache = new ConcurrentDictionary<ulong, DateTime>();
         private Timer timer;
@@ -97,6 +98,8 @@ namespace Neo.Plugins
                 Console.WriteLine("Please open wallet first!");
                 return;
             }
+
+            this.wallet = wallet;
 
             using (var snapshot = Blockchain.Singleton.GetSnapshot())
             {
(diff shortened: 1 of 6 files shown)


src/OracleService/OracleService.cs
```diff
@@ -60,6 +61,19 @@ protected override void Configure()
                 p.Configure();
         }
 
+        protected override void OnPluginsLoaded()
```
bettybao1209, line 64: We should change to use `OnSystemLoaded` method based on the latest neo core.

src/StateService/Settings.cs
```diff
@@ -6,13 +6,15 @@ internal class Settings
...
         public bool FullState { get; }
+        public bool AutoStart { get; }
 
         public static Settings Default { get; private set; }
 
         private Settings(IConfigurationSection section)
         {
             Path = string.Format(section.GetValue("Path", "Data_MPT_{0}"), ProtocolSettings.Default.Magic.ToString("X8"));
             FullState = section.GetValue("FullState", false);
+            AutoStart = section.GetValue("AutoStart", false);
```
ZhangTao1596, line 17: How about `AutoVerify`?
Because `StateService` has two features: 1. calculate mpt root 2. be state verifier.
Since calcaluting mpt root starts as long as StateService installed, this setting is to start verification. So `AutoStart` will be confusing. How do you think?

shargon: Done

steven1227 (approved):

src/OracleService/OracleService.cs
```diff
@@ -72,9 +86,12 @@ public override void Dispose()
         [ConsoleCommand("start oracle", Category = "Oracle", Description = "Start oracle service")]
         private void OnStart()
         {
-            if (started) return;
+            Start(walletProvider.GetWallet());
+        }
 
-            wallet = GetService<IWalletProvider>().GetWallet();
```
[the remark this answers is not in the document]
shargon: Done

erikzhang (approved):

PR #528, opened 2021-03-01
[StateService]check magic in console command

ZhangTao1596:
Currently we don't initialize state service,  if `Network` in state service is not equal `Magic`.
So that we should disable the console commands too.

### DIFF
--- a/src/StateService/StatePlugin.cs
+++ b/src/StateService/StatePlugin.cs
@@ -80,6 +80,7 @@ namespace Neo.Plugins.StateService
         [ConsoleCommand("start states", Category = "StateService", Description = "Start as a state verifier if wallet is open")]
         private void OnStartVerifyingState()
         {
+            if (System.Settings.Magic != Settings.Default.Network) return;
             Start(walletProvider.GetWallet());
         }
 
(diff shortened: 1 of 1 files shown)


shargon (dismissed):

src/StateService/StatePlugin.cs
```diff
@@ -80,6 +80,7 @@ void IPersistencePlugin.OnPersist(NeoSystem system, Block block, DataCache snaps
         [ConsoleCommand("start states", Category = "StateService", Description = "Start as a state verifier if wallet is open")]
         private void OnStartVerifyingState()
         {
+            if (System.Settings.Magic != Settings.Default.Network) throw new InvalidOperationException("Network doesn't match");
```
cloud8little, line 83:
```suggestion
            if (system.Settings.Magic != Settings.Default.Network) throw new InvalidOperationException("Network doesn't match");
```

ZhangTao1596: There is no field named `system`;
I add `System` check.

cloud8little:
Test passed.

592862cd7c2e3f2da12940b5ef6b410

shargon (approved):

cloud8little (approved):

superboyiii:
@erikzhang Could you merge it? Seems work flow is stuck.

Issue #531, opened 2021-03-01
ApplicationLogs and StateService still use the magic to generate folder name


cloud8little:
![5d50749e6e829a4304384fabf0b20af](https://user-images.githubusercontent.com/34291844/109485316-7f1d5080-7abc-11eb-8810-6112f2f4a623.png)

![9731b7f721f057ecb5a592517dee703](https://user-images.githubusercontent.com/34291844/109485329-83e20480-7abc-11eb-939e-18ae56a199c8.png)

![16c70bfce352c07369a061fb11f37cc](https://user-images.githubusercontent.com/34291844/109485332-85abc800-7abc-11eb-902e-dfdaf6427df8.png)

shargon:
I will fix it

erikzhang:
Why shouldn't magic be used?

cloud8little:
why some use magic, some does not?  @erikzhang 

![4bf774dbbff0505a80079e9c7878314](https://user-images.githubusercontent.com/34291844/109519214-fe734a00-7ae5-11eb-943b-f0df4bc5de6a.png)

erikzhang:
I think all folders should contain the network magic number.

cloud8little:
> I think all folders should contain the network magic number.

neo-cli will deal with Logs and Data_LevelDB folder without the magic number, should they also be updated? 

```cs
        public LoggerSettings(IConfigurationSection section)
        {
            this.Path = section.GetValue("Path", "Logs_{0}");
            this.ConsoleOutput = section.GetValue("ConsoleOutput", false);
            this.Active = section.GetValue("Active", false);
        }
```

```cs
        public StorageSettings(IConfigurationSection section)
        {
            this.Engine = section.GetValue("Engine", "LevelDBStore");
            this.Path = section.GetValue("Path", "Data_LevelDB_{0}");
        }
```

PR #532, opened 2021-03-01
Fix network path

shargon:
Close https://github.com/neo-project/neo-modules/issues/531

### DIFF
--- a/src/RpcNep17Tracker/RpcNep17Tracker.cs
+++ b/src/RpcNep17Tracker/RpcNep17Tracker.cs
@@ -45,7 +45,7 @@ namespace Neo.Plugins
         {
             if (_db == null)
             {
-                var dbPath = GetConfiguration().GetSection("DBPath").Value ?? "Nep17BalanceData";
+                var dbPath = string.Format(GetConfiguration().GetSection("DBPath").Value ?? "Nep17BalanceData", System.Settings.Magic.ToString("X8"));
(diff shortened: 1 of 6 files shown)


shargon:
@cloud8little clould you check it now?

erikzhang (dismissed):

cloud8little:
cbcf3bf69d0506e4cd880371fab3992

src/RpcNep17Tracker/RpcNep17Tracker.cs
```diff
@@ -38,16 +39,14 @@ protected override void OnSystemLoaded(NeoSystem system)
         {
             if (system.Settings.Magic != _network) return;
             System = system;
+            string path = string.Format(_dbPath, system.Settings.Magic.ToString("x8"));
```
cloud8little, line 42:
```suggestion
            string path = string.Format(_dbPath, system.Settings.Magic.ToString("X8"));
```

erikzhang (approved):

cloud8little:
Test passed:

1. "DBPath" in config file has "{0}" string, and network id equal to magic id. Folder (Nep17BalanceData_004F454E) generate successfully.  
2.  "DBPath" in config file has not symbol similar with "{0}" and network id equal to magic id, folder with the string in config file generate successfully.
3.  network id not same with magic id, there exist a folder name with the config file in "DBPath" field, no data generated.
if there is no folder name with the "DBPath" field, no folder generated.

cloud8little (approved):

PR #534, opened 2021-03-02
[RpcService] add server

ZhangTao1596:
Fix plugin can't register rpc methods if it is loaded after rpc server.


superboyiii:
Test **PASS**
The load sequence is different on different machine. Mine is unavailable before beacuse of this bug but it works well now.

superboyiii (approved):

shargon (approved):

PR #540, opened 2021-03-06
Fix State Autoverify when Oracle Service Unavailable

cloud8little:
The case is when designate State Service node and not designate Oracle service node, both oracle and state with autostart, then start the node. It will throw an exception with "oracle service is unavailable" and state service start failed.


superboyiii:
Test **PASS**

superboyiii (approved):

shargon (approved):

PR #553, opened 2021-03-12
[RpcNep17Tracker]Fix null exception when OnPersist before OnSystemLoaded

ZhangTao1596:
Fix node start failed with all plugins on osx.

**Reason**
 Since persist genesis block and `OnSystemLoaded` of plugins execute in different threads. `Plugin.OnPersist` might be called before `Plugin.OnSystemLoaded` on different os. If so, there will be object null exception in `RpcNep17Tracker`.

### DIFF
--- a/src/RpcNep17Tracker/RpcNep17Tracker.cs
+++ b/src/RpcNep17Tracker/RpcNep17Tracker.cs
@@ -14,6 +14,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Numerics;
+using System.Threading;
 using static System.IO.Path;
 
 namespace Neo.Plugins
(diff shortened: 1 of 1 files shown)


shargon:
We should find another way

superboyiii:
Any way, it fixed the problem. Test **Pass** on mine.

erikzhang:
I think we should revert neo-project/neo#2368

ZhangTao1596:
> I think we should revert neo-project/neo#2368

Why?

erikzhang:
> Plugin.OnPersist might be called before Plugin.OnSystemLoaded on different os.

It is impossible.

ZhangTao1596:
> > Plugin.OnPersist might be called before Plugin.OnSystemLoaded on different os.
> 
> It is impossible.

It does. We persist genesis block in `Blockchain` constructor.

erikzhang:
Please check neo-project/neo#2397

ZhangTao1596:
Close this because neo-project/neo#2397 can make all plugins work as we expect.

PR #554, opened 2021-03-12
[Oracle] Optimize neofs return type

ZhangTao1596:
Currently `filter` can't be applied to neofs results.
 
Changes:
 * Return string from all neofs reponse.
 * Convert `Object` to JSON in `GetHeaderAsync`.

### DIFF
--- a/src/OracleService/Protocols/OracleNeoFSProtocol.cs
+++ b/src/OracleService/Protocols/OracleNeoFSProtocol.cs
@@ -51,11 +51,15 @@ namespace Neo.Plugins
             }
         }
 
-        /*
-            GetAsync returns neofs object from the provided url.
-            URI scheme is "neofs:<Container-ID>/<Object-ID>/<Command>/<offset|length>".
-            If Command is not provided, full object is requested.
-        */
+
+        /// <summary>
+        /// GetAsync returns neofs object from the provided url.
+        /// If Command is not provided, full object is requested.
+        /// </summary>
+        /// <param name="uri">URI scheme is "neofs:<Container-ID>/<Object-ID>/<Command>/<offset|length>".</param>
+        /// <param name="host">Client host.</param>
+        /// <param name="cancellation">Cancellation token object.</param>
+        /// <returns>Returns neofs object.</returns>
         private Task<string> GetAsync(Uri uri, string host, CancellationToken cancellation)
         {
             string[] ps = uri.AbsolutePath.Split("/");
--- a/src/OracleService/Protocols/OracleNeoFSProtocol.cs
+++ b/src/OracleService/Protocols/OracleNeoFSProtocol.cs
@@ -84,7 +84,7 @@ namespace Neo.Plugins
         private static async Task<string> GetPayloadAsync(Client client, Address addr, CancellationToken cancellation)
         {
             Object obj = await client.GetObject(cancellation, new GetObjectParams() { Address = addr }, new CallOptions { Ttl = 2 });
-            return Convert.ToBase64String(obj.Payload.ToByteArray());
+            return Utility.StrictUTF8.GetString(obj.Payload.ToByteArray());
         }
 
         private static async Task<string> GetRangeAsync(Client client, Address addr, string[] ps, CancellationToken cancellation)
--- a/src/OracleService/Protocols/OracleNeoFSProtocol.cs
+++ b/src/OracleService/Protocols/OracleNeoFSProtocol.cs
@@ -3,6 +3,7 @@ using Neo.FileSystem.API.Client;
 using Neo.FileSystem.API.Client.ObjectParams;
 using Neo.FileSystem.API.Cryptography;
 using Neo.FileSystem.API.Refs;
+using Neo.IO.Json;
 using Neo.Network.P2P.Payloads;
 using Neo.Wallets;
 using System;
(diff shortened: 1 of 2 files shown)


shargon (approved):

src/OracleService/Protocols/OracleNeoFSProtocol.cs
```diff
@@ -76,36 +81,37 @@ private Task<byte[]> GetAsync(Uri uri, string host, CancellationToken cancellati
             };
         }
 
-        private static async Task<byte[]> GetPayloadAsync(Client client, Address addr, CancellationToken cancellation)
+        private static async Task<string> GetPayloadAsync(Client client, Address addr, CancellationToken cancellation)
         {
             Object obj = await client.GetObject(cancellation, new GetObjectParams() { Address = addr }, new CallOptions { Ttl = 2 });
-            return obj.Payload.ToByteArray();
+            return Convert.ToBase64String(obj.Payload.ToByteArray());
```
erikzhang, line 87: I think it should be `Utility.StrictUTF8.GetString(obj.Payload.ToByteArray())`. Because we support JSON only for now, right?

erikzhang: https://github.com/neo-project/neo-modules/blob/603ecec356d4ec675023501cb59c24d1e4ded73b/src/OracleService/OracleService/config.json#L7

erikzhang: And is there is way to check the content-types of the neofs object?

realloc: The only way I see is to define an optional well-known [`Attribute`](https://github.com/nspcc-dev/neofs-api/blob/master/object/types.proto#L137) for content-type and check it, if it was set.

ZhangTao1596: Done.

erikzhang:
Test?

src/OracleService/Protocols/OracleNeoFSProtocol.cs
```diff
@@ -102,7 +102,7 @@ private static async Task<string> GetRangeAsync(Client client, Address addr, str
         private static async Task<string> GetHeaderAsync(Client client, Address addr, CancellationToken cancellation)
         {
             var obj = await client.GetObjectHeader(cancellation, new ObjectHeaderParams() { Address = addr }, new CallOptions { Ttl = 2 });
-            return obj.Payload.ToString(Utility.StrictUTF8);
+            return obj.ToString();
```
erikzhang, line 105: Where is the definition of the ToString() method?

ZhangTao1596: https://github.com/neo-ngd/neofs-api-csharp/blob/055f93206e374ed6c30fa7865273b6cda9c9c548/src/Neo.FileStorage.API/object/Types.cs#L1848-L1851

It's auto-generated.

erikzhang: > return pb::JsonFormatter.ToDiagnosticString(this);

Is it deterministic? Maybe we should reads the content and use `JObject` in neo core.

ProDog:
Tested ok.

image

src/OracleService/Protocols/OracleNeoFSProtocol.cs
ZhangTao1596: Will do. BTW, `obj` is a instance of neofs `Object` class with empty `Payload`. We need transfer it to JSON.

ZhangTao1596: Done.

src/OracleService/Protocols/OracleNeoFSProtocol.cs
```diff
@@ -76,36 +86,37 @@ private Task<byte[]> GetAsync(Uri uri, string host, CancellationToken cancellati
...
+            var res = await client.GetObjectPayloadRangeData(cancellation, new RangeDataParams() { Address = addr, Range = range }, new CallOptions { Ttl = 2 });
+            return Utility.StrictUTF8.GetString(res);
         }
 
-        private static async Task<byte[]> GetHeaderAsync(Client client, Address addr, CancellationToken cancellation)
+        private static async Task<string> GetHeaderAsync(Client client, Address addr, CancellationToken cancellation)
         {
             var obj = await client.GetObjectHeader(cancellation, new ObjectHeaderParams() { Address = addr }, new CallOptions { Ttl = 2 });
-            return Utility.StrictUTF8.GetBytes(obj.ToString());
+            return JObject.Parse(obj.ToString()).ToString();
```
erikzhang, line 106: Can we read the original byte array directly? Otherwise it is converted many times.

ZhangTao1596: Do you mean `obj.ToByteArray()`? It's a byte array in protobuf encoding which we can't handle in contract.

erikzhang: There is no way to read the original bytes?

ZhangTao1596: Fixed. `ToJson` added. @ProDog Can you test again?

ProDog: Tested OK.

erikzhang (approved):

shargon (approved):

PR #562, opened 2021-03-15
[RpcServer] fix decimal

bettybao1209:


cloud8little:
Test passed.

cloud8little (approved):

erikzhang:
What's the difference?

cloud8little:
> What's the difference?

0bdc5ccb16d472d974449a95360d184

the second is right.

erikzhang:
I remember that we decided to return `BigInteger` in RPC instead of `BigDecimal`.

bettybao1209:
> I remember that we decided to return `BigInteger` in RPC instead of `BigDecimal`.

Yes, we use `BigInteger`  in both the input and output in RPC.

superboyiii:
@erikzhang Merge?

erikzhang (approved):

Issue #552, opened 2021-03-12
RpcServer Plugin: unexpected `invokecontractverify` behaviour


AnnaShaleva:
**Describe the bug**
Initially, `invokecontractverify` RPC API was created to define network fee for contract witness verification (https://github.com/neo-project/neo/issues/1977).  But it performs a `System.Contract.Call` instead of behaving like [Helper.VerifyWitness](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L190). So the resulting RPC verification fee got from `invokecontractverify` is different from the one we got during the real witness verification.

During real contract witness verification, the whole contract's script is loaded into VM, and then we jump to the verify method. Thus, to define the exact contract verification price, we should act like this (and not just perform the `System.Contract.Call` of the `verify` method).

**Expected behavior**
`invokecontractverify` behaviour should exactly match the  [`Helper.VerifyWitness` behaviour](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L190) to calculate contract witness verification fee properly.

See https://github.com/nspcc-dev/neo-go/pull/1825 for the reference implementation.

**Platform:**
 - Version: Neo 3

Ashuaidehao:
This api requires signers' private key in rpc node wallet, how to use it in client side? I wonder even after repair the behaviour, it cannot work properly in your scenarios.

PR #564, opened 2021-03-16
fix verify api

Ashuaidehao:
Close https://github.com/neo-project/neo-modules/issues/552

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -105,11 +105,11 @@ namespace Neo.Plugins
                         Invocation = u["invocation"]?.AsString(),
                         Verification = u["verification"]?.AsString()
                     })
-                    .Where(x => x.Invocation != null && x.Verification != null)
+                    .Where(x => x.Invocation != null || x.Verification != null)
                     .Select(x => new Witness()
                     {
-                        InvocationScript = Convert.FromBase64String(x.Invocation),
-                        VerificationScript = Convert.FromBase64String(x.Verification)
+                        InvocationScript = Convert.FromBase64String(x.Invocation ?? string.Empty),
+                        VerificationScript = Convert.FromBase64String(x.Verification ?? string.Empty)
                     }).ToArray()
             };
 
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -97,7 +97,20 @@ namespace Neo.Plugins
                 Scopes = (WitnessScope)Enum.Parse(typeof(WitnessScope), u["scopes"]?.AsString()),
                 AllowedContracts = ((JArray)u["allowedcontracts"])?.Select(p => UInt160.Parse(p.AsString())).ToArray(),
                 AllowedGroups = ((JArray)u["allowedgroups"])?.Select(p => ECPoint.Parse(p.AsString(), ECCurve.Secp256r1)).ToArray()
-            }).ToArray());
+            }).ToArray())
+            {
+                Witnesses = _params.Select(u => new
+                    {
+                        Invocation = u["invocation"]?.AsString(),
+                        Verification = u["verification"]?.AsString()
+                    })
+                    .Where(x => x.Invocation != null && x.Verification != null)
+                    .Select(x => new Witness()
+                    {
+                        InvocationScript = Convert.FromBase64String(x.Invocation),
+                        VerificationScript = Convert.FromBase64String(x.Verification)
+                    }).ToArray()
+            };
 
             // Validate format
 
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -99,7 +99,8 @@ namespace Neo.Plugins
                 AllowedGroups = ((JArray)u["allowedgroups"])?.Select(p => ECPoint.Parse(p.AsString(), ECCurve.Secp256r1)).ToArray()
             }).ToArray())
             {
-                Witnesses = _params.Select(u => new
+                Witnesses = _params
+                    .Select(u => new
                     {
                         Invocation = u["invocation"]?.AsString(),
                         Verification = u["verification"]?.AsString()
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -368,6 +368,7 @@ namespace Neo.Plugins
                 for (int i = args.Length - 1; i >= 0; i--)
                     sb.EmitPush(args[i]);
 
+                tx.Witnesses ??= new Witness[] {new() {InvocationScript = sb.ToArray()}};
                 engine.LoadScript(new Script(sb.ToArray()), configureState: p => p.CallFlags = CallFlags.None);
             }
             JObject json = new JObject();
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -362,17 +362,21 @@ namespace Neo.Plugins
             };
             using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: system.Settings);
             engine.LoadContract(contract, md, CallFlags.ReadOnly);
+
+            var invocationScript = new byte[] { };
             if (args.Length > 0)
             {
                 using ScriptBuilder sb = new ScriptBuilder();
                 for (int i = args.Length - 1; i >= 0; i--)
                     sb.EmitPush(args[i]);
 
-                tx.Witnesses ??= new Witness[] {new() {InvocationScript = sb.ToArray()}};
-                engine.LoadScript(new Script(sb.ToArray()), configureState: p => p.CallFlags = CallFlags.None);
+                invocationScript = sb.ToArray();
+                tx.Witnesses ??= new Witness[] { new() { InvocationScript = invocationScript } };
+                engine.LoadScript(new Script(invocationScript), configureState: p => p.CallFlags = CallFlags.None);
             }
             JObject json = new JObject();
-            json["script"] = Convert.ToBase64String(contract.Script);
+
+            json["script"] = Convert.ToBase64String(invocationScript);
             json["state"] = engine.Execute();
             json["gasconsumed"] = engine.GasConsumed.ToString();
             json["exception"] = GetExceptionMessage(engine.FaultException);
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -333,7 +333,6 @@ namespace Neo.Plugins
         [RpcMethod]
         protected virtual JObject InvokeContractVerify(JArray _params)
         {
-            CheckWallet();
             UInt160 script_hash = UInt160.Parse(_params[0].AsString());
             ContractParameter[] args = _params.Count >= 2 ? ((JArray)_params[1]).Select(p => ContractParameter.FromJson(p)).ToArray() : new ContractParameter[0];
             Signers signers = _params.Count >= 3 ? SignersFromJson((JArray)_params[2], system.Settings) : null;
@@ -348,20 +347,20 @@ namespace Neo.Plugins
             {
                 throw new RpcException(-100, "Unknown contract");
             }
-            var methodName = "verify";
+            var md = contract.Manifest.Abi.GetMethod("verify", 0);
+            if (md is null)
+                throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method without arguments");
+            if (md.ReturnType != ContractParameterType.Boolean)
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
             Transaction tx = signers == null ? null : new Transaction
             {
                 Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                Attributes = Array.Empty<TransactionAttribute>(),
+                Witnesses = signers.Witnesses,
             };
-            ContractParametersContext context = new ContractParametersContext(snapshot, tx);
-            wallet.Sign(context);
-            tx.Witnesses = context.Completed ? context.GetWitnesses() : null;
-
             using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: system.Settings);
-            engine.LoadScript(new ScriptBuilder().EmitDynamicCall(scriptHash, methodName, args).ToArray(), rvcount: 1);
-
+            engine.LoadContract(contract, md, CallFlags.ReadOnly);
             JObject json = new JObject();
             json["script"] = Convert.ToBase64String(contract.Script);
             json["state"] = engine.Execute();
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -353,14 +353,9 @@ namespace Neo.Plugins
             if (md.ReturnType != ContractParameterType.Boolean)
                 throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
-            var defaultSigner = new List<Signer>() { new() { Account = scriptHash } };
-            if (signers != null)
-            {
-                defaultSigner.AddRange(signers.GetSigners());
-            }
             Transaction tx = new Transaction
             {
-                Signers = defaultSigner.ToArray(),
+                Signers = signers == null ? new Signer[] { new() { Account = scriptHash } } : signers.GetSigners(),
                 Attributes = Array.Empty<TransactionAttribute>(),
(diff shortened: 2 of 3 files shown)


src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,20 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+                throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method without arguments");
+            if (md.ReturnType != ContractParameterType.Boolean)
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
             Transaction tx = signers == null ? null : new Transaction
             {
                 Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                Attributes = Array.Empty<TransactionAttribute>(),
+                Witnesses = signers.Witnesses,
```
shargon, line 360: `Witnesses` it wil lbe null, isn't it?

Ashuaidehao: Yes, just like https://github.com/neo-project/neo-modules/blob/93fb62019bec6606c406e337706d5c5e235f948c/src/RpcServer/RpcServer.SmartContract.cs#L69

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,20 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
             {
                 throw new RpcException(-100, "Unknown contract");
             }
-            var methodName = "verify";
+            var md = contract.Manifest.Abi.GetMethod("verify", 0);
```
AnnaShaleva, line 350: Can we set `pcount` to -1 here in order to be able to call `verify` with arguments? Just like in https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L222.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,20 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+                Witnesses = signers.Witnesses,
             };
-            ContractParametersContext context = new ContractParametersContext(snapshot, tx);
-            wallet.Sign(context);
-            tx.Witnesses = context.Completed ? context.GetWitnesses() : null;
-
             using ApplicationEngine engine = ApplicationEngine.Create(TriggerType.Verification, tx, snapshot.CreateSnapshot(), settings: system.Settings);
-            engine.LoadScript(new ScriptBuilder().EmitDynamicCall(scriptHash, methodName, args).ToArray(), rvcount: 1);
-
+            engine.LoadContract(contract, md, CallFlags.ReadOnly);
```
AnnaShaleva, line 363: And create an invocation script from the given args (it will push all arguments on the stack). And load the script right after the contract script, just like in https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L246.

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,20 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
                 throw new RpcException(-100, "Unknown contract");
             }
-            var methodName = "verify";
+            var md = contract.Manifest.Abi.GetMethod("verify", 0);
+            if (md is null)
+                throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method without arguments");
+            if (md.ReturnType != ContractParameterType.Boolean)
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
             Transaction tx = signers == null ? null : new Transaction
```
AnnaShaleva, line 356: Can we also fill the transaction when `signers == null`? We can do it because we know exactly at least one signer (it's the contract to be called). And we also know the witness for this signer: invocation script pushes `args` on the stack; verification script is empty because it's a contract witness.

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Removed it since no use here.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Updated it.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Updated it. Please test it like:
```
{
    "jsonrpc": "2.0",
    "method": "invokecontractverify",
    "params": [
        "0x31f714e04766139c8f3705f5ce94518e1411ec07",
        [
            {
                "type": "ByteArray",
                "value": "DED2HMp00JZ2zcJYjKgTp4Tl0GODm5r14PKN6ip8i2+8ox3JYqN2xNHxQzCgzENqSD52GhVF4eGx2i405zxM1KTH"
            }
        ],
        [
            {
                "account": "NMA2FKN8up2cEwaJgtmAiDrZWB69ApnDfp",
                "scopes": "CalledByEntry"
            }
        ]
    ],
    "id": 1
}
```

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,23 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
             {
                 throw new RpcException(-100, "Unknown contract");
             }
-            var methodName = "verify";
+            var md = contract.Manifest.Abi.GetMethod("verify", -1);
+            if (md is null)
+                throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method without arguments");
```
AnnaShaleva, line 352: Exception message should also be updated:
```
throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method");
```

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,23 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+            if (md.ReturnType != ContractParameterType.Boolean)
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
             Transaction tx = signers == null ? null : new Transaction
             {
                 Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                Attributes = Array.Empty<TransactionAttribute>(),
```
AnnaShaleva, line 359: Can we also fill the transaction script? I think that []byte{OpCode.RET} is enough. 

This is needed for those `verify` methods that use `System.Runtime.GetScriptContainer` interop, because transaction can't be serialized without a script.

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
AnnaShaleva: Can we pass witnesses via RPC request parameters, just like signers are passed? And then fill `Witnesses` here.

src/RpcServer/RpcServer.Wallet.cs
AnnaShaleva: It's not very convenient to construct an invocation script on the client. I would prefer to pass simple standard parameters to the server as we do in the `InvokeFunction` RPC method. Like the following:
```
{
    "jsonrpc": "2.0",
    "method": "invokecontractverify",
    "params": [
        "0x31f714e04766139c8f3705f5ce94518e1411ec07",
        [
            {
                "type": "Boolean",
                "value": "true"
            },
            {
                "type": "Integer",
                "value": 123
            }
        ],
        [
            {
                "account": "NMA2FKN8up2cEwaJgtmAiDrZWB69ApnDfp",
                "scopes": "CalledByEntry"
            }
        ]
    ],
    "id": 1
}
```

And then we can construct an invocation script on the server by performing the following:
```
using (ScriptBuilder sb = new ScriptBuilder())
{
        for (int i = args.Count - 1; i >= 0; i--)
                sb.EmitPush(args[i]);
}
```

shargon: Agree with @AnnaShaleva

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Ok.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Updated.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Updated.

superboyiii:
@AnnaShaleva Need your review again.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,33 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+                throw new RpcException(-101, $"The smart contract {contract.Hash} haven't got verify method.");
+            if (md.ReturnType != ContractParameterType.Boolean)
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
+
+            var defaultSigner = new List<Signer>() { new() { Account = scriptHash } };
+            if (signers != null)
             {
-                Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                defaultSigner.AddRange(signers.GetSigners());
```
AnnaShaleva, line 359: Can we make here `defaultSigner = signers.GetSigners();`?
It shouldn't be mandatory for the contract signer to be a sender. And if the user wants to provide signers by himself, then he should include the contract signer with the desired scope in the parameters list.

The other option is to check that `signers` does not contain the contract signer and if the contract signer is missing, we should add it to the `signers` list.

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: Updated it.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,28 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
-            Transaction tx = signers == null ? null : new Transaction
+            Transaction tx = new Transaction
             {
-                Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                Signers = signers == null ? new Signer[] { new() { Account = scriptHash } } : signers.GetSigners(),
+                Attributes = Array.Empty<TransactionAttribute>(),
+                Script = new[] { (byte)OpCode.RET }
```
AnnaShaleva, line 360: And the last issue: at the current moment, the transaction is completely missing witnesses. I would suggest to pass witnesses along with signers via RPC request parameters using the following structure:
```
{
    "jsonrpc": "2.0",
    "method": "invokecontractverify",
    "params": [
        "0x31f714e04766139c8f3705f5ce94518e1411ec07",
        [ ],
        [
            {
                "account": "NiRqSd5MtRZT5yUhgWd7oG11brkDG76Jim",
                "scopes": "CalledByEntry",
                "invocation": "DEDr2gA/8T/wxQvgOZVfCdkbj6uGrprkDgJvpOJCcbl+tvlKZkZytCZEWm6NoZhJyIlEI3VQSLtU3AHuJfShAT5L",
                "verification": "DCEDAS1H52IQrsc745qz0YbgpA/o2Gv6PU+r/aV7oTuI+WoLQZVEDXg=",
            },
            {
                "account": "NMA2FKN8up2cEwaJgtmAiDrZWB69ApnDfp",
                "scopes": "CalledByEntry",
            }
        ]
    ],
    "id": 1
}
```


And then parse the provided witnesses in `GetSignersFromJson method (https://github.com/neo-project/neo-modules/blob/master/src/RpcServer/RpcServer.SmartContract.cs#L94).

And then fill the transaction like the following:
```
            Transaction tx = new Transaction
            {
                Witnesses = signers.Witnesses,
                ...
            }
```

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
AnnaShaleva: It will help us to get rid of the private key dependency (because if we parse witnesses, then we don't need private keys in the RPC node wallet anymore).

Ashuaidehao: Ok

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,29 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+                throw new RpcException(-102, "The verify method doesn't return boolean value.");
 
-            Transaction tx = signers == null ? null : new Transaction
+            Transaction tx = new Transaction
             {
-                Signers = signers.GetSigners(),
-                Attributes = Array.Empty<TransactionAttribute>()
+                Signers = signers == null ? new Signer[] { new() { Account = scriptHash } } : signers.GetSigners(),
+                Attributes = Array.Empty<TransactionAttribute>(),
+                Witnesses = signers?.Witnesses,
```
AnnaShaleva, line 360:
```suggestion
                Witnesses = signers == null ? new Witness() { InvocationScript = sb.ToArray() } : signers.Witnesses,
```

where `sb` is from
```
if (args.Length > 0)
            {
                using ScriptBuilder sb = new ScriptBuilder();
                for (int i = args.Length - 1; i >= 0; i--)
                    sb.EmitPush(args[i]);
             }
```

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -97,7 +97,21 @@ private static Signers SignersFromJson(JArray _params, ProtocolSettings settings
...
+                    {
+                        Invocation = u["invocation"]?.AsString(),
+                        Verification = u["verification"]?.AsString()
+                    })
+                    .Where(x => x.Invocation != null && x.Verification != null)
+                    .Select(x => new Witness()
+                    {
+                        InvocationScript = Convert.FromBase64String(x.Invocation),
+                        VerificationScript = Convert.FromBase64String(x.Verification)
+                    }).ToArray()
```
AnnaShaleva, line 113:
```suggestion
                    .Where(x => x.Invocation != null || x.Verification != null)
                    .Select(x => new Witness()
                    {
                        InvocationScript = x.Invocation != null ?? Convert.FromBase64String(x.Invocation) : null,
                        VerificationScript = x.Verification != null ?? Convert.FromBase64String(x.Verification) : null
                    }).ToArray()
```

AnnaShaleva (changes requested):
Almost there.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: `sb` will be loaded into engine directly, is this `InvocationScript ` necessary?

AnnaShaleva: We know exactly the invocation script, why not to fill it in? By doing this we're getting closer to the real transaction witnesses.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -348,20 +347,29 @@ private JObject GetVerificationResult(UInt160 scriptHash, ContractParameter[] ar
...
+            if (args.Length > 0)
+            {
+                using ScriptBuilder sb = new ScriptBuilder();
+                for (int i = args.Length - 1; i >= 0; i--)
+                    sb.EmitPush(args[i]);
 
+                engine.LoadScript(new Script(sb.ToArray()), configureState: p => p.CallFlags = CallFlags.None);
+            }
             JObject json = new JObject();
             json["script"] = Convert.ToBase64String(contract.Script);
```
AnnaShaleva, line 374: And let the resulting `script` be the InvocationScript (which is `sb.ToArray()`). Because 1) the user is aware of the contract being invoked and 2) the user should have the ability to check that all the parameters were converted as expected.

AnnaShaleva (changes requested):

src/RpcServer/RpcServer.Wallet.cs
AnnaShaleva: See https://github.com/nspcc-dev/neo-go/pull/1825#discussion_r592256453.

Ashuaidehao: What `script` mean if `sb` is empty?

AnnaShaleva: That no parameters were pushed on the stack before `verify` execution.

AnnaShaleva (approved):
Nice changes, it solves the problem.

superboyiii (approved):

PR #517, opened 2021-02-08
Sync neo

bettybao1209:
Close https://github.com/neo-project/neo-modules/issues/519

~~1. Needs to add a new message type `GetInstance` in `LocalNode` class;~~
~~2. `UpdateExtensibleWitnessWhiteList` in `Blockchain` class and `Verify` in `ExtensiblePayload` class should be `public`, since they are needed in DBFT plugin.~~
~~3. https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/LedgerContract.cs#L72 will throw NullReferenceException, which is caused by the non-existence of the  `CreateStorageKey(Prefix_CurrentBlock)` (invoked before PostPersist).~~

PR #568, opened 2021-03-19
StatesDumper: fix bug with missing states dumps

AnnaShaleva:
### Problem

StatesDumper does not save states dumps (because of https://github.com/neo-project/neo-modules/pull/517 changes).

### Solution

We need to put states to `bs_cache` even if it's empty.


superboyiii:
@shargon Need your review.

shargon (approved):

Issue #573, opened 2021-03-26
invokefunction with open wallet has internal hardcoded GAS limit


snowypowers:
**Describe the bug**
I have been trying to automate a simple setup for my private chain docker. One of the newer features I wish to add to my docker image is to have the chain have candidates registered so I can start voting. 

I discovered that no matter what settings I set, I seem to be unable to get `invokefunction` RPC call to return me a serialized transaction.

**To Reproduce**
1. `docker pull snowypowers/neo-privatenet:v3.0.0-rc1` (assuming you are connected to docker.io) [source](https://github.com/snowypowers/neo-privatenet-docker/tree/v3-solonet/v3/solonet/rc1)
2. Run the docker image (`docker run --rm -d  -p 20332:20332/tcp -p 20333:20333/tcp -p 20334:20334/tcp snowypowers/neo-privatenet:v3.0.0-rc1`)
3. Open the wallet with curl: `curl -d '{"jsonrpc": "2.0","method": "openwallet", "params": ["wallet.json","wallet"],"id": 3}' -H 'Content-Type: application/json' http://localhost:20332`
4. Attempt to `invokefunction` to register a candidate: 
```
curl -d '
{
    "jsonrpc": "2.0",
    "method": "invokefunction",
    "params": [
        "ef4073a0f2b305a38ec4050e4d3d28bc40ea63f5",
        "registerCandidate",
        [{
            "type": "ByteArray",
            "value": "AgKKmYJu3AyX0Y4itpMjc9kI0yOqf5JlanfsJuiGFpnv"
        }],
        [
            {
                "account": "b611bfa60886125e0071b56deeaf6efb40bd2336",
                "scopes": "Global"
            }
        ]
    ],
    "id": 3
}' -H 'Content-Type: application/json' http://localhost:20332
```
5. See that the response is:
```
{
    "jsonrpc": "2.0",
    "id": 3,
    "result": {
        "script": "DCECAoqZgm7cDJfRjiK2kyNz2QjTI6p/kmVqd+wm6IYWme8RwB8MEXJlZ2lzdGVyQ2FuZGlkYXRlDBT1Y+pAvCg9TQ4FxI6jBbPyoHNA70FifVtS",
        "state": "HALT",
        "gasconsumed": "100001045290",
        "exception": "Insufficient GAS.",
        "stack": [
            {
                "type": "Boolean",
                "value": true
            }
        ]
    }
}
```

Some digging through the code:

[...]

PR #574, opened 2021-03-27
Fix invokefunction

shargon:
Close https://github.com/neo-project/neo-modules/issues/573

### DIFF
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -178,14 +178,14 @@ namespace Neo.Plugins
             Transaction tx;
             try
             {
-                tx = wallet.MakeTransaction(system.StoreView, Convert.FromBase64String(result["script"].AsString()), sender, witnessSigners);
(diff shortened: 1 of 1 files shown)


Qiao-Jin:
Default `MaxGasInvoke` is 20. For the specified problem in #573, same error would still occur even with this PR patched. Maybe should also add a log reminding user to increase `MaxGasInvoke` in such cases?

https://github.com/neo-project/neo-modules/blob/e77f07401d81724b088b65fd6c63b428c6625837/src/RpcServer/RpcServer/config.json#L13

superboyiii:
I think it's OK, we can wait for @snowypowers 's response.

shargon:
> Maybe should also add a log reminding user to increase MaxGasInvoke in such cases?

Any error will be catched in line 

https://github.com/neo-project/neo-modules/pull/574/files#diff-f4edf66b165030d591ad126a9d9a91daeed2f74df8c6aff0150a46fa193e5d4eR185

snowypowers:
Yes this should work out for my scenario. I forgot to mention I did set my config's MaxGasInvoke to 10000 so this fix will help me out!

superboyiii (approved):

Qiao-Jin (approved):

PR #575, opened 2021-03-29
[StateService] Add sender check

### LINKED ISSUE #572: Stateroots are being broadcasted multiple times
**Describe the bug**
Stateroot messages are spread through the network as extensible payloads and these payloads are signed by each state validator individually using different senders. Even though all state validators end up with the same stateroot contents (witnesses may differ, but that's the same stateroot signed anyway) the resulting extensible payloads are different. Each of state validators then sends this payload of its own to the network and we end up with N (number of state validators) independent messages bearing the same contents instead of one. This creates additional useless P2P traffic and uses more of `RelayCache` then is needed.

**Expected behavior**
One stateroot message (corresponding to height H) broadcasted with one payload.

ZhangTao1596:
Close #572

### DIFF
--- a/src/StateService/Verification/VerificationContext.cs
+++ b/src/StateService/Verification/VerificationContext.cs
@@ -33,6 +33,15 @@ namespace Neo.Plugins.StateService.Verification
         public int MyIndex => myIndex;
         public uint RootIndex => rootIndex;
         public ECPoint[] Verifiers => verifiers;
+        public int Sender
+        {
+            get
+            {
+                int p = ((int)rootIndex - Retries) % verifiers.Length;
+                return p >= 0 ? p : p + verifiers.Length;
+            }
+        }
+        public bool IsSender => myIndex == Sender;
         public ICancelable Timer;
         public StateRoot StateRoot
         {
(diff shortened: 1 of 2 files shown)


src/StateService/Verification/VerificationService.cs
```diff
@@ -45,7 +45,7 @@ private void OnStateRootVote(Vote vote)
 
         private void CheckVotes(VerificationContext context)
         {
-            if (context.CheckSignatures())
+            if (context.CheckSignatures() && context.IsSender)
```
shargon, line 48:
```suggestion
            if (context.IsSender && context.CheckSignatures())
```

ZhangTao1596: Done.

src/StateService/Verification/VerificationContext.cs
```diff
@@ -111,17 +120,20 @@ public bool CheckSignatures()
...
+            if (StateRoot.Witness is null)
             {
-                if (!signatures.TryGetValue(i, out byte[] sig)) continue;
-                sc.AddSignature(contract, verifiers[i], sig);
-                j++;
+                Contract contract = Contract.CreateMultiSigContract(M, verifiers);
+                ContractParametersContext sc = new ContractParametersContext(StatePlugin.System.StoreView, StateRoot);
+                for (int i = 0, j = 0; i < verifiers.Length && j < M; i++)
+                {
+                    if (!signatures.TryGetValue(i, out byte[] sig)) continue;
```
shargon, line 129:
```suggestion
                    if (!signatures.TryGetValue(i, out byte[] sig)) return false;
```

ZhangTao1596: Why? It is uncertain from which verifiers the signature was received.

shargon (approved):

PR #577, opened 2021-03-31
InvokeIterator

chenzhitong:


src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -144,6 +144,65 @@ protected virtual JObject InvokeScript(JArray _params)
...
+                    var iterator = result.GetInterface<IIterator>();
+                    for (int i = 0; iterator.Next(); i++)
+                    {
+                        if (i >= startIndex && i < endIndex)
+                            resultList.Add(iterator.Value());
+                    }
+                    json["result"] = new JArray(resultList.Select(p => ((VM.Types.StackItem)p).ToJson()));
+                }
+                json["startindex"] = startIndex;
+                json["endindex"] = endIndex;
```
shargon, line 192: It will be the same as the request, we can remove them

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -144,6 +144,65 @@ protected virtual JObject InvokeScript(JArray _params)
...
+                if (result is null)
+                {
+                    json["result"] = null;
+                }
+                else
+                {
+                    var iterator = result.GetInterface<IIterator>();
+                    for (int i = 0; iterator.Next(); i++)
+                    {
+                        if (i >= startIndex && i < endIndex)
```
shargon, line 186: null check?

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -144,6 +144,65 @@ protected virtual JObject InvokeScript(JArray _params)
             return GetInvokeResult(script, signers);
         }
 
+        [RpcMethod]
+        protected virtual JObject InvokeIterator(JArray _params)
```
devhawk, line 148: Can someone please provide the reasoning behind adding this new Rpc method? What's the use case for it?

superboyiii: > Can someone please provide the reasoning behind adding this new Rpc method? What's the use case for it?

You could find `Iterator` in (Nep11 NFT Token)[https://github.com/neo-project/proposals/blob/nft/nep-11.mediawiki] and it's really important so that we need this RPC.

src/RpcServer/RpcServer.SmartContract.cs
superboyiii: > It will be the same as the request, we can remove them

It's in use for hyper result. (for prevent malicious attack)

src/RpcServer/RpcServer.SmartContract.cs
shargon: Iterate if the result it's an interator automatically?

nicolegys:
Test passed.

src/RpcServer/RpcServer.SmartContract.cs
chenzhitong: > Iterate if the result it's an interator automatically?

But it takes two additional parameters, startIndex and endIndex,

superboyiii: > Iterate if the result it's an interator automatically?

Then why we use list? It's using iterator for a senario that if huge amount of data, we could get it one by one so that not cause any response issue.

superboyiii (approved):

superboyiii:
Need your review again. @shargon @erikzhang

shargon:
The code looks good, but I think that it's better to return the output without call an extra method. Please take a look to https://github.com/neo-project/neo-modules/pull/583

PR #579, opened 2021-04-07
Use only one flag to indicate Oracle started status

joeqian10:
I am wondering why we need both `started` and `stopped` flags. Isn't `stopped` equal to NOT `started`?

### DIFF
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -38,7 +38,6 @@ namespace Neo.Plugins
         private Timer timer;
         private readonly CancellationTokenSource cancelSource = new CancellationTokenSource();
         private bool started = false;
-        private bool stopped = false;
         private IWalletProvider walletProvider;
         private int counter;
         private NeoSystem System;
@@ -84,7 +83,7 @@ namespace Neo.Plugins
         public override void Dispose()
         {
             OnStop();
-            while (!stopped)
+            while (started)
                 Thread.Sleep(100);
             foreach (var p in protocols)
                 p.Value.Dispose();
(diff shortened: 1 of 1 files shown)


shargon (approved):

shargon:
@superboyiii test?

erikzhang:
In fact there are three states. `!started && !stopped` is the state of not started.

joeqian10:
> In fact there are three states. `!started && !stopped` is the state of not started.

`!started && !stopped` means the Oracle has never been started. `started && !stopped` means the Oracle has been started but has not been stopped yet. `started && stopped` means the Oracle has been started and then been stopped. The above three are all correct states. Still I don't think they are necessary.
But `!started && stopped` means the Oracle has never been started but been stopped which could be possibly achieved by using the "stop oracle" command in neo-cli @superboyiii could you make a test on this?

superboyiii:
Test *PASS*
It works well on mine.
image

superboyiii (approved):

superboyiii:
> In fact there are three states. `!started && !stopped` is the state of not started.

It's a `true` or `false` issue, no need to make three states. Just pay attention to current state is started or not.

erikzhang:
You can't start it again if it is stopped. How do you do this if it's `true` or `false`?

superboyiii:
> You can't start it again if it is stopped. How do you do this if it's `true` or `false`?

Oracle Service is not recommand to stop, maybe we can  remove stop command totally. There's no senario to stop Oracle, it's almost equal stop node to some extent.

shargon:
I think that all services should be able to stop

src/OracleService/OracleService.cs
```diff
@@ -128,19 +127,24 @@ public void Start(Wallet wallet)
         [ConsoleCommand("stop oracle", Category = "Oracle", Description = "Stop oracle service")]
         private void OnStop()
         {
+            if (!started)
+            {
+                Console.WriteLine("Oracle service is not started");
+                return;
+            }
```
shargon, line 134: We can remove this lines
```suggestion
```

superboyiii:
> I think that all services should be able to stop

It's the same to input `exit` for an Oracle server.

PR #583, opened 2021-04-15
Process iterator in RPC result

shargon:
Alternative to https://github.com/neo-project/neo-modules/pull/577

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -7,8 +7,10 @@ using Neo.IO.Json;
 using Neo.Network.P2P.Payloads;
 using Neo.Persistence;
 using Neo.SmartContract;
+using Neo.SmartContract.Iterators;
 using Neo.SmartContract.Native;
 using Neo.VM;
+using Neo.VM.Types;
 using Neo.Wallets;
 using System;
 using System.IO;
@@ -76,17 +78,43 @@ namespace Neo.Plugins
             json["exception"] = GetExceptionMessage(engine.FaultException);
             try
             {
-                json["stack"] = new JArray(engine.ResultStack.Select(p => p.ToJson()));
+                int max = settings.MaxResultItems;
+                json["stack"] = new JArray(engine.ResultStack.Select(p => ToJson(p, ref max)));
             }
             catch (InvalidOperationException)
             {
-                json["stack"] = "error: recursive reference";
+                json["stack"] = "error: invalid operation";
             }
             if (engine.State != VMState.FAULT)
             {
                 ProcessInvokeWithWallet(json, signers);
             }
             return json;
+        }
+
+        private JObject ToJson(StackItem p, ref int max)
+        {
+            if (max <= 0) throw new InvalidOperationException();
+            max--;
+
+            if (p is InteropInterface result)
+            {
+                var iterator = result.GetInterface<IIterator>();
+                if (iterator != null)
+                {
+                    VM.Types.Array resultList = new();
+                    for (int i = 0; iterator.Next(); i++)
+                    {
+                        resultList.Add(iterator.Value());
+                        max--;
+                        if (max <= 0) throw new InvalidOperationException();
+                    }
+
+                    return ToJson(resultList, ref max);
+                }
+            }
+
+            return p.ToJson();
         }
 
         private static Signers SignersFromJson(JArray _params, ProtocolSettings settings)
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -103,7 +103,7 @@ namespace Neo.Plugins
                 if (iterator != null)
                 {
                     VM.Types.Array resultList = new();
-                    for (int i = 0; iterator.Next(); i++)
+                    while (iterator.Next())
                     {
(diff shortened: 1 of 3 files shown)


erikzhang:
If you return the values of the iterator directly, the developer will not be able to distinguish whether the method returns an array or an iterator.

shargon:
> If you return the values of the iterator directly, the developer will not be able to distinguish whether the method returns an array or an iterator.

I think that api users doesn't care about the low level type, they want the value. Otherwise you will need to know first that it's an iterator result in order to make a different call.

erikzhang:
At least we should not modify its type.

```
{
  "type": "InteropInterface",
  "value": []
}
```

satoshichou:
I tried modules, making everything to config is not a good way, and server can't modify config when it's running a node. I find more and more paramaters are added to config.json, it's too complex for reading and understanding for a new developer. I like the style of https://github.com/neo-project/neo-modules/pull/577 much more.

superboyiii:
As @satoshichou said, I totally agree. Don't make config too heavy and user could choose how much times iterator could run.

steven1227:
I agree with @SueNEO  and @satoshichou  and I prefer that [Pull Request](https://github.com/neo-project/neo-modules/pull/577)

erikzhang:
I support @shargon's solution. Because if you use index, it may cause data inconsistency. For example, if I call `invokeiterator` twice, the first request is 1~100, and the second request is 101~200. But if a new database insertion or deletion occurs between the two requests, the data returned between these two requests is inconsistent.

nicolegys:
I'm confused about `MaxResultItems`. My test shows that when I set the value 10, if the amount of return result is more than 10, I will get this returned `"stack": "error: invalid operation"`. I think it should return the first 10 result and give a message, to tell the user that there are still some data not showed.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -89,6 +92,34 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
...
+                        max--;
+                        if (max <= 0) throw new InvalidOperationException();
+                    }
+
+                    JObject ret = new();
+                    ret["type"] = "InteropInterface";
+                    ret["value"] = ToJson(resultList, ref max);
+                    return ret;
+                }
+            }
```
satoshichou, line 118: At least should show some warning and return current `InteropInterface` value.
```suggestion
string warning = null;
            if (max <= 0) throw new InvalidOperationException();
            max--;

            if (p is InteropInterface result)
            {
                var iterator = result.GetInterface<IIterator>();
                if (iterator != null)
                {
                    VM.Types.Array resultList = new();
                    while (iterator.Next())
                    {
                        resultList.Add(iterator.Value());
                        max--;
                        if (max <= 0)
                        {
                            max++;
                            warning = "MaxIteratorItems is acheived. Can't show all values.";
                            break;
                        }
                    }

                    JObject ret = new();
                    ret["type"] = "InteropInterface";
                    ret["value"] = ToJson(resultList, ref max);
                    ret["warning"] = warning;
                    return ret;
                }
            }
```

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -89,6 +92,41 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
...
+                {
+                    VM.Types.Array resultList = new();
+                    while (iterator.Next())
+                    {
+                        resultList.Add(iterator.Value());
+                        max--;
+                        if (max <= 0)
+                        {
+                            max++;
+                            warning = "MaxIteratorItems is acheived. Can't show all values.";
```
shargon, line 114:
```suggestion
                            warning = "MaxIteratorItems is achieved. The values was truncated.";
```

nicolegys:
There's a boundary case. If the value of `MaxResultItems` is just enough to iterate over the data, the return result still shows the warning.

erikzhang (dismissed):

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -78,13 +78,7 @@ private JObject GetInvokeResult(byte[] script, Signers signers = null)
...
             try
             {
-                int max = settings.MaxResultItems;
-                string warning = null;
-                json["stack"] = new JArray(engine.ResultStack.Select(p => ToJson(p, ref max, ref warning)).Where(u => u is not null));
-                if (!string.IsNullOrEmpty(warning))
-                {
-                    json["warning"] = warning;
-                }
+                json["stack"] = new JArray(engine.ResultStack.Select(p => ToJson(p, settings.MaxResultItems)));
```
shargon, line 81: Now is not a `MaxResultItems` it's a `MaxIteratorResultItems`

erikzhang: Yes. We should rename it.

nicolegys (dismissed):
Test pass, just need to rename the parameter.

nicolegys:
Since this solution not supports index, when the amount of data is very large, does user need to run his own node to get the complete data?

erikzhang:
> Since this solution not supports index, when the amount of data is very large, does user need to run his own node to get the complete data?

Yes. I think public nodes are not suitable for running high-load tasks.

erikzhang (approved):

nicolegys (approved):

PR #590, opened 2021-05-18
Add VRF nonce to the consensus.

Liaojinghui:
This pull request adds VRF support to the dBFT consensus.
Proof of the VRF only exists in the PrepareRequest message.

### DIFF
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -564,9 +564,14 @@ namespace Neo.Consensus
         private Transaction CreateNonceTransaction(uint blockIndex, ECPoint pubkey, ProtocolSettings settings)
         {
             var (_, nonce) = GetNonce(keyPair.PrivateKey);
-            using ScriptBuilder sb = new();
-            sb.Emit(OpCode.RET);
-            var tx = wallet.MakeTransaction(Snapshot, sb.ToArray(), null, new[]
+            byte[] script;
+            using (ScriptBuilder scriptBuilder = new())
+            {
+                scriptBuilder.Emit(OpCode.RET);
+                script = scriptBuilder.ToArray();
+            };
+
+            var tx = wallet.MakeTransaction(Snapshot, script, null, new[]
                 {
                     new Signer
                     {
@@ -574,7 +579,7 @@ namespace Neo.Consensus
                         Scopes = WitnessScope.None
                     }
                 }, Array.Empty<TransactionAttribute>());
-
+           
             ContractParametersContext sc;
             try
             {
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -338,21 +338,14 @@ namespace Neo.Consensus
 
         private Tuple<byte[], uint> GetNonce(byte[] prikey)
         {
-            byte[] alpha;
-            if (Block.Index > 1000)
-            {
-                // To prevent the primary uses the same block hash in Height (1000, 2000)
-                // also add the Height to the VRF input.
-                alpha = NativeContract.Ledger
-                    .GetBlockHash(Snapshot, Block.Index - 1000)
+
+            // To prevent the primary uses the same block hash in Height (1000, 2000)
+            // also add the Height to the VRF input.
+            byte[] alpha = NativeContract.Ledger
+                    .GetBlockHash(Snapshot, (uint)(Block.Index - (Block.Index > 1000 ? 1000 : 0)))
                     .ToArray()
                     .Concat(BitConverter.GetBytes(Block.Index))
                     .ToArray();
-            }
-            else
-            {
-                alpha = Block.PrevHash.ToArray();
-            }
 
             var proof = VRF.Prove(prikey, alpha);
             var nonce = VRF.ProofToHash(proof);
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -329,7 +329,6 @@ namespace Neo.Consensus
                 UInt256.Length +    // MerkleRoot
                 sizeof(ulong) +     // Timestamp
                 sizeof(uint) +      // Index
-                sizeof(uint) +      // Nonce
                 sizeof(byte) +      // PrimaryIndex
                 UInt160.Length +    // NextConsensus
                 1 + _witnessSize +  // Witness
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -338,15 +338,23 @@ namespace Neo.Consensus
 
         private Tuple<byte[], uint> GetNonce(byte[] prikey)
         {
-            UInt256 aplha;
+            byte[] aplha;
             if (Block.Index > 1000)
-                aplha = NativeContract.Ledger.GetBlockHash(Snapshot, Block.Index - 1000);
+            {
+                // To prevent the primary uses the same prevHash in Height (1000, 2000)
+                // also add the Height to the VRF input.
+                aplha = NativeContract.Ledger
+                    .GetBlockHash(Snapshot, Block.Index - 1000)
+                    .ToArray()
+                    .Concat(BitConverter.GetBytes(Block.Index))
+                    .ToArray();
+            }
             else
             {
-                aplha = Block.PrevHash;
+                aplha = Block.PrevHash.ToArray();
             }
 
-            var proof = VRF.Prove(prikey, aplha.ToArray());
+            var proof = VRF.Prove(prikey, aplha);
             var nonce = VRF.ProofToHash(proof);
             VRFProof = proof;
             return new Tuple<byte[], uint>(proof, BitConverter.ToUInt32(nonce[..4]));
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -234,7 +234,6 @@ namespace Neo.Consensus
             return p >= 0 ? (byte)p : (byte)(p + Validators.Length);
         }
 
-
         public UInt160 GetSender(int index)
         {
             return Contract.CreateSignatureRedeemScript(Validators[index]).ToScriptHash();
--- a/src/DBFTPlugin/ConsensusContext.cs
+++ b/src/DBFTPlugin/ConsensusContext.cs
@@ -127,6 +127,7 @@ namespace Neo.Consensus
                 Category = "dBFT",
                 ValidBlockStart = 0,
                 ValidBlockEnd = message.BlockIndex,
+                //VRFProof = GetNonce(keyPair.PrivateKey, Block.PrevHash.ToArray()),
                 Sender = GetSender(message.ValidatorIndex),
                 Data = message.ToArray(),
                 Witness = invocationScript is null ? null : new Witness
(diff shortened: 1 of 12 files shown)


src/DBFTPlugin/PrepareRequest.cs
```diff
@@ -10,12 +10,14 @@ public class PrepareRequest : ConsensusMessage
...
         public UInt256 PrevHash;
         public ulong Timestamp;
+        public byte[] VRFProof;
         public UInt256[] TransactionHashes;
 
         public override int Size => base.Size
             + sizeof(uint)                      //Version
             + UInt256.Length                    //PrevHash
             + sizeof(ulong)                     //Timestamp
+            + VRFProof.GetVarSize()             // Nonce
```
shargon, line 20: If it's fixed size (81) we should use 81
```suggestion
            + sizeof(ulong)                     //Timestamp
            + 81                                // Nonce
```

src/DBFTPlugin/ConsensusService.cs
```diff
@@ -423,17 +423,39 @@ private void OnPrepareRequestReceived(ExtensiblePayload payload, PrepareRequest
...
+            if (message.BlockIndex > 1000)
+                nonce = VRF.Verify(context.Validators[message.ValidatorIndex], message.VRFProof, NativeContract.Ledger.GetBlockHash(context.Snapshot, message.BlockIndex - 1000).ToArray());
+            else
+            {
+                nonce = VRF.Verify(context.Validators[message.ValidatorIndex], message.VRFProof, message.PrevHash.ToArray());
+            }
+
+            if (nonce == null)
+            {
+                Log($"Random number verification failed: {message.VRFProof}", LogLevel.Warning);
```
shargon, line 446: I prefer throw an error in random method instead of break the consensus

src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -232,6 +234,7 @@ public byte GetPrimaryIndex(byte viewNumber)
             return p >= 0 ? (byte)p : (byte)(p + Validators.Length);
         }
 
+
         public UInt160 GetSender(int index)
```
shargon, line 238:
```suggestion
        public UInt160 GetSender(int index)
```

src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -327,12 +329,37 @@ internal int GetExpectedBlockSizeWithoutTransactions(int expectedTransactions)
...
         }
 
+        private Tuple<byte[], uint> GetNonce(byte[] prikey)
+        {
+            byte[] aplha;
+            if (Block.Index > 1000)
+            {
+                // To prevent the primary uses the same block hash in Height (1000, 2000)
+                // also add the Height to the VRF input.
+                aplha = NativeContract.Ledger
```
vncoelho, line 346: alpha?

src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -327,12 +329,37 @@ internal int GetExpectedBlockSizeWithoutTransactions(int expectedTransactions)
...
                 sizeof(byte) +      // PrimaryIndex
                 UInt160.Length +    // NextConsensus
                 1 + _witnessSize +  // Witness
                 IO.Helper.GetVarSize(expectedTransactions);
         }
 
+        private Tuple<byte[], uint> GetNonce(byte[] prikey)
+        {
+            byte[] aplha;
+            if (Block.Index > 1000)
```
vncoelho, line 342: Maybe the % with 1000?

src/DBFTPlugin/ConsensusService.cs
vncoelho: And maybe ask ChangeView.

src/DBFTPlugin/ConsensusContext.cs
Liaojinghui: `alpha` is the term used by draft-irtf-cfrg-vrf-08 to name the vrf input. Check it https://tools.ietf.org/pdf/draft-irtf-cfrg-vrf-08

vncoelho: I mean change `aplha -> alpha`

Liaojinghui: lol，sorry  did not notice the typo.

src/DBFTPlugin/ConsensusService.cs
Liaojinghui: > And maybe ask ChangeView.

I just follow the way how you process other check failures. If you think ChangeView is better, I can add it here.

Liaojinghui: > I prefer throw an error in random method instead of break the consensus

Updated by adding exception

src/DBFTPlugin/ConsensusService.cs
```diff
@@ -423,17 +423,35 @@ private void OnPrepareRequestReceived(ExtensiblePayload payload, PrepareRequest
...
                 Log($"Invalid request: transaction already exists", LogLevel.Warning);
                 return;
             }
 
+            // Possible vulnerability of the random number, dishonest nodes cooperate to
+            // attck the system. 
+            byte[] nonce;
+            try
+            {
+                nonce = VRF.Verify(context.Validators[message.ValidatorIndex], message.VRFProof, message.PrevHash);
```
shargon, line 438: We should add the current merkletree in `data`

Liaojinghui: If we add the current Merkle tree in the `data`, the result would be mutable for the primary cause he can decide the Merkle tree.

shargon: >  he can decide the Merkle tree.

But not the `PrevHash`, it makes the things more complicated

Liaojinghui: Using the block hash is a normal and the most easy way to get a well known input. Currently, almost all projects that use VRF take block hash as input, either prehash or previous n blocks. But if u have a better input source, definitely it is a good thing.

src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -530,5 +551,49 @@ public void Serialize(BinaryWriter writer)
...
+            var tx = new Transaction()
+            {
+                Version = 0,
+                Nonce = nonce,
+                ValidUntilBlock = blockIndex + settings.MaxValidUntilBlockIncrement,
+                Signers = new[]
+               {
+                    new Signer
+                    {
+                        Account = pubkey.ToScriptHash(),
```
nicolegys, line 574: `pubkey.ToScriptHash()` gets a PublicKeyHash, not the account scriptHash

nicolegys (changes requested):

src/DBFTPlugin/ConsensusContext.cs
Liaojinghui: > `pubkey.ToScriptHash()` gets a PublicKeyHash, not the account scriptHash

`Contract.CreateSignatureRedeemScript(PublicKey).ToScriptHash()`?

shargon: > Contract.CreateSignatureRedeemScript(PublicKey).ToScriptHash()? @nicolegys

Yes

Liaojinghui: @shargon thank you shargon

nicolegys:
Sometimes run into an exception.
31bytes
image
image

Liaojinghui:
I did not get it? Should not it be right to throw exception? The size should be 32 bytes.

nicolegys:
@Liaojinghui I think here's why
image

Liaojinghui:
> @Liaojinghui I think here's why
> image

Fixed the error, the issue is neo added extra `length` check to the `prikey` @nicolegys

src/DBFTPlugin/VRF.cs
```diff
@@ -0,0 +1,365 @@
...
+            // Step 4: Gamma = x * H
+            var gamma_point = h_point * prikey;
+
+            // Step 5: k = ECVRF_nonce_generation(SK, h_string)
+            var k = GenerateNonce(prikey, h_string);
+            var kBytes = k.ToByteArray(true, true);
+
+            // Step 6: c = ECVRF_hash_points(H, Gamma, k*B, k*H)
+            var u_point = DerivePubkeyPoint(k.ToByteArray(true, true));
+            var v_point = h_point * kBytes;
```
nicolegys, line 222: > Fixed the error, the issue is neo added extra `length` check to the `prikey` @nicolegys

We should append leading zeros to `kBytes` after `var kBytes = k.ToByteArray(true, true)` in `Prove` method, because the operator `*` requires the second param to be 32 bits.

nicolegys:
I sent a tx to extract neo from the CNs' multi-sig address, then the CNs threw exception and stop generating blocks, and cannot recover.
image

Liaojinghui:
well, old blocks dont have such nonce tx. maybe we should check whether a block contains a nonce tx.

Liaojinghui:
Fixed, I changed the index of `nonce_tx` in the txs, and missed checking its verification code. @nicolegys

nicolegys:
> Fixed, I changed the index of `nonce_tx` in the txs, and missed checking its verification code. @nicolegys
> Can we have a test project for the dbft~

I used the newest code, and the issue still existed. TwT

src/DBFTPlugin/VRF.cs
nicolegys: @Liaojinghui I think you've forgotten this issue 0.0

Liaojinghui: > @Liaojinghui I think you've forgotten this issue 0.0

My fault TT

nicolegys:
Failed to generate block_1 in multiple CN mode.
Cause:
The primary (index=1) create a nonce_tx but it has no gas to pay for the network fee, then throw exception `Insufficient GAS`.

image

nicolegys:
If view changed, tx.Nonce is still the one created by old primary, but the backup CNs receive the new nonce, so `CheckNonce` will fail.
image

src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -342,7 +354,8 @@ internal void EnsureMaxBlockLimitation(IEnumerable<Transaction> txs)
             uint maxTransactionsPerBlock = neoSystem.Settings.MaxTransactionsPerBlock;
 
             // Limit Speaker proposal to the limit `MaxTransactionsPerBlock` or all available transactions of the mempool
-            txs = txs.Take((int)maxTransactionsPerBlock);
+            txs = txs.Take((int)maxTransactionsPerBlock - 1);
```
nicolegys, line 357:
```suggestion
            txs = txs.Take((int)maxTransactionsPerBlock);
```

nicolegys (commented):
No nonce_tx now.

nicolegys:
Could you please also update `HeaderFromJson` in `RpcClient`?
https://github.com/Liaojinghui/neo-modules/blob/master/src/RpcClient/Utility.cs#L137

src/RpcClient/Utility.cs
```diff
@@ -142,7 +142,7 @@ public static Header HeaderFromJson(JObject json, ProtocolSettings protocolSetti
                 PrevHash = UInt256.Parse(json["previousblockhash"].AsString()),
                 MerkleRoot = UInt256.Parse(json["merkleroot"].AsString()),
                 Timestamp = (ulong)json["time"].AsNumber(),
-                Nonce = (ulong)json["nonce"].AsNumber(),
+                Nonce = UInt64.Parse(json["nonce"].AsString()),
```
nicolegys, line 145: I think you should use`UInt64.Parse(json["nonce"].AsString(), NumberStyles.AllowHexSpecifier)` to convert a hex string to uint64.

PR #596, opened 2021-06-10
Add Random number to N3

Liaojinghui:
Close pull request #590


src/DBFTPlugin/ConsensusContext.cs
```diff
@@ -397,13 +411,16 @@ public ExtensiblePayload MakeRecoveryMessage()
...
+                var (proof, nonce) = GetNonce(keyPair.PrivateKey);
+                Block.Header.Nonce = nonce;
                 prepareRequestMessage = new PrepareRequest
                 {
                     Version = Block.Version,
                     PrevHash = Block.PrevHash,
                     ViewNumber = ViewNumber,
                     Timestamp = Block.Timestamp,
                     BlockIndex = Block.Index,
+                    VRFProof = proof, // Add vrf proof to the prepare request
```
ZhangTao1596, line 423: You should use the proof in `RrepareRequest` instead creating a new one. Proof should be caculated from Primary private key.

nicolegys: Yes, creating a new one will cause issue when stopping the CNs at different stages. They can no longer achieve the consensus.
image

Liaojinghui: > You should use the proof in `RrepareRequest` instead creating a new one. Proof should be caculated from Primary private key.

I did not get that, isn't that only the primary could send the recovery message?

src/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -256,178 +205,6 @@ public bool Load()
...
-
-        public ExtensiblePayload MakePrepareResponse()
-        {
-            return PreparationPayloads[MyIndex] = MakeSignedPayload(new PrepareResponse
-            {
-                PreparationHash = PreparationPayloads[Block.PrimaryIndex].Hash
-            });
-        }
-
         public void Reset(byte viewNumber)
```
nicolegys, line 208: `nonce` needs to be reset, too

nicolegys:
Test passed, waiting for review.

shargon:
Too many consensus changes, could you split the refactor and the random number in two different pr?

Liaojinghui:
> Too many consensus changes, could you split the refactor and the random number in two different pr?

I also thought about create another pr, but if so, I will either have to wait until random is merged or face tons of conflicts afterward. Since I just make big class into smaller partial files and did not change any code, it will not cause any issue to the logic.

Qiao-Jin (dismissed):

Qiao-Jin:
Please correct format

src/DBFTPlugin/DBFTPlugin.csproj
```diff
@@ -5,15 +5,37 @@
...
+    <CustomCommands>
+      <CustomCommands>
+        <Command>
+          <type>BeforeBuild</type>
+          <command>dotnet-format</command>
+          <externalConsole>True</externalConsole>
+        </Command>
+      </CustomCommands>
+    </CustomCommands>
+  </PropertyGroup>
```
shargon, line 18: If we add this we should add it to all the projects, I think that it can be removed

Liaojinghui: lol, i just think this would save a lot of work in format updating ( or is it just me who make a lot of format errors?).  it wont hurt even if we add it to all projects, right?

shargon: I think that it's not a bad change, but seems weird if only in one project was added, isn't it?

superboyiii: > @shargon lol, i just think this would save a lot of work in format updating ( or is it just me who make a lot of format errors?). it wont hurt even if we add it to all projects, right?

I think you could add to the whole project.

superboyiii:
https://github.com/Liaojinghui/neo-modules/blob/f12250decb632e57aec04522217a699e28ddc424/src/Directory.Build.props#L18
Please change to `3.0.0-CI01285`. @Liaojinghui

Qiao-Jin:
Ping

superboyiii:
@shargon Pls review again.

src/ApplicationLogs/ApplicationLogs.csproj
```diff
@@ -5,6 +5,17 @@
     <RootNamespace>Neo.Plugins</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
```
shargon, line 8: It could be done in `Directory.Build.props`?

Liaojinghui: > It could be done in `Directory.Build.props`?

Well, let's find out

Liaojinghui: @shargon unfortunately, I did not figure out a way to do that.

superboyiii:
@shargon Pls review again.

superboyiii (approved):

erikzhang (approved):

PR #599, opened 2021-06-18
Add ContentTypeNotSupported OracleResponseCode

superboyiii:
Require: https://github.com/neo-project/neo/pull/2505
Close: https://github.com/neo-project/examples/issues/39

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo" Version="3.0.0-CI01290" />
+    <PackageReference Include="Neo" Version="3.0.0-CI01282" />
   </ItemGroup>
 
 </Project>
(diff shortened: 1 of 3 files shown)


shargon:
@superboyiii NeoFS.API is not managed by 'neo-project' but it has a dependence of 'neo' , so if we want to upgrade the neo nuget, we can't do that only in this repository

ZhangTao1596:
NeoFS.API updated. `<PackageReference Include="NeoFS.API" Version="0.0.53" />`

ZhangTao1596:
UT need fix.

superboyiii:
> @superboyiii NeoFS.API is not managed by 'neo-project' but it has a dependence of 'neo' , so if we want to upgrade the neo nuget, we can't do that only in this repository

What's this? - -!
image

superboyiii:
> @superboyiii NeoFS.API is not managed by 'neo-project' but it has a dependence of 'neo' , so if we want to upgrade the neo nuget, we can't do that only in this repository

Done

superboyiii:
@shargon Could you review again?

shargon (approved):

PR #604, opened 2021-06-30
Fix gaslimit

Ashuaidehao:
Relate https://github.com/neo-project/neo/pull/2493

### DIFF
--- a/src/RpcServer/RpcServer.Wallet.cs
+++ b/src/RpcServer/RpcServer.Wallet.cs
@@ -279,7 +279,7 @@ namespace Neo.Plugins
             tx.Witnesses = transContext.GetWitnesses();
             if (tx.Size > 1024)
             {
-                long calFee = tx.Size * 1000 + 100000;
+                long calFee = tx.Size * NativeContract.Policy.GetFeePerByte(snapshot) + 100000;
                 if (tx.NetworkFee < calFee)
                     tx.NetworkFee = calFee;
             }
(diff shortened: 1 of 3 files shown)


src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -228,7 +228,7 @@ protected virtual JObject SendFrom(JArray _params)
             tx.Witnesses = transContext.GetWitnesses();
             if (tx.Size > 1024)
             {
-                long calFee = tx.Size * 1000 + 100000;
+                long calFee = tx.Size * NativeContract.Policy.GetFeePerByte(snapshot) + 100000;
```
ProDog, line 231: `SendMany` and `SendToAddress`  also needs to be modified.

ProDog (approved):

src/StateService/Network/StateRoot.cs
```diff
@@ -90,7 +90,7 @@ public void SerializeUnsigned(BinaryWriter writer)
 
         public bool Verify(ProtocolSettings settings, DataCache snapshot)
         {
-            return this.VerifyWitnesses(settings, snapshot, 1_00000000);
+            return this.VerifyWitnesses(settings, snapshot, 3_00000000L);
```
shargon, line 93: It's not enough with 2?

shargon (approved):
tested?

src/StateService/Network/StateRoot.cs
ProDog: But it's always `MaxVerificationGas` 🤔.
https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Helper.cs#L275

ProDog:
> tested?

Tested OK.

Issue #606, opened 2021-07-01
Expired Oracle Response Transaction Not Removed


cloud8little:
**Describe the bug**
relate to https://github.com/neo-project/neo/issues/2516

Currently when constructing a response tx, its validuntilblock is added by the requestTx index. https://github.com/neo-project/neo-modules/blob/master/src/OracleService/OracleService.cs#L328 

https://github.com/neo-project/neo-modules/blob/9d1166a86309a54095b595a27b49fba2c850c864/src/OracleService/OracleService.cs#L317-L342

which will be smaller than the latest block height, and still the tx is not removed, therefore will be reserved in the oracle pending list

Issue #607, opened 2021-07-02
Oracle Response with different result always stay in the queue 


cloud8little:
**Describe the bug**

Each oracle node will construct its oracle response with the signature,

https://github.com/neo-project/neo-modules/blob/9d1166a86309a54095b595a27b49fba2c850c864/src/OracleService/OracleService.cs#L244-L278 and then send the submitoracleresponse request to other oracle nodes.

when another oracle node receive the rpc request, it will get the url response and construct a new response tx body, and then verify the signature, since the result data is different(for example, a url return a random number), it will fail to check the signature, and then fail to relay a complete response tx. 
https://github.com/neo-project/neo-modules/blob/9d1166a86309a54095b595a27b49fba2c850c864/src/OracleService/OracleService.cs#L175-L199

expect result is to deal with in the backup sign tx in the first piece code, while when massive oracle request continuously received, it will be out of the refresh interval. and it will not be able to be dealt with afterwards.

PR #608, opened 2021-07-02
Fix oracle

Ashuaidehao:
Close https://github.com/neo-project/neo-modules/issues/606 https://github.com/neo-project/neo-modules/issues/607


src/OracleService/OracleService.cs
```diff
@@ -314,18 +315,23 @@ private async Task<(OracleResponseCode, string)> ProcessUrlAsync(string url)
...
             var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
             var n = oracleNodes.Length;
             var m = n - (n - 1) / 3;
             var oracleSignContract = Contract.CreateMultiSigContract(m, oracleNodes);
-
+            uint height = NativeContract.Ledger.CurrentIndex(snapshot);
+            var validUntilBlock = requestTx.BlockIndex + settings.MaxValidUntilBlockIncrement;
+            while (useCurrentHeight && validUntilBlock <= height)
+            {
+                validUntilBlock += settings.MaxValidUntilBlockIncrement;
```
shargon, line 328: We could check that `MaxValidUntilBlockIncrement` is not 0, but "is not" possible...

```csharp
public uint MaxValidUntilBlockIncrement => 86400000 / MillisecondsPerBlock;
```

shargon (approved):

cloud8little:
Test failed. 

635ee931e9daf12d8e1f8fb0d4ed844

when response transaction is expired, pendingQueue will remove it and thus it can not be dealt with in the onTimer function, then the backup transaction can not be relayed.

cloud8little:
The request queue is finally empty, while the throughput of dealing with the expired response transactions is very low.(about 1hour less than 40requests), which is really inefficient.

cloud8little:
While the requests are dealt with in a very low speed, the function test passed.

cloud8little (approved):

superboyiii:
@shargon Pls review again.

superboyiii:
Ping @shargon

src/OracleService/OracleService.cs
```diff
@@ -314,18 +327,23 @@ private async Task<(OracleResponseCode, string)> ProcessUrlAsync(string url)
...
 
-        public static Transaction CreateResponseTx(DataCache snapshot, OracleRequest request, OracleResponse response, ECPoint[] oracleNodes, ProtocolSettings settings)
+        public static Transaction CreateResponseTx(DataCache snapshot, OracleRequest request, OracleResponse response, ECPoint[] oracleNodes, ProtocolSettings settings, bool useCurrentHeight = false)
         {
             var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
             var n = oracleNodes.Length;
             var m = n - (n - 1) / 3;
             var oracleSignContract = Contract.CreateMultiSigContract(m, oracleNodes);
-
+            uint height = NativeContract.Ledger.CurrentIndex(snapshot);
```
shargon, line 336: I think that this can produce a different responseTx, if the tx doesn't arrive in the same block to all nodes, it coud produce an error. We should try to unify this

Ashuaidehao: The probability of produceing diffrent responseTx is about 1/5760. When it happens, it will be removed from queue when reach "MaxTaskTimeout", then the request will re-enqueue the ProcessRequestAsync method. 
These steps can repeat many times until this oracle request was processed successfully.

shargon: I think that we should reduce this probability, during this time we have the transaction in memory all because this field.

Ashuaidehao: Yes. There are still some more serious problems with current oracle architecture. See https://github.com/neo-project/neo-modules/issues/611 .
I think we should refact the oracle service to resolve these problems.

roman-khimov: I think this particular problem is actually way less probable than 1/5760. It's only when `height > validUntilBlock` the increment is triggered and for this to happen we have to be already `MaxValidUntilBlockIncrement` blocks away from `requestTx.BlockIndex` which is a 100 blocks. On a healthy network I'd expect most of oracle requests to be handled within 1-2 blocks from `requestTx.BlockIndex` with maybe something like 99% of requests completing in 10 blocks. Then of course there is a long tail of less successful requests with some crossing this `MaxValidUntilBlockIncrement` line, but even then oracle nodes would have to be a little unsynchronized to trigger different response tx problem. Then when every oracle nodes crosses `requestTx.BlockIndex + MaxValidUntilBlockIncrement` there is another ~`MaxValidUntilBlockIncrement` window of time where this problem won't occur.

So it should be fine as a fallback mechanism (which it is now). Not perfect, but this problem doesn't seem to have any perfect solution.

#611 is more interesting though.

shargon: If two nodes produces a different `ValidUntilBlock`, his signatures wont be valid. So if the tx arrives in different blocks to oracles nodes (I think that it could be very easier) it won't be valid.

roman-khimov: > if the tx arrives in different blocks to oracles nodes

But isn't `requestTx.BlockIndex` is a block number that contained `requestTx` and there could only be one block with this tx ever?
```
            var requestTx = NativeContract.Ledger.GetTransactionState(snapshot, request.OriginalTxid);
```

shargon: You are right, I saw a different thing 😆

shargon (approved):
Tested?

superboyiii:
@cloud8little If retest OK I'll merge.

cloud8little:
Test passed, issue #611 won't occur when oracle requests arrived continuously, now  submitoracleresponse rpc requests received by each node is about 5+ times of the number of requests, which IMO is reasonable.

cloud8little (approved):

PR #626, opened 2021-08-11
Dispose store

Liaojinghui:
It is possible to crash the leveldb if the connection to the database is not closed properly.

Reproduce this issue:

run `restart` command here https://github.com/neo-project/neo-node/pull/809

### DIFF
--- a/src/LevelDBStore/Plugins/Storage/LevelDBStore.cs
+++ b/src/LevelDBStore/Plugins/Storage/LevelDBStore.cs
@@ -10,22 +10,22 @@ namespace Neo.Plugins.Storage
     {
         public override string Description => "Uses LevelDB to store the blockchain data";
 
-        private List<WeakReference> dbs = new();
+        private Stack<WeakReference> dbs = new();
 
         public IStore GetStore(string path)
         {
             if (Environment.CommandLine.Split(' ').Any(p => p == "/repair" || p == "--repair"))
                 DB.Repair(path, Options.Default);
             IStore store = new Store(path);
-            dbs.Add(new WeakReference(store));
+            dbs.Push(new WeakReference(store));
             return store;
         }
         public override void Dispose()
         {
             base.Dispose();
-            foreach (var a in dbs)
+            while (dbs.Count > 0)
             {
-                IStore store = a.Target as IStore;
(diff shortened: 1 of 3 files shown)


Liaojinghui:
Crash exception if db connection not closed properly.
```C#
System.Runtime.InteropServices.SEHException (0x80004005): External component has thrown an exception.
   at Neo.IO.Data.LevelDB.Native.leveldb_open(IntPtr options, String name, IntPtr& error)
   at Neo.IO.Data.LevelDB.DB.Open(String name, Options options)
```

Liaojinghui:
This problem happened to me yesterday, at first, it can be recovered by restarting the program, but after dozens of times, leveldb crashed permanently, even a computer reboot can not fix it.

erikzhang:
```
./neo-cli --repair
```

Liaojinghui:
> ```
> ./neo-cli --repair
> ```

Tried, did not work, even tried to delete the DB folder and replace libleveldb. Only made it work again by deleting the whole project and reclone from Github.

src/LevelDBStore/Plugins/Storage/LevelDBStore.cs
```diff
@@ -9,11 +10,25 @@ public class LevelDBStore : Plugin, IStorageProvider
...
         public override string Description => "Uses LevelDB to store the blockchain data";
 
+        private Stack<WeakReference> dbs = new();
+
         public IStore GetStore(string path)
         {
             if (Environment.CommandLine.Split(' ').Any(p => p == "/repair" || p == "--repair"))
                 DB.Repair(path, Options.Default);
-            return new Store(path);
+            IStore store = new Store(path);
```
Qiao-Jin, line 19: Dispose other references in dbs upon GetStore?

Liaojinghui: I have no idea whether those db connections are closed or not, therefore I have to keep a weak reference of them in a stack.
In this way, I can make sure that all connections are closed in Dispose with the slightest modification to existing code.

erikzhang: This method is designed for the user to be responsible for releasing the db connection.

PR #632, opened 2021-08-18
fix IPersistencePlugin

### LINKED ISSUE #631: Misuse of IPersistencePlugin
IPersistencePlugin is invented to allow plugins to observe the persisted blocks. There are currently several plugins implementing this interface. However, misuses exist within them.

According to  https://github.com/neo-project/neo/pull/568:
> If there are multiple persistence plugins loaded and one of them throws an exception, then none of them should actually commit anything to a persistent store. To facilitate this behavior, I added an OnCommit method to be called after all plugins have had a chance to process results from OnPersist and throw if something was wrong.
A persistence plugin implementing IPersistencePlugin should do any processing needed to stage data to persist in its OnPersist method, and then perform actual persistence in its OnCommit method. In this way database integrity is maximized.

IPersistencePlugin should not persist data directly in _OnPersist_. But such practises do exist in LogReader & StatePlugin.

Qiao-Jin:
Close #631

### DIFF
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -16,6 +16,7 @@ namespace Neo.Plugins
     public class LogReader : Plugin, IPersistencePlugin
     {
         private DB db;
+        private WriteBatch _writeBatch;
 
         public override string Name => "ApplicationLogs";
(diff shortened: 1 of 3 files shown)


shargon (approved):
tested?

superboyiii:
> tested?

I will test this.

superboyiii:
Test OK.

PR #634, opened 2021-08-23
[StateService] Allow value empty

### LINKED ISSUE #633: StateService: Unexpected MPT behaviour for contract storage items with an empty value
**Describe the bug**
It is not prohibited for contracts to put storage items with an *empty value* in the storage. However, these storage changes are not correctly reflected in MPT, i.e. if a storage item with an empty value is put into storage, then MPT considers the corresponding node as the one to be removed and will [try to delete it](https://github.com/neo-project/neo-modules/blob/ef59a0eb27fb23c49610c0362e1fe7d9d0ec7839/src/StateService/MPT/MPTTrie.Put.cs#L30). As a result, the current implementation allows an Empty MPT node to point to a storage item with nil value.

So, the unwanted consequences of described behaviour are:
1. **Broken MPT proofs**. The current implementation is unable to provide (and verify) proof for storage items with nil value, because they are not included in the MPT. Although the contract itself may easily put and retrieve such items from the storage.
2. **Incomplete MPT state**. As far as storage items with empty values are not included in MPT and we can’t distinguish EmptyNode from the one pointing to SI with nil value, it’s impossible to retrieve such items from MPT. MPT just does not contain complete information about all contract storage items, which adds complexity overhead to [P2P state exchange process](https://github.com/neo-project/neo/issues/2373#issuecomment-853354023) implementation.

Described problems are relevant for both native (i.g. [NEO](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/NeoToken.cs#L138), [Policy](https://github.com/neo-project/neo/blob/master/src/neo/SmartContract/Native/PolicyContract.cs#L133)) and non-native contracts.

**Expected behaviour**
Operations with storage items with an empty value should be properly reflected in MPT. I.e.:
- Storage.Put(key, System.Array.Empty<byte>()) should result in a new LeafNode with an empty value added to the MPT.
- Storage.Delete(key) should result in removal of this LeafNode from the MPT using regular rules.

So that it will be possible to get proof for a storage item with an empty value, verify this proof and completely restore the whole set of contract storage items given MPT leaves.

**Platform:**
 - Version: neo-modules v3.0.1

ZhangTao1596:
Fix #633

### DIFF
--- a/src/StateService/MPT/MPTTrie.Delete.cs
+++ b/src/StateService/MPT/MPTTrie.Delete.cs
@@ -10,7 +10,7 @@ namespace Neo.Plugins.MPT
         public bool Delete(TKey key)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key));
             return TryDelete(ref root, path);
         }
 
--- a/src/StateService/MPT/MPTTrie.Delete.cs
+++ b/src/StateService/MPT/MPTTrie.Delete.cs
@@ -10,7 +10,7 @@ namespace Neo.Plugins.MPT
         public bool Delete(TKey key)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) return false;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
             return TryDelete(ref root, path);
         }
 
--- a/src/StateService/MPT/MPTTrie.Proof.cs
+++ b/src/StateService/MPT/MPTTrie.Proof.cs
@@ -13,7 +13,7 @@ namespace Neo.Plugins.MPT
         public bool TryGetProof(TKey key, out HashSet<byte[]> proof)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key));
             proof = new HashSet<byte[]>(ByteArrayEqualityComparer.Default);
             return GetProof(ref root, path, proof);
         }
--- a/src/StateService/MPT/MPTTrie.Proof.cs
+++ b/src/StateService/MPT/MPTTrie.Proof.cs
@@ -10,13 +10,12 @@ namespace Neo.Plugins.MPT
 {
     partial class MPTTrie<TKey, TValue>
     {
-        public HashSet<byte[]> GetProof(TKey key)
+        public bool TryGetProof(TKey key, out HashSet<byte[]> proof)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) return null;
-            HashSet<byte[]> set = new HashSet<byte[]>(ByteArrayEqualityComparer.Default);
-            if (!GetProof(ref root, path, set)) return null;
-            return set;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
+            proof = new HashSet<byte[]>(ByteArrayEqualityComparer.Default);
+            return GetProof(ref root, path, proof);
         }
 
         private bool GetProof(ref MPTNode node, ReadOnlySpan<byte> path, HashSet<byte[]> set)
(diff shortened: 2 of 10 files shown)


src/StateService/MPT/MPTTrie.Put.cs
```diff
@@ -23,12 +23,8 @@ public bool Put(TKey key, TValue value)
         {
             var path = ToNibbles(key.ToArray());
             var val = value.ToArray();
-            if (path.Length == 0 || path.Length > MPTNode.MaxKeyLength)
+            if (path.Length > MPTNode.MaxKeyLength || val.Length > MPTNode.MaxValueLength)
```
AnnaShaleva, line 26: Technically, we can allow having empty paths in MPT and MPT will be OK with that. But practically, where do empty paths come from? Even if `System.Storage.Put` allows to use an empty key, all storage items have at least contract ID bytes as the key prefix, so the storage item key is always non-empty. Thus, if MPT should put an item with an empty key, then it's a bug somewhere.

I consider it's safer to keep an old behaviour for the `path` (do not allow 0-lenght `path`).

AnnaShaleva (changes requested):

src/StateService/MPT/MPTTrie.Put.cs
ZhangTao1596: Done.

src/StateService/MPT/MPTTrie.Delete.cs
```diff
@@ -10,7 +10,7 @@ partial class MPTTrie<TKey, TValue>
         public bool Delete(TKey key)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) return false;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
```
shargon, line 13:
```suggestion
            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key));
```

src/StateService/Storage/StateStore.cs
```diff
@@ -54,11 +54,11 @@ public StateSnapshot GetSnapshot()
...
-        public HashSet<byte[]> GetProof(UInt256 root, StorageKey skey)
+        public bool TryGetProof(UInt256 root, StorageKey skey, out HashSet<byte[]> proof)
         {
             using ISnapshot snapshot = store.GetSnapshot();
             var trie = new MPTTrie<StorageKey, StorageItem>(snapshot, root);
-            return trie.GetProof(skey);
+            return trie.TryGetProof(skey, out proof);
         }
 
         protected override void OnReceive(object message)
```
shargon, line 64: On `UpdateLocalStateRoot` we don't check the result of `Trie.Put`

ZhangTao1596: I will remove the bool result. Put is always expected to be successful or throwing exception.

src/StateService/MPT/MPTTrie.Proof.cs
```diff
@@ -10,13 +10,12 @@ namespace Neo.Plugins.MPT
...
     {
-        public HashSet<byte[]> GetProof(TKey key)
+        public bool TryGetProof(TKey key, out HashSet<byte[]> proof)
         {
             var path = ToNibbles(key.ToArray());
-            if (path.Length == 0) return null;
-            HashSet<byte[]> set = new HashSet<byte[]>(ByteArrayEqualityComparer.Default);
-            if (!GetProof(ref root, path, set)) return null;
-            return set;
+            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key)); ;
```
shargon, line 16:
```suggestion
            if (path.Length == 0) throw new ArgumentException("could not be empty", nameof(key));
```

superboyiii:
> @superboyiii Can you test these optimizations?

Sure

roman-khimov:
Any update on this?

ZhangTao1596:
> Any update on this?

I'm waiting @superboyiii 's test feedback.

superboyiii (approved):
Tested OK.
Set an empty value in storage and finally got and verified proof successfully. (Currently returns unknown value error on testnet.)

superboyiii:
@shargon @roman-khimov

roman-khimov (approved):

AnnaShaleva (approved):

shargon (approved):

Issue #635, opened 2021-08-23
RpcClient missing state root related methods


devhawk:
* https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/getstateheight.html
* https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/getstateroot.html
* https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/getproof.html
* https://docs.neo.org/docs/en-us/reference/rpc/latest-version/api/verifyproof.html

Issue #629, opened 2021-08-17
Support Iterator in RpcClient


RichardBelsum:
https://github.com/neo-project/neo-modules/blob/ef59a0eb27fb23c49610c0362e1fe7d9d0ec7839/src/RpcClient/Utility.cs#L214

PR #638, opened 2021-08-30
[StateService] get historical state

ZhangTao1596:
Changes:
    * Add `GetState` `FindStates` API to get historical storage or list storages by prefix
    * Add `README` 
@devhawk @shargon

### DIFF
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -74,6 +74,8 @@ namespace Neo.Plugins.MPT
                 offset = prefix.Length * 2;
                 from = ToNibbles(from.AsSpan());
             }
+            if (path.Length > MPTNode.MaxKeyLength || from.Length > MPTNode.MaxKeyLength)
+                throw new ArgumentException("exceeds limit");
             path = Seek(ref root, path, out MPTNode start).ToArray();
             return Travers(start, path, from, offset)
                 .Select(p => (FromNibbles(p.Key).AsSerializable<TKey>(), p.Value.AsSerializable<TValue>()));
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -65,26 +65,29 @@ namespace Neo.Plugins.MPT
         public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix, byte[] from = null)
         {
             var path = ToNibbles(prefix);
+            int offset = 0;
             if (from is null) from = Array.Empty<byte>();
             if (0 < from.Length)
             {
                 if (!from.AsSpan().StartsWith(prefix))
                     throw new InvalidOperationException("invalid from key");
-                from = ToNibbles(from[prefix.Length..].AsSpan());
+                offset = prefix.Length * 2;
+                from = ToNibbles(from.AsSpan());
             }
             path = Seek(ref root, path, out MPTNode start).ToArray();
-            return Travers(start, path, from)
+            return Travers(start, path, from, offset)
                 .Select(p => (FromNibbles(p.Key).AsSerializable<TKey>(), p.Value.AsSerializable<TValue>()));
         }
 
-        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path, byte[] from)
+        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path, byte[] from, int offset)
         {
             if (node is null) yield break;
+            if (offset < 0) throw new InvalidOperationException("invalid offset");
             switch (node.Type)
             {
                 case NodeType.LeafNode:
                     {
-                        if (from.Length == 0)
+                        if (from.Length <= offset && !path.SequenceEqual(from))
                             yield return (path, (byte[])node.Value.Clone());
                         break;
                     }
@@ -95,31 +98,31 @@ namespace Neo.Plugins.MPT
                         var newNode = cache.Resolve(node.Hash);
                         if (newNode is null) throw new InvalidOperationException("Internal error, can't resolve hash when mpt find");
                         node = newNode;
-                        foreach (var item in Travers(node, path, from))
+                        foreach (var item in Travers(node, path, from, offset))
                             yield return item;
                         break;
                     }
                 case NodeType.BranchNode:
                     {
-                        if (0 < from.Length)
+                        if (offset < from.Length)
                         {
                             for (int i = 0; i < MPTNode.BranchChildCount - 1; i++)
                             {
-                                if (from[0] < i)
-                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
+                                if (from[offset] < i)
+                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), from, from.Length))
                                         yield return item;
-                                else if (i == from[0])
-                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), from[1..]))
+                                else if (i == from[offset])
+                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), from, offset + 1))
                                         yield return item;
                             }
                         }
                         else
                         {
-                            foreach (var item in Travers(node.Children[MPTNode.BranchChildCount - 1], path, Array.Empty<byte>()))
+                            foreach (var item in Travers(node.Children[MPTNode.BranchChildCount - 1], path, from, offset))
                                 yield return item;
                             for (int i = 0; i < MPTNode.BranchChildCount - 1; i++)
                             {
-                                foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
+                                foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), from, offset))
                                     yield return item;
                             }
                         }
@@ -127,11 +130,11 @@ namespace Neo.Plugins.MPT
                     }
                 case NodeType.ExtensionNode:
                     {
-                        if (from.AsSpan().StartsWith(node.Key))
-                            foreach (var item in Travers(node.Next, Concat(path, node.Key), from[node.Key.Length..]))
+                        if (offset < from.Length && from.AsSpan()[offset..].StartsWith(node.Key))
+                            foreach (var item in Travers(node.Next, Concat(path, node.Key), from, offset + node.Key.Length))
                                 yield return item;
-                        else if (0 == from.Length || 0 < node.Key.CompareTo(from))
-                            foreach (var item in Travers(node.Next, Concat(path, node.Key), Array.Empty<byte>()))
+                        else if (from.Length <= offset || 0 < node.Key.CompareTo(from[offset..]))
+                            foreach (var item in Travers(node.Next, Concat(path, node.Key), from, from.Length))
                                 yield return item;
                         break;
                     }
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -101,16 +101,27 @@ namespace Neo.Plugins.MPT
                     }
                 case NodeType.BranchNode:
                     {
-                        for (int i = 0; i < MPTNode.BranchChildCount; i++)
+                        if (0 < from.Length)
                         {
-                            if (i == MPTNode.BranchChildCount - 1 && 0 < from.Length)
-                                break;
-                            if (0 == from.Length || from[0] < i)
-                                foreach (var item in Travers(node.Children[i], i == MPTNode.BranchChildCount - 1 ? path : Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
+                            for (int i = 0; i < MPTNode.BranchChildCount - 1; i++)
+                            {
+                                if (from[0] < i)
+                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
+                                        yield return item;
+                                else if (i == from[0])
+                                    foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), from[1..]))
+                                        yield return item;
+                            }
+                        }
+                        else
+                        {
+                            foreach (var item in Travers(node.Children[MPTNode.BranchChildCount - 1], path, Array.Empty<byte>()))
+                                yield return item;
+                            for (int i = 0; i < MPTNode.BranchChildCount - 1; i++)
+                            {
+                                foreach (var item in Travers(node.Children[i], Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
                                     yield return item;
-                            else if (i == from[0])
-                                foreach (var item in Travers(node.Children[i], i == MPTNode.BranchChildCount - 1 ? path : Concat(path, new byte[] { (byte)i }), from[1..]))
-                                    yield return item;
+                            }
                         }
                         break;
                     }
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -66,9 +66,9 @@ namespace Neo.Plugins.MPT
         {
             var path = ToNibbles(prefix);
             if (from is null) from = Array.Empty<byte>();
-            if (from.Any())
+            if (0 < from.Length)
             {
-                if (from.Length < prefix.Length || !from.AsSpan().StartsWith(prefix))
+                if (!from.AsSpan().StartsWith(prefix))
                     throw new InvalidOperationException("invalid from key");
                 from = ToNibbles(from[prefix.Length..].AsSpan());
             }
@@ -84,7 +84,7 @@ namespace Neo.Plugins.MPT
             {
                 case NodeType.LeafNode:
                     {
-                        if (!from.Any())
+                        if (from.Length == 0)
                             yield return (path, (byte[])node.Value.Clone());
                         break;
                     }
@@ -103,9 +103,9 @@ namespace Neo.Plugins.MPT
                     {
                         for (int i = 0; i < MPTNode.BranchChildCount; i++)
                         {
-                            if (i == MPTNode.BranchChildCount - 1 && from.Any())
+                            if (i == MPTNode.BranchChildCount - 1 && 0 < from.Length)
                                 break;
-                            if (!from.Any() || from[0] < i)
+                            if (0 == from.Length || from[0] < i)
                                 foreach (var item in Travers(node.Children[i], i == MPTNode.BranchChildCount - 1 ? path : Concat(path, new byte[] { (byte)i }), Array.Empty<byte>()))
                                     yield return item;
                             else if (i == from[0])
@@ -119,7 +119,7 @@ namespace Neo.Plugins.MPT
                         if (from.AsSpan().StartsWith(node.Key))
                             foreach (var item in Travers(node.Next, Concat(path, node.Key), from[node.Key.Length..]))
                                 yield return item;
-                        else if (!from.Any() || 0 < node.Key.CompareTo(from))
+                        else if (0 == from.Length || 0 < node.Key.CompareTo(from))
                             foreach (var item in Travers(node.Next, Concat(path, node.Key), Array.Empty<byte>()))
                                 yield return item;
                         break;
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -71,12 +71,12 @@ namespace Neo.Plugins.MPT
             {
                 if (!from.AsSpan().StartsWith(prefix))
                     throw new InvalidOperationException("invalid from key");
-                offset = prefix.Length * 2;
                 from = ToNibbles(from.AsSpan());
             }
             if (path.Length > MPTNode.MaxKeyLength || from.Length > MPTNode.MaxKeyLength)
                 throw new ArgumentException("exceeds limit");
             path = Seek(ref root, path, out MPTNode start).ToArray();
+            offset = path.Length;
             return Travers(start, path, from, offset)
                 .Select(p => (FromNibbles(p.Key).AsSerializable<TKey>(), p.Value.AsSerializable<TValue>()));
         }
--- a/src/StateService/MPT/MPTTrie.Find.cs
+++ b/src/StateService/MPT/MPTTrie.Find.cs
@@ -62,22 +62,30 @@ namespace Neo.Plugins.MPT
             return ReadOnlySpan<byte>.Empty;
         }
 
-        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix)
+        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix, byte[] from = null)
         {
             var path = ToNibbles(prefix);
+            if (from is null) from = Array.Empty<byte>();
+            if (from.Any())
+            {
+                if (from.Length < prefix.Length || !from.AsSpan().StartsWith(prefix))
+                    throw new InvalidOperationException("invalid from key");
+                from = ToNibbles(from[prefix.Length..].AsSpan());
+            }
             path = Seek(ref root, path, out MPTNode start).ToArray();
-            return Travers(start, path)
+            return Travers(start, path, from)
                 .Select(p => (FromNibbles(p.Key).AsSerializable<TKey>(), p.Value.AsSerializable<TValue>()));
         }
 
-        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path)
+        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path, byte[] from)
         {
             if (node is null) yield break;
             switch (node.Type)
             {
                 case NodeType.LeafNode:
                     {
-                        yield return (path, (byte[])node.Value.Clone());
+                        if (!from.Any())
+                            yield return (path, (byte[])node.Value.Clone());
                         break;
                     }
                 case NodeType.Empty:
@@ -87,7 +95,7 @@ namespace Neo.Plugins.MPT
                         var newNode = cache.Resolve(node.Hash);
                         if (newNode is null) throw new InvalidOperationException("Internal error, can't resolve hash when mpt find");
                         node = newNode;
-                        foreach (var item in Travers(node, path))
+                        foreach (var item in Travers(node, path, from))
                             yield return item;
                         break;
                     }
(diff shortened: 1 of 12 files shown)


src/StateService/Storage/StateStore.cs
```diff
@@ -54,11 +54,13 @@ public StateSnapshot GetSnapshot()
             return new StateSnapshot(store);
         }
 
-        public HashSet<byte[]> GetProof(UInt256 root, StorageKey skey)
+        /// <summary>
+        /// DO NOT WRITE
```
shargon, line 58: ?

src/StateService/StatePlugin.cs
```diff
@@ -233,5 +235,69 @@ public JObject GetStateHeight(JArray _params)
...
+                if (!UInt256.TryParse(_params[0].AsString(), out UInt256 root_hash))
+                    throw new RpcException(-100, "Invalid block hash or index");
+                var header = NativeContract.Ledger.GetHeader(System.StoreView, root_hash);
+                if (header is null)
+                    throw new RpcException(-100, "Unkown block");
+                index = header.Index;
+            }
+            var script_hash = UInt160.Parse(_params[1].AsString());
+            var key = Convert.FromBase64String(_params[2].AsString());
+            var isFind = _params.Count > 3 && _params[3].AsBoolean();
```
devhawk, line 260: It seems like this should be two separate Rpc Methods - one for get and one for find - instead of using a parameter to change behavior

src/StateService/StatePlugin.cs
```diff
@@ -233,5 +235,69 @@ public JObject GetStateHeight(JArray _params)
...
+                    JObject j = new();
+                    j["key"] = Convert.ToBase64String(ikey.Key);
+                    j["value"] = Convert.ToBase64String(ivalue.Value);
+                    array.Add(j);
+                }
+                max--;
+            };
+            JObject json = new();
+            json["array"] = array;
+            json["truncated"] = max < 0;
```
devhawk, line 299: There needs to be a mechanism to retrieve records that were truncated in the first request. possibly an optional `page number` parameter which will skip over `max` * `page number` records

devhawk (changes requested):
comments left in GetState

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/Storage/StateStore.cs
ZhangTao1596: Removed.

roman-khimov:
I think returning state without proofs is not a very good thing. And to get a single KV pair we can already use `getstateroot` with `getproof`, so only problem seems to be a set of KV pairs that of course can be served by `getstate` but
 * accepting stateroot hash instead of block index/hash (it's trivial to get via `getstateroot` and it's needed for subsequent checks anyway)
 * without `IsFind` parameter (assuming it to always be true)
 * always returning a set of KV pairs along with boundary proofs (for the first and the last pairs)
This way we could always not just get some data from RPC node, but actually verify it and trust this data correctness.

src/StateService/README.md
```diff
@@ -0,0 +1,60 @@
...
+Value in base64 string
+
+### GetStateheight
+#### Result
+|Name|Type|Summary|
+|-|-|-|
+|localrootindex|number|root hash index calculated locally|
+|validatedrootindex|number|root hash index verified by validators|
+
+### GetState
```
devhawk, line 45: docs need to be updated to match implementation

src/StateService/StatePlugin.cs
```diff
@@ -233,5 +235,81 @@ public JObject GetStateHeight(JArray _params)
...
+        }
+
+        [RpcMethod]
+        public JObject FindState(JArray _params)
+        {
+            PrepareStateParams(_params, out var trie, out var skey);
+            int pageNumber = 0;
+            if (3 < _params.Count)
+                pageNumber = int.Parse(_params[3].AsString());
+            if (Settings.Default.MaxPageNumber <= pageNumber)
```
devhawk, line 288: What is the rationale for specifying a max page number?

src/StateService/StateService/config.json
```diff
@@ -3,7 +3,9 @@
     "Path": "Data_MPT_{0}",
     "FullState": false,
     "Network": 860833102,
-    "AutoVerify": false
+    "AutoVerify": false,
+    "MaxFindResultItems": 100,
+    "MaxPageNumber": 100
```
devhawk, line 8: What happens if there is a find query with over 100x100 results?

devhawk:
> I think returning state without proofs is not a very good thing.

Given that we're not removing the ability to get proofs, what's the concern @roman-khimov? Are you worried about incenting developers to trust remote notes w/o proofs? Instead of a separate method for providing state w/o proof, could we have a method that returns the proof and the state, so that it can be retrieved in one round trip rather than requiring every key to require two round trips (getproof + verifyproof) 

> always returning a set of KV pairs along with boundary proofs (for the first and the last pairs) This way we could always not just get some data from RPC node, but actually verify it and trust this data correctness.

I love this idea. Since the data may span multiple pages, is it sufficient to return the boundary proofs for the first and last pairs *on the page being returned*?

roman-khimov:
> Are you worried about incenting developers to trust remote notes w/o proofs?

Yeah, it's not the end of the world thing, but still we better return things that can be checked when we can.

>  is it sufficient to return the boundary proofs for the first and last pairs on the page being returned?

Yep, that should work, one can reconstruct a part of the MPT from a set of KV pairs and boundary proofs, per-page proofs work just fine for this too.

devhawk:
> Yeah, it's not the end of the world thing, but still we better return things that can be checked when we can.

OK, so let's change the new getstate method to return *both* the proof and the value. I'd rather add a new method and not break the existing getproof method (though I suspect the number of users of that API at this moment is very low)

src/StateService/StateService/config.json
ZhangTao1596: This is the default settings, if you want more, you can change it.

ZhangTao1596:
> * always returning a set of KV pairs along with boundary proofs (for the first and the last pairs)
>   This way we could always not just get some data from RPC node, but actually verify it and trust this data correctness.

If you are suspicious of `getstate` results,  you can use `getproof` to verify. `getstate` is a extension of `getstorage`.

src/StateService/README.md
ZhangTao1596: Will do.

src/StateService/StatePlugin.cs
ZhangTao1596: User can use Empty as prefix and scan all storages, Considering there is no memory protection mechanism in MPT impl, it's dangerous to resolve all branches.

src/StateService/README.md
ZhangTao1596: Updated.

roman-khimov:
> I don't think it's necessary to return both KV and proof, this makes result very large.

You don't need to return proof for every KV, only the first and the last element in a set (page) that you return, it's not that hard and StateService has all the data anyway.

src/StateService/StatePlugin.cs
roman-khimov: IIUC it's contract-bound anyway, although NEO or GAS contracts can of course have a lot of data. At the same time, walking through whole NEO or GAS storage doesn't mean storing whole MPT for them in memory, there is no need to (reminds me of nspcc-dev/neo-go#2019 and the way it throws away from memory things it doesn't need).

But it can be improved even further if we're to add `from` path into request, this way you can start resolving `from` (not iterating whole `prefix` until you end up on a requested page) and then walk the number of elements needed and return them. The next request will have new `from`.

ZhangTao1596:
> only the first and the last element in a set (page) that you return,

Why proofs of the first and last can prove all results reliable?

roman-khimov:
> Why proofs of the first and last can prove all results reliable?

Consider this simplified trie and a request for prefix 2:
s

You trust root hash and you get yellow proofs in return along with all green KV pairs. If you have all green KV pairs (that's important, they need to be sequential and there can't be any omissions in a sequence, but for prefix-based search that's exactly what we have) you can easily reconstruct this part of MPT and compare yellow paths, that will prove all of this subtrie (because you have a hash of 212 in 21 and you get 212 hash by adding 2121, etc).

ZhangTao1596:
> > Why proofs of the first and last can prove all results reliable?
> 
> Consider this simplified trie and a request for prefix 2:
> s
> 
> You trust root hash and you get yellow proofs in return along with all green KV pairs. If you have all green KV pairs (that's important, they need to be sequential and there can't be any omissions in a sequence, but for prefix-based search that's exactly what we have) you can easily reconstruct this part of MPT and compare yellow paths, that will prove all of this subtrie (because you have a hash of 212 in 21 and you get 212 hash by adding 2121, etc).

So user can verify results only using the very first proof and the very last proof cross pages.  We don't need to add proof in every page. We have `getproof`.

src/StateService/StatePlugin.cs
ZhangTao1596: Use `from` is better.

roman-khimov: This technically also allows for user-defined paging (specifying number of results to return).

roman-khimov:
Then consumer of this API will have to retrieve and store all pages before he can verify the result. It's not very productive IMO and some users of this API might as well stop iterating at some point. I'd prefer proving data as it comes (in pages).

src/StateService/StatePlugin.cs
ZhangTao1596: Updated.

ZhangTao1596:
I still don't like both proofs and KV in one request. We have `getproof`. If so, the method should be percisely `FindStatesWithProof`. Should we add proof in `getstate`?

devhawk:
> I still don't like both proofs and KV in one request. We have `getproof`. If so, the method should be percisely `FindStatesWithProof`. Should we add proof in `getstate`?

I'm *personally* fine with having `getstate` not include the proof since we also have `getproof`/`verifyproof`. If we were designing in a greenfield, I'd argue for a single RPC method that returned storage item + proof, but since we have shipped the current version I think adding a `getstate` without proof makes the most sense.

> So user can verify results only using the very first proof and the very last proof cross pages. We don't need to add proof in every page. We have `getproof`

Part of the point is to retrieve information in as few round trips as possible. For `findstate` I'd rather get the proof for the first and last element of each page than require `findstate`, `getproof` and `verifyproof`

devhawk (approved):

src/StateService/StatePlugin.cs
devhawk: This is a nice improvement. Great suggestion @roman-khimov

ZhangTao1596:
@superboyiii Hi bro. Can you test this pr?

devhawk:
@shargon can you give this PR a look?

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -62,22 +62,30 @@ private ReadOnlySpan<byte> Seek(ref MPTNode node, ReadOnlySpan<byte> path, out M
...
-        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path)
+        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path, byte[] from)
         {
             if (node is null) yield break;
             switch (node.Type)
             {
                 case NodeType.LeafNode:
                     {
-                        yield return (path, (byte[])node.Value.Clone());
+                        if (!from.Any())
```
shargon, line 87: Linq again

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -87,23 +95,33 @@ private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] pat
...
                             yield return item;
                         break;
                     }
                 case NodeType.BranchNode:
                     {
                         for (int i = 0; i < MPTNode.BranchChildCount; i++)
                         {
-                            foreach (var item in Travers(node.Children[i], i == MPTNode.BranchChildCount - 1 ? path : Concat(path, new byte[] { (byte)i })))
-                                yield return item;
+                            if (i == MPTNode.BranchChildCount - 1 && from.Any())
```
shargon, line 106: Avoid Linq

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -87,23 +95,33 @@ private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] pat
...
                         break;
                     }
                 case NodeType.ExtensionNode:
                     {
-                        foreach (var item in Travers(node.Next, Concat(path, node.Key)))
-                            yield return item;
+                        if (from.AsSpan().StartsWith(node.Key))
+                            foreach (var item in Travers(node.Next, Concat(path, node.Key), from[node.Key.Length..]))
+                                yield return item;
+                        else if (!from.Any() || 0 < node.Key.CompareTo(from))
```
shargon, line 122: Avoid Linq

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,72 @@ public JObject GetStateHeight(JArray _params)
...
         }
+
+        private ContractState GetHistoricalContractState(MPTTrie<StorageKey, StorageItem> trie, UInt160 script_hash)
+        {
+            const byte prefix = 8;
+            StorageKey skey = new KeyBuilder(NativeContract.ContractManagement.Id, prefix).Add(script_hash);
+            return trie[skey]?.GetInteroperable<ContractState>();
+        }
+
+        private void PrepareStateParams(JArray _params, out MPTTrie<StorageKey, StorageItem> trie, out StorageKey skey)
```
shargon, line 253: I prefer to avoid JArray, and receive the arguments. It will be easy to know what should I send to `FindState`

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,72 @@ public JObject GetStateHeight(JArray _params)
...
+                fkey.Key = Convert.FromBase64String(_params[3].AsString());
+            int count = Settings.Default.MaxFindResultItems;
+            if (4 < _params.Count)
+                count = int.Parse(_params[4].AsString());
+            if (Settings.Default.MaxFindResultItems < count)
+                count = Settings.Default.MaxFindResultItems;
+            JArray array = new();
+            if (count == 0) return array;
+            int i = 0;
+            foreach (var (ikey, ivalue) in trie.Find(skey.ToArray(), fkey.ToArray()))
```
shargon, line 291: Does you tested with a big tree if we could have a DoS here?

shargon:
could you add also https://github.com/neo-project/neo-modules/issues/635 ?

ZhangTao1596:
> could you add also #635 ?

Will do.

src/StateService/StateService/config.json
```diff
@@ -3,7 +3,8 @@
     "Path": "Data_MPT_{0}",
     "FullState": false,
     "Network": 860833102,
-    "AutoVerify": false
+    "AutoVerify": false,
+    "MaxFindResultItems": 100
```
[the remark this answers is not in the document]
ZhangTao1596: We don't use pages anymore. We use `from`.

tests/Neo.Plugins.StateService.Tests/MPT/UT_MPTTrie.cs
```diff
@@ -585,7 +585,6 @@ public void TestBranchPutDirty()
             Assert.AreEqual(7, snapshot.Size);
         }
 
-        [TestMethod]
```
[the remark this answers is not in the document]
ZhangTao1596: Mistake.

src/StateService/StatePlugin.cs
ZhangTao1596: I will. But the values are all contract bound. Besides with `from` we only walk through part of tree.

tests/Neo.Plugins.StateService.Tests/MPT/UT_MPTTrie.cs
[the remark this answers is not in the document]
ZhangTao1596: Updated.

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -62,22 +62,30 @@ private ReadOnlySpan<byte> Seek(ref MPTNode node, ReadOnlySpan<byte> path, out M
             return ReadOnlySpan<byte>.Empty;
         }
 
-        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix)
+        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix, byte[] from = null)
         {
             var path = ToNibbles(prefix);
+            if (from is null) from = Array.Empty<byte>();
+            if (from.Any())
```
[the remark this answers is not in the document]
ZhangTao1596: Done.

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -62,22 +62,30 @@ private ReadOnlySpan<byte> Seek(ref MPTNode node, ReadOnlySpan<byte> path, out M
...
+            {
+                if (from.Length < prefix.Length || !from.AsSpan().StartsWith(prefix))
```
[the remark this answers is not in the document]
ZhangTao1596: Updated.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Updated.

src/StateService/StatePlugin.cs
ZhangTao1596: Updated.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Updated.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Updated.

src/StateService/StateService/config.json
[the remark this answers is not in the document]
shargon: Then we don't need https://github.com/neo-project/neo-modules/pull/638/files#diff-a2e7e45020d834da7043d6854062d505dbac22640129ce2e82803df4cdf59719R12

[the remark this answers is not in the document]
ZhangTao1596: Done.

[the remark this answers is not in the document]
ZhangTao1596: Done.

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -87,23 +95,33 @@ private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] pat
...
                             yield return item;
                         break;
                     }
                 case NodeType.BranchNode:
                     {
                         for (int i = 0; i < MPTNode.BranchChildCount; i++)
                         {
-                            foreach (var item in Travers(node.Children[i], i == MPTNode.BranchChildCount - 1 ? path : Concat(path, new byte[] { (byte)i })))
-                                yield return item;
+                            if (i == MPTNode.BranchChildCount - 1 && 0 < from.Length)
```
shargon, line 106: The condition `0 < from.Length` could be outside the `for`

ZhangTao1596: Done. Please have a second review.

shargon: @superboyiii please test it

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,81 @@ public JObject GetStateHeight(JArray _params)
...
+            byte[] key = Array.Empty<byte>();
+            if (3 < _params.Count)
+                key = Convert.FromBase64String(_params[3].AsString());
+            int count = Settings.Default.MaxFindResultItems;
+            if (4 < _params.Count)
+                count = int.Parse(_params[4].AsString());
+            if (Settings.Default.MaxFindResultItems < count)
+                count = Settings.Default.MaxFindResultItems;
+            if (!Settings.Default.FullState && StateStore.Singleton.CurrentLocalRootHash != root_hash)
+                throw new RpcException(-100, "Old state not supported");
```
shargon, line 268: Move this condition to line 257?

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,81 @@ public JObject GetStateHeight(JArray _params)
...
+        }
+
+        [RpcMethod]
+        public JObject GetState(JArray _params)
+        {
+            var root_hash = UInt256.Parse(_params[0].AsString());
+            var script_hash = UInt160.Parse(_params[1].AsString());
+            var key = Convert.FromBase64String(_params[2].AsString());
+            if (!Settings.Default.FullState && StateStore.Singleton.CurrentLocalRootHash != root_hash)
+                throw new RpcException(-100, "Old state not supported");
```
shargon, line 308: Move after parse `root_hash`?

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -62,22 +62,30 @@ private ReadOnlySpan<byte> Seek(ref MPTNode node, ReadOnlySpan<byte> path, out M
...
+                from = ToNibbles(from[prefix.Length..].AsSpan());
+            }
             path = Seek(ref root, path, out MPTNode start).ToArray();
-            return Travers(start, path)
+            return Travers(start, path, from)
                 .Select(p => (FromNibbles(p.Key).AsSerializable<TKey>(), p.Value.AsSerializable<TValue>()));
         }
 
-        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path)
+        private IEnumerable<(byte[] Key, byte[] Value)> Travers(MPTNode node, byte[] path, byte[] from)
```
shargon, line 80: Maybe we can receive a `CancellationToken` and break the iteration if the connection died?

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,81 @@ public JObject GetStateHeight(JArray _params)
...
+                if (i < count)
+                {
+                    JObject j = new();
+                    j["key"] = Convert.ToBase64String(ikey.Key);
+                    j["value"] = Convert.ToBase64String(ivalue.Value);
+                    array.Add(j);
+                }
+                i++;
+            };
+            return array;
```
devhawk, line 298: In addition to the data, you need to return some indicator if there are more items to be found

devhawk (changes requested):
FindState RPC method needs a way of indicating if we are at the end of the array of keys or if there are more to retrieve

src/StateService/MPT/MPTTrie.Find.cs
```diff
@@ -62,22 +62,30 @@ private ReadOnlySpan<byte> Seek(ref MPTNode node, ReadOnlySpan<byte> path, out M
             return ReadOnlySpan<byte>.Empty;
         }
 
-        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix)
+        public IEnumerable<(TKey Key, TValue Value)> Find(ReadOnlySpan<byte> prefix, byte[] from = null)
```
devhawk, line 65: If from key is specified, it should not be included in the results returned by Find.

devhawk:
FYI, I set up a local neo-cli node for testnet running StateService compiled out of this branch. Worked like a charm the first time out!

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/StatePlugin.cs
ZhangTao1596: Done.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Currently rpc method can't get `HttpContext`. I'm afraid this change will affect all rpc method.
Maybe I can create another pr.
Do you have any suggestion about get http context here without influence on other rpc methods?

src/StateService/MPT/MPTTrie.Find.cs
superboyiii: > @superboyiii please test it

Works well.

src/StateService/StatePlugin.cs
superboyiii: > Did you tested with a big tree and result if we could have a DoS here?

100 items costs 5ms, 1000 items costs 25ms from my side. I think it's OK, not very cost time, much better than getNep17transfers.

superboyiii:
Tested. Pass
Works well from my side.

src/StateService/MPT/MPTTrie.Find.cs
devhawk: Note, this is still an issue as of latest commit in this PR (661284c08907608a93c7331fd90084de89af27a1)

src/StateService/MPT/MPTTrie.Find.cs
devhawk: We should do this in a separate PR to address #629 generally, not just for state service

devhawk:
Did @roman-khimov sign off on *not* including boundary proofs in FindStates?

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: > We should do this in a separate PR to address #629 generally, not just for state service

Agree.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: I have not finished this fix. Will inform you then.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Done.

src/StateService/MPT/MPTTrie.Find.cs
ZhangTao1596: Done.

ZhangTao1596:
> Did @roman-khimov sign off on _not_ including boundary proofs in FindStates?

Proof added.

ZhangTao1596:
Merge?

devhawk (dismissed):

shargon (approved):

devhawk:
> Merge?

Yes please

roman-khimov:
We can return an empty array of KV pairs and proof of nonexistence (trie path up to the point where it can't go further because of missing element).

ZhangTao1596:
> Actually, one thing. If you call `FindStates` with an unused contract hash (like `UInt160.Zero`) you get an error `The given key was not present in the dictionary`. I was expecting an empty array of states. Throwing exceptions when there is no data doesn't seem like the right choice for state service RPC endpoints to me, but I'm interested in what @shargon and @roman-khimov think

It should be `Unkown contract`.

ZhangTao1596:
> We can return an empty array of KV pairs and proof of nonexistence (trie path up to the point where it can't go further because of missing element).

We can't. If the contract is not deployed, There is no contract id. If we want to prove, we can prove the contract inexistent in contract management contract.

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,91 @@ public JObject GetStateHeight(JArray _params)
...
+                throw new RpcException(-100, "Old state not supported");
+            var script_hash = UInt160.Parse(_params[1].AsString());
+            var prefix = Convert.FromBase64String(_params[2].AsString());
+            byte[] key = Array.Empty<byte>();
+            if (3 < _params.Count)
+                key = Convert.FromBase64String(_params[3].AsString());
+            int count = Settings.Default.MaxFindResultItems;
+            if (4 < _params.Count)
+                count = int.Parse(_params[4].AsString());
+            if (Settings.Default.MaxFindResultItems < count)
```
vncoelho, line 267: Maybe it should be returned as an exception because it could mislead the interpretation.

devhawk (changes requested):
~How come ledger contract data is not included in state service? ~

Never mind

roman-khimov:
Primary reasons for that are described in #520. But we have `getblock` and `getrawtransaction` for ledger things.

devhawk (approved):

src/StateService/StatePlugin.cs
ZhangTao1596: How? Can you describe this in detail? We have `truncated` in response and the results are ordered by key.

src/StateService/StatePlugin.cs
```diff
@@ -239,5 +242,91 @@ public JObject GetStateHeight(JArray _params)
...
+        }
+
+        [RpcMethod]
+        public JObject FindStates(JArray _params)
+        {
+            var root_hash = UInt256.Parse(_params[0].AsString());
+            if (!Settings.Default.FullState && StateStore.Singleton.CurrentLocalRootHash != root_hash)
+                throw new RpcException(-100, "Old state not supported");
+            var script_hash = UInt160.Parse(_params[1].AsString());
+            var prefix = Convert.FromBase64String(_params[2].AsString());
```
shargon, line 260: We should check the max length of this argument, ensure that is less than `MaxKeyLength`

ZhangTao1596: Done.

shargon (approved):

superboyiii:
Tested Pass.

superboyiii (approved):

superboyiii:
@shargon Could you review again?

superboyiii:
@shargon Merge?

shargon (approved):

Issue #636, opened 2021-08-27
Provide protocol configuration data via RPC


roman-khimov:
**Summary or problem description**
Clients may need some addition protocol configuration data to adjust their behaviour wrt the network. A typical thing is `MillisecondsPerBlock` value that could be used to adjust some timeouts/polling intervals. But `MaxValidUntilBlockIncrement` can also be useful to construct proper transactions as well as `MaxTraceableBlocks` to set some assumptions on data availability.

**Do you have any solution you want to propose?**
We can either create an additional API for that or just stuff this data into `getversion` (like it was done for #358).

**Where in the software does this update applies to?**
- RPC (HTTP)

devhawk:
Also need to include Address Version in the list of RPC available protocol config

PR #639, opened 2021-08-30
Add protocol values to getversion

shargon:
Close https://github.com/neo-project/neo-modules/issues/636

### DIFF
--- a/src/RpcClient/Models/RpcVersion.cs
+++ b/src/RpcClient/Models/RpcVersion.cs
@@ -11,6 +11,9 @@ namespace Neo.Network.RPC.Models
             public uint MaxValidUntilBlockIncrement { get; set; }
             public uint MaxTraceableBlocks { get; set; }
             public byte AddressVersion { get; set; }
+            public uint MaxTransactionsPerBlock { get; set; }
+            public int MemoryPoolMaxTransactions { get; set; }
+            public ulong InitialGasDistribution { get; set; }
 
             public JObject ToJson()
             {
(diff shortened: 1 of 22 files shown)


cloud8little:
seems Address Version should also be included, @shargon

besides, there are other parameters, maybe they are also needed. @devhawk what do you think?

1. MaxTransactionsPerBlock
2. MemoryPoolMaxTransactions
3. InitialGasDistribution

shargon:
If we decide which parameters should be added, we can do it now :)

devhawk:
> besides, there are other parameters, maybe they are also needed. @devhawk what do you think?

It would seem reasonable that `getversion` returns all of these values. I'm not sure how often they would be used, but I don't think there's much in the way of additional incremental cost to include them

cloud8little:
> > besides, there are other parameters, maybe they are also needed. @devhawk what do you think?
> 
> It would seem reasonable that `getversion` returns all of these values. I'm not sure how often they would be used, but I don't think there's much in the way of additional incremental cost to include them

Alright, let's add them @shargon

src/RpcClient/Models/RpcVersion.cs
```diff
@@ -14,27 +44,28 @@ public class RpcVersion
 
         public string UserAgent { get; set; }
 
+        public RpcProtocol Protocol { get; set; } = new();
 
         public JObject ToJson()
         {
             JObject json = new();
-            json["network"] = Network;
```
roman-khimov, line 21: This will break compatibility, do we really want that?

shargon: We can create `getprotocol`

roman-khimov: I'd rather just return `network` here and in `protocol` subsection, make a notice that top-level `network` is deprecated and will be removed and then remove it in some subsequent release.

src/RpcServer/RpcServer.Node.cs
```diff
@@ -62,7 +62,15 @@ protected virtual JObject GetVersion(JArray _params)
             json["wsport"] = localNode.ListenerWsPort;
             json["nonce"] = LocalNode.Nonce;
             json["useragent"] = LocalNode.UserAgent;
-            json["network"] = system.Settings.Network;
+
```
cloud8little, line 65:
```suggestion
json["protocol"] = new JObject();
```

cloud8little (approved):

cloud8little:
Test passed. wait for @erikzhang review.

erikzhang (approved):

Issue #542, opened 2021-03-09
NEP-11 tracker


roman-khimov:
**Summary or problem description**
I think we need an RPC interface for NEP-11 data, just like NEP-17 tracker but adapted to NEP-11 needs.

**Do you have any solution you want to propose?**
`getnep11transfers` and `getnep11balances` RPC calls with NEP-11 data. Transfers are kinda obvious, they are similar to NEP-17, just need `tokenId` to be added. Balances are more involved, we need to replace `amount` with `tokens` int the result to get something like this:
```
{
   "balance" : [
      {
         "lastupdatedblock" : 251604,
         "assethash" : "a48b6e1291ba24211ad11bb90ae2a10bf1fcd5a8",
         "tokens" : [
             {
                 "id": "A",
                 "amount": 1
             },
             {
                 "id": "B",
                 "amount": 1
             }
         ]
      }
   ],
   "address" : "N..."
}
```

**Where in the software does this update applies to?**
- Plugins
- RPC (HTTP)

PR #651, opened 2021-10-09
Add nep11 tracker

Ashuaidehao:
Close https://github.com/neo-project/neo-modules/issues/542
Add nep11 tracker.

### DIFF
--- a/src/RpcNep11Tracker/RpcNep11Tracker.cs
+++ b/src/RpcNep11Tracker/RpcNep11Tracker.cs
@@ -110,13 +110,14 @@ namespace RpcNep11Tracker
                 if (!contracts.ContainsKey(transferRecord.asset))
                 {
                     var state = NativeContract.ContractManagement.GetContract(snapshot, transferRecord.asset);
-                    var balanceMethod = state.Manifest.Abi.Methods.FirstOrDefault(m => m.Name == "balanceOf");
-                    if (balanceMethod == null)
+                    var balanceMethod = state.Manifest.Abi.GetMethod("balanceOf", 1);
+                    var balanceMethod2 = state.Manifest.Abi.GetMethod("balanceOf", 2);
+                    if (balanceMethod == null && balanceMethod2 == null)
                     {
                         Log($"{state.Hash} is not nft!", LogLevel.Warning);
                         continue;
                     }
-                    var isDivisible = balanceMethod.Parameters.Length == 2;
+                    var isDivisible = balanceMethod2 != null;
                     contracts[transferRecord.asset] = (isDivisible, state);
                 }
                 var asset = contracts[transferRecord.asset];
--- a/src/RpcNep11Tracker/RpcNep11Tracker.cs
+++ b/src/RpcNep11Tracker/RpcNep11Tracker.cs
@@ -171,7 +171,6 @@ namespace RpcNep11Tracker
         }
 
 
-
         private void HandleNotification(DataCache snapshot, IVerifiable scriptContainer, UInt160 scriptHash, string eventName, VmArray stateItems, List<TransferRecord> transfers, ref ushort transferIndex)
         {
             if (eventName != "Transfer") return;
@@ -199,8 +198,8 @@ namespace RpcNep11Tracker
                 return;
             if (fromBytes == null && toBytes == null) return;
 
-            var from = UInt160.Zero;
-            var to = UInt160.Zero;
+            var from = fromBytes == null ? UInt160.Zero : new UInt160(fromBytes);
+            var to = toBytes == null ? UInt160.Zero : new UInt160(toBytes);
             var tokenId = tokenIdItem as ByteString;
 
             transfers.Add(new TransferRecord(scriptHash, from, to, tokenId, amountItem.GetInteger()));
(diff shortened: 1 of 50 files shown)


superboyiii:
Test **Pass**
From my side, it solved current requirements of showing Nep11 tx history and current NFTs in specific address with group by contract hash. But still need @roman-khimov and @shargon review again.

superboyiii:
> @Ashuaidehao, the consensus and nodes stopped syncing when testing this PR.

Could you tell us more details? I can't replay your issue.

vncoelho:
@superboyiii,

testing the following config:
```bash
# https://github.com/neo-project/neo-node/commits/master
NEO_CLI_BLOCKCHAIN_URL=https://github.com/neo-project/neo-node.git
NEO_CLI_BRANCH=master
NEO_CLI_COMMIT=8153553dcd8d68c752b9988ce7d7b9e360529b65

# https://github.com/neo-project/neo/commits/master
NEO_BLOCKCHAIN_URL=https://github.com/Qiao-Jin/neo.git
NEO_BLOCKCHAIN_BRANCH=taskmanager_optimization
NEO_BLOCKCHAIN_COMMIT=456fe0242001eec2ce92b44fbba84c5d5e9393b1

# https://github.com/neo-project/neo-modules/commits/master
NEO_PLUGINS_URL=https://github.com/neo-project/neo-modules.git
NEO_PLUGINS_BRANCH=master
NEO_PLUGINS_COMMIT=d0f143ac43fe6d4496430c01f737f3dd9f232904
```

Worked.
Now, when changing to:

```sh
# https://github.com/neo-project/neo-modules/commits/master
NEO_PLUGINS_URL=https://github.com/Ashuaidehao/neo-modules.git
NEO_PLUGINS_BRANCH=add-nep11
NEO_PLUGINS_COMMIT=7556a43
```

It stopped synced and stuck on changeview.


In fact, it looks like it is because the *.dll that have dependency on RpcServer can not be on the folder if the node does not have RpcServer enabled.
Since two out of four of our CNs do not have RpcServer, I noticed that I had to also delete it before starting the node:
```sh
if [[ ${IS_RPC_SERVER} = "0" ]]; then
	echo "Deleting RPCServer"
	rm /opt/node/neo-cli/Plugins/RpcServer.dll
	rm /opt/node/neo-cli/Plugins/RpcNep17Tracker.dll
	rm /opt/node/neo-cli/Plugins/RpcNep11Tracker.dll
	rm /opt/node/neo-cli/Plugins/ApplicationLogs.dll
	rm /opt/node/neo-cli/Plugins/StateService.dll	
	sleep 1
fi
```

In this sense, the problem is more generic than this PR,it is a result of not having the `RpcServer`. 
Perhaps the *.dll could be in the folder and we could verify if the `config` file is created or node, something like that.
Anyway, the problem was on our side here by not deleting the * .dll for some nodes.

superboyiii:
> In this sense, the problem is more generic than this PR,it is a result of not having the `RpcServer`. Perhaps the *.dll could be in the folder and we could verify if the `config` file is created or node, something like that. Anyway, the problem was on our side here by not deleting the * .dll for some nodes.

It's an issue about depedency check. Since all RpcServer required plugins have this issue, we could make it into another PR to check dependency before onload.

superboyiii (dismissed):

src/RpcNep11Tracker/RpcNep11Tracker.cs
```diff
@@ -0,0 +1,428 @@
...
+            using ScriptBuilder sb = new ScriptBuilder();
+            sb.EmitDynamicCall(record.asset, "balanceOf", record.from, record.tokenId);
+            sb.EmitDynamicCall(record.asset, "balanceOf", record.to, record.tokenId);
+            using ApplicationEngine engine = ApplicationEngine.Run(sb.ToArray(), snapshot, settings: neoSystem.Settings);
+            if (engine.State.HasFlag(VMState.FAULT) || engine.ResultStack.Count != 2)
+            {
+                Console.WriteLine($"Fault:{record.from},{record.to},{record.tokenId.GetSpan().ToHexString()}");
+                return;
+            }
+            Put(Nep11BalancePrefix, new Nep11BalanceKey(record.to, record.asset, record.tokenId), new Nep11Balance() { Balance = engine.ResultStack.Pop().GetInteger(), LastUpdatedBlock = _currentHeight });
```
shargon, line 159: Is not easy to deny the service if the contract is prepared to return a wrong result? it will deny the service to any node that will have this plugin, I think that it should log the error but not deny the service, imagine if a consensus node has this plugin.

https://github.com/neo-project/neo-vm/blob/b48113f8dc787d059324c2d34c855f77b0e3645d/src/neo-vm/Types/StackItem.cs#L116

shargon (changes requested):

src/RpcNep11Tracker/RpcNep11Tracker.cs
Ashuaidehao: Please check https://github.com/neo-project/neo-modules/pull/671

superboyiii:
I will close this since it's replaced by https://github.com/neo-project/neo-modules/pull/671

Issue #637, opened 2021-08-27
BLS Consensus Structure


Qiao-Jin:
**Summary or problem description**

Since we decide to import on-chain random number into Neo system, there has been a lot of work on this topic such as https://github.com/neo-project/neo/issues/1657, https://github.com/neo-project/neo/issues/2456, https://github.com/neo-project/neo/issues/2019, https://github.com/neo-project/neo/pull/2477, https://github.com/neo-project/neo/pull/2481, https://github.com/neo-project/neo/pull/2476, https://github.com/neo-project/neo/pull/2470, https://github.com/neo-project/neo-modules/pull/590, 
https://github.com/neo-project/neo-modules/pull/596, etc. 

However, currently the random number function in use is imported in https://github.com/neo-project/neo-modules/pull/596, which is simply generating pseudo random number by a C# random number generator. We know this is only a template scheme and need to be replaced soon or later.

&nbsp;

**Do you have any solution you want to propose?**

One choice is BLS random number which is introduced in https://github.com/neo-project/neo/issues/1657#issue-621474006. In this solution, BLS signature upon a certain message (say, prevHash) of multiple nodes are adopted as seeds to create a nonce as random number, which cannot be decided by any single node or nodes less than certain amount (m). On the other hand, signatures of any m nodes can be used to calculate the unique nonce.

There are 2 tasks in this option:
1. Is there available BLS algorithm?
2. How to distribute BLS shared private keys?

For the first one, we have already developed BLS algorithm in this repo: https://github.com/Qiao-Jin/BLSTest.

For the second one, we would like to introduce some updates to current consensus algorithm as follows:

1. Pre-consensus steps, **WHICH IS ONLY EXECUTED AT EPOCH START IF THE LIST OF VALIDATOR CHANGES IN NEW EPOCH**
![image](https://user-images.githubusercontent.com/43407364/131126405-5988ed63-a408-479f-8531-6c62875f333a.png)

(1) Each validator should broadcast a consensus payload, BLSSecretKey, which includes 2 sections:

a. Shared BLS private keys to other validators, which is encrypted by receiver's ECC public keys respectively (this feature will be imported in another issue),
b. Corresponding BLS public keys which can be used to check BLS private keys & BLS signatures.

Meanwhile each validator would also collect & check BLSSecretKey payloads from others and keep a local trusted validator list.

[...]

PR #601, opened 2021-06-20
Anti-MEV attack

Liaojinghui:
This pr tries to mitigate the MEV attack.

PR #670, opened 2021-10-31
Runtime secure random number for N3

Liaojinghui:
This pr implement secure random number for N3 at runtime.

#601 
#637 
#596 
https://github.com/neo-project/neo/issues/2504


Liaojinghui:
`our solution`
image

`Other solution`:
image

roman-khimov:
Transaction list exchange will have serious performance consequences. And the worst thing is that it would affect the performance most exactly when we need it most, under some stress, because transaction pressure tends to lead to a greater deal of unsynchrony between the nodes. Notice also that for high-performing networks (which inevitably have to allow tens of thousands of transactions in a block) that could easily mean exchanging 1MB+ messages in this phase further complicating the situation.

I'd suggest trying this in neo-bench, four CNs and especially 7+2 scenario. 7+2 was added quite recently and it makes all nodes look bad because the network becomes highly unsynchronized, but that's exactly what makes this case more realistic and interesting (we've not optimized for it yet).

Liaojinghui:
@roman-khimov Thank you very much for your suggestion roman, that would be a great help.

PR #673, opened 2021-11-13
.NET 6.0

erikzhang:
Related to neo-project/neo#2625
Fix https://github.com/neo-project/neo-modules/issues/674

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  DOTNET_VERSION: 5.0.100
+  DOTNET_VERSION: 6.0.x
 
 jobs:
 
@@ -20,9 +20,7 @@ jobs:
       with:
         dotnet-version: ${{ env.DOTNET_VERSION }}
     - name: Check format
-      run: |
-        dotnet tool install --version 5.0.142902 --tool-path ./ dotnet-format --add-source https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json
-        ./dotnet-format . --check -v diagnostic
+      run: dotnet format --verify-no-changes --verbosity diagnostic
     - name: Test
       run: |
         sudo apt-get --assume-yes install libleveldb-dev libsnappy-dev libc6-dev
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -31,7 +31,8 @@ namespace Neo.Plugins
 
         private static readonly HttpClient httpClient = new()
         {
-            Timeout = TimeSpan.FromSeconds(5)
+            Timeout = TimeSpan.FromSeconds(5),
+            MaxResponseContentBufferSize = ushort.MaxValue
         };
 
         private Wallet wallet;
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -17,9 +17,8 @@ using Neo.Wallets;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
-using System.Net;
+using System.Net.Http;
 using System.Text;
 using System.Threading;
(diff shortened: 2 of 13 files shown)


src/OracleService/OracleService.cs
```diff
@@ -225,18 +229,8 @@ private static async Task SendContentAsync(Uri url, string content)
...
-                HttpWebRequest request = (HttpWebRequest)WebRequest.Create(url);
-                request.Method = "POST";
-                request.ContentType = "application/json";
-                request.Timeout = 5000;
-                using (StreamWriter dataStream = new StreamWriter(await request.GetRequestStreamAsync()))
-                {
-                    await dataStream.WriteAsync(content);
-                }
-                HttpWebResponse response = (HttpWebResponse)await request.GetResponseAsync();
-                if (response.ContentLength > ushort.MaxValue) throw new Exception("The response it's bigger than allowed");
```
shargon, line 237: We will lose this condition?

erikzhang: But we don’t care about response at all, do we?

shargon: Is the download time included in the timeout ?

shargon (approved):

src/OracleService/OracleService.cs
```diff
@@ -30,6 +29,11 @@ public class OracleService : Plugin, IPersistencePlugin
     {
         private const int RefreshIntervalMilliSeconds = 1000 * 60 * 3;
 
+        private static readonly HttpClient httpClient = new()
+        {
+            Timeout = TimeSpan.FromSeconds(5)
```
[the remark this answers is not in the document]
erikzhang: Can you be sure that all responses are short than `ushort.MaxValue`?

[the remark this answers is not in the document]
shargon: We should set a max response size, maybe 1mb

.github/workflows/main.yml
```diff
@@ -6,7 +6,7 @@ on:
   pull_request:
 
 env:
-  DOTNET_VERSION: 5.0.100
+  DOTNET_VERSION: 6.0.x
```
superboyiii, line 9: Why not?
```suggestion
  DOTNET_VERSION: 6.0.100
```

superboyiii (approved):

erikzhang:
Wait for neo-node.

erikzhang:
Why `NeoFS.API` uses CI version of packages?

erikzhang (commented):
Why NeoFS.API uses CI version of packages?

src/OracleService/OracleService.cs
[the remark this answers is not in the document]
erikzhang: Maybe we need some tests?

ZhangTao1596:
> Why NeoFS.API uses CI version of packages?

There will be problems if `NeoFS.API` use `v3.0.3` and neo-modules use `v3.0.3-CI01319`. So I updated it.

src/OracleService/OracleService.csproj
```diff
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Neo.ConsoleService" Version="1.0.0" />
+    <PackageReference Include="Neo.ConsoleService" Version="1.2.0" />
     <PackageReference Include="NeoFS.API" Version="3.0.3" />
```
ZhangTao1596, line 9:
```suggestion
    <PackageReference Include="NeoFS.API" Version="3.1.0" />
```

It has been updated to `v3.1.0`.

erikzhang:
> There will be problems if NeoFS.API uses v3.0.3 and neo-modules uses v3.0.3-CI01319. So I updated it.

We should never use CI packages in master branch.

vncoelho (commented):
For now it is all working for NeoCompiler Eco.

shargon (approved):

superboyiii (approved):

Liaojinghui (approved):

PR #683, opened 2022-01-17
add DKG for BLS

Liaojinghui:
Thanks to @Qiao-Jin @doubiliu. This is the DKG part of the implementation of BLS random numbers. BLS is implemented in #670.


src/DBFTPlugin/DKG/Utility.cs
```diff
@@ -0,0 +1,145 @@
...
+            result[i].sign = i % 2 == 0;
+            result[i].id = input[i] - 1;
+        }
+
+        return result;
+    }
+
+    public static int GetZeroBits(uint n, uint m)
+    {
+        return (int)(Math.Log2(n * ((long)Math.Pow(n, m) - 1) / (n - 1) + 1) + 3);
```
nicolegys, line 97: If `n=1`, it will divide by zero. We need to support solo mode, although it doesn't fit the theme.

vncoelho: Good point @nicolegys

Liaojinghui:
DKG is inefficient to update, working on adopting `Dynamic-Committee Proactive Secret Sharing`.

PR #684, opened 2022-01-21
decouple the message name and the message type

Liaojinghui:
The name of the consensus message type is **heavily** coupled with its corresponding message name, and it is 
### **incredibly**  **horribly**  **desperately** **hard** 
to debug when errors happen.

When problem happens here, 

No exception
No log
No comment
No way to locate the error, cant even by monitoring the network traffic.

problem addressed with help from @nicolegys


src/DBFTPlugin/Messages/ConsensusMessage.cs
```diff
@@ -37,7 +37,7 @@ public static ConsensusMessage DeserializeFrom(byte[] data)
         {
             ConsensusMessageType type = (ConsensusMessageType)data[0];
             Type t = typeof(ConsensusMessage);
-            t = t.Assembly.GetType($"{t.Namespace}.{type}", false);
+            t = t.Assembly.GetType($"{t.Namespace}.{type.ToMessage()}", false);
             if (t is null) throw new FormatException();
```
shargon, line 41: It should throw this exception if it was not found, could detail the error?

Liaojinghui: This pr https://github.com/neo-project/neo-modules/pull/683

I added two new message types to do the DKG. The message name is DKGShareMessage while the type is DKGShare, the message could be sent, but the node can only receive one response message.


image

shargon: Please, try with this https://github.com/neo-project/neo-modules/pull/685

Liaojinghui: Using enumerated type name to represent the class name already break the basic rules of software pattern. What I learned from the painful debug experience in the last few days is that the current implementation is misleading.

PR #685, opened 2022-01-24
Consensus: Add type check

shargon:


src/DBFTPlugin/Messages/ConsensusMessage.cs
```diff
@@ -36,6 +36,8 @@ public virtual void Deserialize(BinaryReader reader)
         public static ConsensusMessage DeserializeFrom(byte[] data)
         {
             ConsensusMessageType type = (ConsensusMessageType)data[0];
+            if (!Enum.IsDefined(typeof(ConsensusMessageType), type))
+                throw new ArgumentOutOfRangeException(nameof(type));
             Type t = typeof(ConsensusMessage);
             t = t.Assembly.GetType($"{t.Namespace}.{type}", false);
             if (t is null) throw new FormatException();
```
erikzhang, line 43: If we don't modify it, then it will throw an exception here, right?

nicolegys: > If we don't modify it, then it will throw an exception here, right?

Yes, t is null and throw an exception. Maybe his point is to make the exception information more clear. :joy:

shargon: I think so, but https://github.com/neo-project/neo-modules/pull/684#discussion_r790356491

Liaojinghui: I don't know, but i did not see any sign of any exception here.

image

Liaojinghui: Here in `ConsensusService`:

```C#
   ConsensusMessage message;
            try
            {
                message = context.GetMessage(payload);
            }
            catch (FormatException)
            {
                return;
            }
            catch (IOException)
            {
                return;
            }
```

The exception is ignored directly.

shargon: I think that it should be logged, but type is working as a switch

PR #687, opened 2022-01-27
Rework RpcServer.ProcessAsync to be non-terminal ASP.NET Middleware

devhawk:
This will allow neo express to use RpcServer in an HTTP pipeline that RpcServer doesn't create

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -145,7 +145,7 @@ namespace Neo.Plugins
             this.settings = settings;
         }
 
-        private async Task ProcessAsync(HttpContext context)
+        public async Task ProcessAsync(HttpContext context)
         {
             context.Response.Headers["Access-Control-Allow-Origin"] = "*";
             context.Response.Headers["Access-Control-Allow-Methods"] = "GET, POST";
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -119,7 +119,7 @@ namespace Neo.Plugins
             .Configure(app =>
             {
                 app.UseResponseCompression();
-                app.Use(next => context => ProcessAsync(context, next));
+                app.Use(ProcessAsync);
                 app.Run(InvalidRequestAsync);
             })
             .ConfigureServices(services =>
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -119,7 +119,8 @@ namespace Neo.Plugins
             .Configure(app =>
             {
                 app.UseResponseCompression();
-                app.Run(ProcessAsync);
+                app.Use(next => context => ProcessAsync(context, next));
+                app.Run(InvalidRequestAsync);
             })
             .ConfigureServices(services =>
             {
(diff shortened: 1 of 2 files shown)


shargon:
protected and extend the class?

devhawk:
> protected and extend the class?

If we're going to do this, we should provide a mechanism to inject steps into the http pipeline that StartRpcServer creates. Maybe we should separate out the pipeline host code from the rest of RpcServer and make RpcServer a [standard ASP.NET Middleware class](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-6.0)?  Should be pretty straightfoward since RpcServer already plugs into the http pipeline via app.run. WOuldn't take much to make it work with App.use

erikzhang:
> If we're going to do this, we should provide a mechanism to inject steps into the http pipeline that StartRpcServer creates. Maybe we should separate out the pipeline host code from the rest of RpcServer and make RpcServer a standard ASP.NET Middleware class? Should be pretty straightfoward since RpcServer already plugs into the http pipeline via app.run. WOuldn't take much to make it work with App.use

It sounds good.

devhawk:
> > If we're going to do this, we should provide a mechanism to inject steps into the http pipeline that StartRpcServer creates. Maybe we should separate out the pipeline host code from the rest of RpcServer and make RpcServer a standard ASP.NET Middleware class? Should be pretty straightfoward since RpcServer already plugs into the http pipeline via app.run. WOuldn't take much to make it work with App.use
> 
> It sounds good.

I'll work on this later this week.

devhawk:
Ping @erikzhang @shargon

erikzhang:
> modified ProcessRequest to return null for invalid requests (missing id/method/params or non-array params) instead of returning JsonRpc Error response

I don't understand why it should be modified like this.

devhawk:
FYI, with 3.2 release on the horizon, I will get this PR updated shortly

devhawk:
> > modified ProcessRequest to return null for invalid requests (missing id/method/params or non-array params) instead of returning JsonRpc Error response
> 
> I don't understand why it should be modified like this.

So that `ProcessAsync` can tell if the request should be passed thru to the `next` delegate in the ASP.NET middleware pipeline. This let's NeoExpress create its own kestrel service that also composes RpcServer. Express code will look something like this:

``` csharp
.Configure(app =>
{
    app.UseResponseCompression();
    app.Use(next => context => rpcServer.ProcessAsync(context, next));
    app.Run(NeoExpressFunctionality);
})
```

This way, rpcServer always gets first crack at the request so that anything it considers worthy of a response is processed immediately and never gets passed down the pipeline.

devhawk:
One thing I just realized about this PR is how it handles multiple JSON RPC requests in an array. Currently, it simply filters out null responses from `ProcessRequest`. I don't think that's the right choice. Instead, if ANY request in the array gets a valid response, then any invalid responses should get an error return not get filtered out. Will update

devhawk:
cc @erikzhang @shargon I am trying to get this merged for 3.2

src/RpcServer/RpcServer.cs
```diff
@@ -119,7 +119,8 @@ public void StartRpcServer()
             .Configure(app =>
             {
                 app.UseResponseCompression();
-                app.Run(ProcessAsync);
+                app.Use(ProcessAsync);
+                app.Run(InvalidRequestAsync);
```
erikzhang, line 123: I think we should use `InvalidRequestAsync` first and then run `ProcessAsync` as the terminal middleware. Then `InvalidRequestAsync` can catch the exceptions in `ProcessAsync`.

devhawk: If we're going to make `ProcessAsync` the terminal middleware, then I think we should simply go back to [the original commit in this PR](https://github.com/neo-project/neo-modules/pull/687/commits/bb4cccad4c49502ed5f776d1b5493e9fa956f2b9) and make `ProcessAsync` public. 

I split `InvalidRequestAsync` out from `ProcessAsync` so Express could insert its middleware in the ASP.NET Pipeline after `ProcessAsync`. If `ProcessAsync` is terminal middleware, then Express has no choice but to put its middleware before `ProcessAsync`. Compatibility with `RpcServer` is a high priority for Express, so I wanted to give `ProcessAsync` the highest priority position in the ASP.NET pipeline. But that doesn't work if `ProcessAsync` can only be used as terminal middleware.

Honestly, I don't care which approach we take. Simply making `ProcessAsync` public is smaller and has no logic changes. Changing `ProcessAsync` to be non-terminal means for less code for Express. Either approach works for me.

devhawk:
I've renamed this PR to reflect the code changes better. I also opened a new PR #701 that only makes the visibility change. 

As per my last comment, either approach is fine by me. But I need either this PR or #701 to be accepted in order to enable the new features we are adding to Express.

PR #688, opened 2022-02-06
dbft: tune MaxBlock* parameters

roman-khimov:
Follow ExecFeeFactor changes for MaxBlockSystemFee and make it possible to fit
5K transactions into the block. Related to neo-project/neo-node#850.

### DIFF
--- a/src/DBFTPlugin/DBFTPlugin/config.json
+++ b/src/DBFTPlugin/DBFTPlugin/config.json
@@ -4,7 +4,7 @@
     "IgnoreRecoveryLogs": false,
     "AutoStart": false,
     "Network": 860833102,
-    "MaxBlockSize": 262144,
-    "MaxBlockSystemFee": 900000000000
+    "MaxBlockSize": 2097152,
+    "MaxBlockSystemFee": 100000000000
   }
 }


src/DBFTPlugin/DBFTPlugin/config.json
```diff
@@ -4,7 +4,7 @@
     "IgnoreRecoveryLogs": false,
     "AutoStart": false,
     "Network": 860833102,
-    "MaxBlockSize": 262144,
-    "MaxBlockSystemFee": 900000000000
+    "MaxBlockSize": 2097152,
```
shargon, line 7: For more tx lower values?

roman-khimov: It's a bit different change, our public networks have substantially lowered ExecFeeFactor these days compared to the default, so MaxBlockSystemFee should be adjusted (9K GAS is _a lot_ of processing time). And it doesn't seem to be limiting, a typical NEO/GAS transfer transaction easily fits into 0.01 GAS of system fee. I don't have complete statstics for the whole chain lifetime, but recent Flamingo or non-native NEP-17 transactions seem to take 0.1-0.2 GAS in system fee, so 1K limit still seems to be reasonable.

shargon: These configuration should be related to the height 🤔

erikzhang: > These configuration should be related to the height

This is not necessary. These are just consensus node configurations and we can change them back at any time.

shargon (approved):

PR #692, opened 2022-02-26
Prevent SSRF

shargon:
Fix https://github.com/neo-project/neo-modules/issues/693

### DIFF
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -1,3 +1,13 @@
+// Copyright (C) 2015-2021 The Neo Project.
+//
+// The Neo.Plugins.OracleService is free software distributed under the MIT software license,
+// see the accompanying file LICENSE in the main directory of the
+// project or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
 using Neo.Network.P2P.Payloads;
 using System;
 using System.Linq;
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -12,7 +12,7 @@ namespace Neo.Plugins
 {
     class OracleHttpsProtocol : IOracleProtocol
     {
-        private readonly HttpClient client = new HttpClient();
+        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false });
 
         public OracleHttpsProtocol()
         {
@@ -40,7 +40,7 @@ namespace Neo.Plugins
 
             if (!Settings.Default.AllowPrivateHost)
             {
-                IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host);
+                IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
                 if (entry.IsInternal())
                     return (OracleResponseCode.Forbidden, null);
             }
@@ -48,7 +48,20 @@ namespace Neo.Plugins
             HttpResponseMessage message;
             try
             {
+            download:
                 message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
+                if (message.StatusCode == HttpStatusCode.NotFound && message.Headers.Location is not null)
+                {
+                    uri = message.Headers.Location;
+                    if (!Settings.Default.AllowPrivateHost)
+                    {
+                        // Follow
+                        IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
+                        if (entry.IsInternal())
+                            return (OracleResponseCode.Forbidden, null);
+                    }
+                    goto download;
+                }
             }
             catch
             {
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -60,7 +60,7 @@ namespace Neo.Plugins
             {
             download:
                 message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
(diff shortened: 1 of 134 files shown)


src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -40,15 +40,28 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
                 if (entry.IsInternal())
                     return (OracleResponseCode.Forbidden, null);
             }
 
             HttpResponseMessage message;
             try
             {
+            download:
                 message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
+                if (message.StatusCode == HttpStatusCode.NotFound && message.Headers.Location is not null)
```
vang1ong7ang, line 53: should it be `Found`?

shargon: Maybe is not required to check the `StatusCode`

erikzhang:
@shargon @vang1ong7ang Please check #694

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -40,15 +40,28 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
+                {
+                    uri = message.Headers.Location;
+                    if (!Settings.Default.AllowPrivateHost)
+                    {
+                        // Follow
+                        IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
+                        if (entry.IsInternal())
+                            return (OracleResponseCode.Forbidden, null);
+                    }
+                    goto download;
```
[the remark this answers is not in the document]
shargon: Me can add a max-redirect

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -60,7 +60,7 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
             {
```
Liaojinghui, line 60:
```C#
                do
                {
                    {
                        var entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
                        if (entry.IsInternal())
                    }
                    if (message.Headers.Location is not null)
                        uri = message.Headers.Location;
                    else break;
                } while (true);
```

erikzhang (approved):

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -48,17 +48,24 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
+                        if (entry.IsInternal())
+                            return (OracleResponseCode.Forbidden, null);
+                    }
+                    message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
+                    if (message.Headers.Location is not null)
+                    {
+                        uri = message.Headers.Location;
+                        message = null;
+                    }
+                } while (message == null);
```
roman-khimov, line 68: So it's an infinite loop, right?

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -48,17 +48,24 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
-
             HttpResponseMessage message;
             try
             {
-                message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
+                do
+                {
+                    if (!Settings.Default.AllowPrivateHost)
+                    {
+                        IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
```
dusmart, line 58: It seems that we should check `message.RequestMessage.RequestUri` instead of `uri`.


> RequestMessage is set to the request message which led to this response message. In the case of a request sent using HttpClient, this property will point to the actual request message leading to the final response. Note that this may not be the same message the user provided when sending the request. This is typically the case if the request needs to be resent due to redirects or authentication. This property can be used to determine what URL actually created the response (useful in case of redirects).

> Following a redirection header does not change the original RequestUri property. 

Copied from
- [RequestMessage](https://docs.microsoft.com/et-ee/dotnet/api/system.net.http.httpresponsemessage.requestmessage?view=netframework-4.7.2#system-net-http-httpresponsemessage-requestmessage)
- [RequestUri](https://docs.microsoft.com/et-ee/dotnet/api/system.net.httpwebrequest.requesturi?view=netframework-4.7.2&viewFallbackFrom=netstandard-1.3)

dusmart: the same problem as https://github.com/neo-project/neo-modules/pull/694#discussion_r817353589

src/OracleService/Protocols/OracleHttpsProtocol.cs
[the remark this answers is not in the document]
roman-khimov: Which timeout stops infinite redirections in the current implementation?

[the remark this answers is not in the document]
dusmart: > Which timeout stops infinite redirections in the current implementation?

It seems that no timeout will stop the infinite redirections now.

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: There are a timeout

dusmart: Every `client.GetAsync` will have its own timeout. But there is no timeout for the whole loop.

What if some url redirects to itself immediately?

shargon: Reviewing the code it's right, I thought that the cancelation token was used as timeout. My bad, I will fix it.

PR #694, opened 2022-03-02
Fix redirection

### LINKED ISSUE #693: oracle service vulnerability: local information leak
the following way of avoiding local access is not enough:

https://github.com/neo-project/neo-modules/blob/32aacc468ad43600817daabbec834e715017d962/src/OracleService/Protocols/OracleHttpsProtocol.cs#L41-L46

```C#
            {
                IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host);
                if (entry.IsInternal())
            }
```

since a remote server is able to return a redirect response whose target is <https://local-address/x>.

> original issue <https://github.com/neo-project/neo/issues/2662>

erikzhang:
Fix #693

### DIFF
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -1,3 +1,13 @@
+// Copyright (C) 2015-2021 The Neo Project.
+//
+// The Neo.Plugins.OracleService is free software distributed under the MIT software license,
+// see the accompanying file LICENSE in the main directory of the
+// project or http://www.opensource.org/licenses/mit-license.php
+// for more details.
+//
+// Redistribution and use in source and binary forms with or without
+// modifications are permitted.
+
 using Neo.Network.P2P.Payloads;
 using System;
 using System.Linq;
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -22,7 +22,7 @@ namespace Neo.Plugins
 {
     class OracleHttpsProtocol : IOracleProtocol
     {
-        private readonly HttpClient client = new HttpClient();
+        private readonly HttpClient client = new(new HttpClientHandler { MaxAutomaticRedirections = 5 });
 
         public OracleHttpsProtocol()
         {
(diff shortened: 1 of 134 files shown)


src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -62,6 +72,14 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
                 return (OracleResponseCode.Error, null);
             if (!Settings.Default.AllowedContentTypes.Contains(message.Content.Headers.ContentType.MediaType))
                 return (OracleResponseCode.ContentTypeNotSupported, null);
+
+            if (!Settings.Default.AllowPrivateHost && message.RequestMessage.RequestUri != uri)
```
vang1ong7ang, line 76: I don't think it really works to prevent ssrf since the request is already sent in the line:

```
```

dusmart: Yes. It will leak information such as resource existence at least because this check is not the first branch.

dusmart: If we don't want the redirection to occur, disabling it directly may be better.

Or we'd better add a private host check for `message.RequestMessage.RequestUri`.

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -72,6 +72,14 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
+            {
+                IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
```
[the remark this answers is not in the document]
erikzhang: You are right. Fixed.

vang1ong7ang:
i prefer to avoid the request happening. not only the final content uploading.

vang1ong7ang:
[dns rebinding](https://en.m.wikipedia.org/wiki/DNS_rebinding) should be also considered

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: Agree with @vang1ong7ang , I think that the redirection must be manual like https://github.com/neo-project/neo-modules/pull/692

vang1ong7ang: yeah, should not assume `GET` requests have no side effect.

PR #695, opened 2022-03-04
limit result stack

Ashuaidehao:
Close https://github.com/neo-project/neo/issues/2666

### DIFF
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -9,7 +9,9 @@ using Neo.VM;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Neo.VM.Types;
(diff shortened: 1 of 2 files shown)


src/ApplicationLogs/LogReader.cs
```diff
@@ -71,7 +71,14 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
             trigger["gasconsumed"] = appExec.GasConsumed.ToString();
             try
             {
-                trigger["stack"] = appExec.Stack.Select(q => q.ToJson()).ToArray();
+                if (appExec.Stack.Length > Settings.Default.MaxStackItems)
```
roman-khimov, line 74: You can have a top-level array with a lot of things inside.

PR #698, opened 2022-03-11
Add oracle global timeout

shargon:
Related to https://github.com/neo-project/neo-modules/pull/692#discussion_r817332357

### DIFF
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -346,7 +346,7 @@ namespace Neo.Plugins
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
 
-            using CancellationTokenSource ctsTimeout = new(Settings.Default.Https.Timeout);
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.MaxTaskTimeout);
             using CancellationTokenSource ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cancelSource.Token, ctsTimeout.Token);
 
             try
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -346,7 +346,7 @@ namespace Neo.Plugins
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
 
-            using CancellationTokenSource ctsTimeout = new(Settings.Default.MaxTaskTimeout);
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.MaxOracleTimeout);
             using CancellationTokenSource ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cancelSource.Token, ctsTimeout.Token);
 
             try
--- a/src/OracleService/OracleService.cs
+++ b/src/OracleService/OracleService.cs
@@ -345,9 +345,13 @@ namespace Neo.Plugins
                 return (OracleResponseCode.Error, null);
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
+
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.Https.Timeout);
+            using CancellationTokenSource ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cancelSource.Token, ctsTimeout.Token);
+
             try
             {
-                return await protocol.ProcessAsync(uri, cancelSource.Token);
+                return await protocol.ProcessAsync(uri, ctsLinked.Token);
             }
             catch
             {
--- a/src/OracleService/Protocols/OracleHttpsProtocol.cs
+++ b/src/OracleService/Protocols/OracleHttpsProtocol.cs
@@ -10,6 +10,7 @@
 
 using Neo.Network.P2P.Payloads;
 using System;
+using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -51,6 +52,7 @@ namespace Neo.Plugins
             HttpResponseMessage message;
             try
             {
+                Stopwatch stopwatch = Stopwatch.StartNew();
                 do
                 {
                     if (!Settings.Default.AllowPrivateHost)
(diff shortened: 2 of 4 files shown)


src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -62,6 +64,9 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
                     message = await client.GetAsync(uri, HttpCompletionOption.ResponseContentRead, cancellation);
                     if (message.Headers.Location is not null)
                     {
+                        if (stopwatch.Elapsed > client.Timeout)
```
roman-khimov, line 67: How many requests can be made within this time frame?

Liaojinghui:
```C#
                for(int i = 0; i<5 && message == null; i++)
                {
                    {
                        IPHostEntry entry = await Dns.GetHostEntryAsync(uri.Host, cancellation);
                        if (entry.IsInternal())
                    }
                    if (message.Headers.Location is not null)
                    {
                        if (stopwatch.Elapsed > client.Timeout)
                            return (OracleResponseCode.Timeout, null);


                        uri = message.Headers.Location;
                        message = null;
                    }
                };
```

shargon: > How many requests can be made within this time frame?

It's the same time as one request

erikzhang:
Use `CancellationTokenSource`?

```c#
CancellationTokenSource ctsTimeout = new(Settings.Default.Https.Timeout);
CancellationTokenSource ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cancelSource.Token, ctsTimeout.Token);

try
{
    return await protocol.ProcessAsync(uri, ctsLinked.Token);
}
catch
{
}
```

src/OracleService/Protocols/OracleHttpsProtocol.cs
erikzhang: Agree with @roman-khimov

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -51,6 +51,7 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
             HttpResponseMessage message;
             try
             {
+                int redirects = 2;
```
vncoelho, line 54: Hi @shargon, could it be a merge between max redirects and timeout?
And sum the time taken between redirects?

Anyway, this will solve the issue for now.

src/OracleService/OracleService.cs
```diff
@@ -345,9 +345,13 @@ private async Task<(OracleResponseCode, string)> ProcessUrlAsync(string url)
                 return (OracleResponseCode.Error, null);
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
+
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.Https.Timeout);
```
erikzhang, line 349: Https and Oracle have different timeout settings.

erikzhang: Do we need to unify them?

Liaojinghui (approved):

src/OracleService/OracleService.cs
erikzhang: This issue should be fixed before merging.

superboyiii: @shargon Can you fix it?

shargon: I will do the changes

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: Then we should create a new `HttpClient` for each request, because the timeout it's specified in the class

src/OracleService/OracleService.cs
```diff
@@ -346,7 +346,7 @@ private async Task<(OracleResponseCode, string)> ProcessUrlAsync(string url)
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
 
-            using CancellationTokenSource ctsTimeout = new(Settings.Default.Https.Timeout);
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.MaxTaskTimeout);
```
erikzhang, line 349: `MaxTaskTimeout` is used to remove oracle requests from blockchain. We need to create a new field.

src/OracleService/Protocols/OracleHttpsProtocol.cs
Ashuaidehao: Why not use [httpclienthandler.maxautomaticredirections](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.maxautomaticredirections?view=net-6.0#system-net-http-httpclienthandler-maxautomaticredirections) ?

src/OracleService/OracleService.cs
```diff
@@ -345,9 +345,13 @@ private async Task<(OracleResponseCode, string)> ProcessUrlAsync(string url)
                 return (OracleResponseCode.Error, null);
             if (!protocols.TryGetValue(uri.Scheme, out IOracleProtocol protocol))
                 return (OracleResponseCode.ProtocolNotSupported, null);
+
+            using CancellationTokenSource ctsTimeout = new(Settings.Default.MaxOracleTimeout);
+            using CancellationTokenSource ctsLinked = CancellationTokenSource.CreateLinkedTokenSource(cancelSource.Token, ctsTimeout.Token);
```
Ashuaidehao, line 350: Can these two tokens merge to single one?

shargon: We want to cancel it if `cancelSource` was cancel, so we need to join them

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: Because the original issue, we can have a SSRF

src/OracleService/Settings.cs
```diff
@@ -54,6 +55,7 @@ private Settings(IConfigurationSection section)
             Network = section.GetValue("Network", 5195086u);
             Nodes = section.GetSection("Nodes").GetChildren().Select(p => new Uri(p.Get<string>(), UriKind.Absolute)).ToArray();
             MaxTaskTimeout = TimeSpan.FromMilliseconds(section.GetValue("MaxTaskTimeout", 432000000));
+            MaxOracleTimeout = TimeSpan.FromMilliseconds(section.GetValue("MaxOracleTimeout", 10000));
```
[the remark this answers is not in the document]
shargon: Done

erikzhang (approved):

PR #644, opened 2021-09-15
RpcClient missing state root related methods

devhawk:
Fixes #635

Note this is a draft PR until #638 is merged

PR #699, opened 2022-03-21
Fix StateAPI.MakeFindStatesParams

ixje:
It currently is not possible to use the `count` parameter if the `from` parameter is empty. It should be possible to set the page size (controlled by `count`) regardless if the optional `from` parameter is used. This change is needed for upcoming changes in neo express.

ping @devhawk


devhawk (approved):

devhawk:
@erikzhang @shargon Not sure if you want me to go ahead and merge this. @ixje discovered an issue in this code I originally wrote as part of #644

erikzhang (approved):

PR #707, opened 2022-04-22
copy output into the same folder under the name of the plugin

Liaojinghui:


erikzhang:
I don't think we need this change. We can just move the `config.json` file to the root folder.

Liaojinghui:
> I don't think we need this change. We can just move the `config.json` file to the root folder.

Either works fine, i will keep this pr open until we have the whole task done.

PR #714, opened 2022-05-20
fairy tester and debugger: extended RpcServer for convenient remote smart contract testing and debugging

Hecate2:
This is a huge and radical commit inherited from https://github.com/Hecate2/neo-fairy-test , and may need some potentially dangerous changes in `neo-vm`. I can make all kinds of changes or abort the pull request if necessary. 

Preliminary testing can be run semi-automatically with https://github.com/Hecate2/neo-fairy-client , along with the tested contract https://github.com/Hecate2/NFTLoan/releases/tag/0.0.1 . Manual checking is needed for the test output. 

[This commit](https://github.com/neo-project/neo/commit/f2212213761ab8f020c71e293638912794863ec7) (which may be included in neo 3.3.0) is required for method `InvokeFunctionWithSession` and `InvokeScriptWithSession` to compute network fee. 

Docs are not available for now. The basic idea of using this extension is: 
- Testing smart contracts with ready-made testnet or mainnet environment
1. Virtually deploy your fairy contracts and/or invoke fairy functions in your own snapshots (marked by session strings). You can read and utilize the testnet or even mainnet environment. The fairy server write everything in your own snapshots, and nothing to the external blockchain. No GAS is needed as system fee and network fee (but you can view the fees in the results of invoking functions). Your snapshots are saved in the variable `sessionToEngine` in `Fairy.Tester.Snapshot.cs`.
2. Change your NEO and GAS balance, or change any storage item conveniently in your own snapshot. 
3. Change `Runtime.Time` at will by setting the variable `sessionToTimestamp` in `Fairy.Tester.Snapshot.cs`.
- Debugging smart contracts based on snapshots generated by testing
4. Build your own snapshot (or just create one from `system.Storeview`) through testing APIs before debugging. 
5. Register DebugInfo with `.nefdbgnfo` and the result of `dumnef` of your contract. These files map the assembly codes back to source codes, and variable slots back to variable names. Ideally all languages (C#, Python, Go, ...) are supported by this debugger. 
6. Set breakpoints at source codes or assembly codes. 
7. Start debugging with `DebugFunctionWithSession` in `Fairy.Debugger.cs`. The vm engine stops on breakpoints or faults.

This extension should work fluently for fully automatic testing and potential web debugger.

[Necessary changes in `neo-vm` to run the debugger](https://github.com/Hecate2/neo-vm/commit/d10ab378bdd6cb332419720715643747717f8789):
Change
```csharp
internal protected set
```
of `public VMState State` to
```csharp
set
```
Change
```csharp
internal protected void ExecuteNext()
```
to
```csharp
public void ExecuteNext()
```


Hecate2:
(The current test workflows of this github repository are probably meaningless for this pull request. I can write tests if necessary.)

erikzhang:
Is it possible not modifying the `ExecutionEngine` by using [neo-vm built-in debugger](https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Debugger.cs)?

src/RpcServer/Fairy.Debugger.DebugInfo.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            string param0 = _params[0].AsString();
+            UInt160 scriptHash = UInt160.Parse(param0);
+            // nccs YourContractProject.csproj --debug
+            // find .nefdbgnfo beside your .nef contract, and
+            // give me the base64encode(content) of .nefdbgnfo file
+            JObject nefDbgNfo = JObject.Parse(Unzip(Convert.FromBase64String(_params[1].AsString())));
+            contractScriptHashToNefDbgNfo[scriptHash] = nefDbgNfo;
+            // https://github.com/devhawk/DumpNef
+            // dumpnef contract.nef > contract.nef.txt
```
roman-khimov, line 58: Why NEF itself isn't sufficient?

roman-khimov (commented):
@ixje, do you think it can be an alternative to nspcc-dev/neo-go#2406?

Hecate2:
> Is it possible not modifying the `ExecutionEngine` by using [neo-vm built-in debugger](https://github.com/neo-project/neo-vm/blob/master/src/neo-vm/Debugger.cs)?

It's possible, but probably I will need to build a brand-new debugger for that many features.

src/RpcServer/Fairy.Debugger.DebugInfo.cs
Hecate2: > Why NEF itself isn't sufficient?

nef+nefdbgnfo is actually sufficient for a map from source code line num to assembly InstructionPointer, but no source code content is included in nefdbgnfo. DumpNef results have raw source code contents included.

erikzhang:
1. Please don't modify neo-vm.
2. Maybe you can create a new plugin. Plugins can register methods in RpcServer.

Hecate2:
> 1. Please don't modify neo-vm.
> 2. Maybe you can create a new plugin. Plugins can register methods in RpcServer.

OK

Hecate2:
Refactored as a standalone plugin. No need to change `ExecutionEngine.cs` of neo-vm.
The following pull requests are required to run the new plugin:
https://github.com/neo-project/neo/pull/2751 , to allow correct [RpcMethod] registration (currently RpcMethod cannot be registered if the RpcServer plugin is loaded later than the registrant plugin which tries to register methods)
https://github.com/neo-project/neo/pull/2750 , to enabled fully-featured debugging at application engine level.

devhawk (changes requested):
I'm on vacation until 5/24 and I want to make sure I get a chance to review this in depth before it gets committed.

src/RpcServer/Fairy.Debugger.DebugInfo.cs
devhawk: DumpNef dumps the contents of the NEF file, intermixing source code if it can find the nefdbgnfo file + the source code of the original contract. I don't understand why you would need the output of DumpNef in order to debug. https://github.com/neo-project/neo-debugger certainly doesn't

devhawk:
I've only taken a cursory look, but this doesn't seem like a good candidate plugin to include in this repo. Seems like it would only be deployed on specialty nodes by developer projects.

Also, I don't think we need to update RpcServer to add custom endpoints. We refactored RpcServer to separate the RpcServerPlugin from the RpcServer code so NeoExpress could control how the RpcServer was loaded independently of the plugin Architecture. Couldn't the Fairy plugin also launch the RpcServer (and register it's custom methods) from it's own plugin?

Hecate2:
> I've only taken a cursory look, but this doesn't seem like a good candidate plugin to include in this repo. Seems like it would only be deployed on specialty nodes by developer projects.
> 
> Also, I don't think we need to update RpcServer to add custom endpoints. We refactored RpcServer to separate the RpcServerPlugin from the RpcServer code so NeoExpress could control how the RpcServer was loaded independently of the plugin Architecture. Couldn't the Fairy plugin also launch the RpcServer (and register it's custom methods) from it's own plugin?

I run this plugin in my private neo-cli to debug and test my contracts with the mainnet or testnet environment. Sometimes users need to work with other already-deployed contracts.

devhawk:
FYI, next version of Express will have a new command [`contract download`](https://github.com/neo-project/neo-express/pull/209/commits) that lets you download contract NEF + storage from a public network. (Thanks @ixje for the contribution!).

Also, feel free to open an issue in neo express and/or debugger repos for "debug and test my contracts with the mainnet or testnet environment" if that's a feature that you're interested in seeing supported in Neo Blockchain Toolkit. We already have [code](https://github.com/ngdenterprise/neo-blockchaintoolkit-library/blob/master/src/bctklib/persistence/StateServiceStore.cs) in the toolkit that can pull storage info from mainnet / testnet. That code powers the [NeoTrace](https://github.com/neo-project/neo-express#neotrace) tool. We could look at integrating more deeply into express and debugger if there is developer interest.

vncoelho (commented):
I think custom endpoints is a good option to be kept in this PR

erikzhang:
> I think custom endpoints is a good option to be kept in this PR

You can call `RpcServerPlugin.RegisterMethods` to register custom endpoints.

erikzhang:
What's the difference between assembly breakpoints and source code breakpoints?

Hecate2:
> What's the difference between assembly breakpoints and source code breakpoints?

source code breakpoints are set by source code file name + line number, and only at assembly instructions that are the beginnings of source code lines. 
assembly breakpoints are set by InstructionPointer, and can be the beginning of any assembly instruction

erikzhang:
Now we don't need neo-project/neo#2750.

src/Fairy/config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "PluginConfiguration": {
+    "MaxGasInvoke": 200,
+    "MaxIteratorResultItems": 100,
```
erikzhang, line 4: I guess we don't need it anymore because we support sessions now.

erikzhang:
Debugging via RPC is a bit odd to me. Why not build a local environment to debug?

Hecate2:
> Debugging via RPC is a bit odd to me. Why not build a local environment to debug?

It can be terribly difficult to build a contract & storage environment in local debugging

devhawk:
> > Debugging via RPC is a bit odd to me. Why not build a local environment to debug?
> 
> It can be terribly difficult to build a contract & storage environment in local debugging. The mainnet itself is the most convincing environment for debugging & testing.

Please provide a video of how this works in practice. I's like to see how this compare to the experience of using Neo Debugger for VSCode

devhawk:
> I run this plugin in my private neo-cli to debug and test my contracts with the mainnet or testnet environment. Sometimes users need to work with other already-deployed contracts. And I will try to launch Fairy with the separated RpcServerPlugin!

Why does this plugin need to be a part of the official set of modules? This doesn't seem to be a general purpose plugin

Hecate2:
> Please provide a video of how this works in practice. I's like to see how this compare to the experience of using Neo Debugger for VSCode

OK. I do not have any visual-friendly frontend interface for now, and I only have a Python client for this debugger.

devhawk:
> OK. I do not have any visual-friendly frontend interface for now, and I only have a Python client for this debugger.

Given that you can install whatever plugins you want in your own local node, I'm don't understand why this plugin should be included in this repo. It sounds like a very specialized tool with a small potential audience. 

Personally, I would want to see a more complete tool experience with a visual front end before considering accepting this PR - though I would be interested to hear what the other core developers think.

Hecate2:
> Given that you can install whatever plugins you want in your own local node, I'm don't understand why this plugin should be included in this repo. It sounds like a very specialized tool with a small potential audience.
> 
> Personally, I would want to see a more complete tool experience with a ~visual~ developer-friendly front end before considering accepting this PR - though I would be interested to hear what the other core developers think.
> 
> Update: I shouldn't have said "visual" front end. I'm a huge fan of command line interfaces. But I don't think debugging via python print statements is a reasonable approach for a general purpose tool.

The frontend can be Python, postman, a webpage

Hecate2:
Sorry but I closed the pull request by mistake...

devhawk:
> The frontend can be Python, postman, a webpage, etc. Sorry but I closed the pull request by mistake...

I think a debugger needs to have an *interactive* front end. I don't think writing python code (or postman or typescript, etc) to drive a debugger is a developer friendly experience.

Hecate2:
Maybe it's better for me to integrate this project to
https://devpost.com/software/caas-for-neo3

Hecate2:
This plugin will be maintained as an independent repository

PR #715, opened 2022-05-22
Allow traverse iterators

erikzhang:
Close neo-project/neo#2699


src/RpcServer/Signers.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using Neo.Network.P2P.Payloads;
+using Neo.Persistence;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Neo.Plugins
+{
+    class Signers : IVerifiable
```
devhawk, line 19: If we're moving this around, can we also rename this something more descriptive?

src/RpcServer/Session.cs
```diff
@@ -0,0 +1,61 @@
...
+using Neo.SmartContract.Iterators;
+using Neo.SmartContract.Native;
+using System;
+using System.Collections.Generic;
+
+namespace Neo.Plugins
+{
+    class Session : IDisposable
+    {
+        public static readonly TimeSpan ExpirationTime = TimeSpan.FromMinutes(5);
```
devhawk, line 23: ExpirationTime probably should be configurable. 5 minutes seems VERY long and subject to DOS

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -142,19 +149,14 @@ private static JObject ToJson(System.Collections.Generic.IEnumerable<DataCache.T
             return array;
         }
 
-        private static JObject ToJson(StackItem item, int max)
+        private static JObject ToJson(StackItem item, Session session)
```
devhawk, line 152: This implementation doesn't handle iterators inside of arrays or maps. Is that a scenario we consider important?

devhawk:
This is certainly a better approach than I sketched out with #708. 

One thing I got to thinking about while I was sketching #708: Should we consider exposing some RPC methods via WebSockets and/or SignalR? WebSockets would be a better model for exposing iterators that anything that runs over HTTP. Plus we could implement callback style operations like "notify me for every new block" and "submit this transaction and wait for it to be included in a block" with reasonable efficiency.

Obviously, such an approach should be a separate PR if we decide to go this route. I opened #716 to discuss

roman-khimov:
> Should we consider exposing some RPC methods via WebSockets

NeoGo provides that along with [event subscription mechanism](https://github.com/nspcc-dev/neo-go/blob/73ef36e03e4761c0542f59e9c4f5398864a291e2/docs/notifications.md).

src/RpcServer/Session.cs
erikzhang: Done.

src/RpcServer/RpcServer.SmartContract.cs
erikzhang: No, it doesn't. Currently I don't see a need for this. We can add this support in the future if the need arises.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -141,79 +149,108 @@ private static JObject ToJson(System.Collections.Generic.IEnumerable<DataCache.T
...
+            Guid sid = Guid.Parse(_params[0].GetString());
+            Guid iid = Guid.Parse(_params[1].GetString());
+            int count = _params[2].GetInt32();
+            if (count > settings.MaxIteratorResultItems)
+                throw new ArgumentOutOfRangeException(nameof(count));
+            Session session;
+            lock (sessions)
+            {
+                session = sessions[sid];
+                session.ResetExpiration();
```
shargon, line 235: If we are able to reset the expiration time, we can facilitate a DoS attack by out of memory.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -20,97 +20,105 @@
...
-                  obj["eventname"] = n.EventName;
-                  obj["contract"] = n.ScriptHash.ToString();
-                  obj["state"] = ToJson(n.State, settings.MaxIteratorResultItems);
-                  return obj;
-              }));
-            if (useDiagnostic)
-            {
-                Diagnostic diagnostic = (Diagnostic)engine.Diagnostic;
-                json["diagnostics"] = new JObject()
+            Session session = new(system, script, signers, witnesses, settings.MaxGasInvoke, useDiagnostic ? new Diagnostic() : null);
```
shargon, line 70: Do we always need to create it? Maybe it is better only if we send the parameter "cookie"

erikzhang: If we don't need it, we will destroy it later.

src/RpcServer/RpcServer.SmartContract.cs
erikzhang: I think `session` should not be enabled on public servers. I added a setting to disable it.

erikzhang:
Do we need to update the RpcClient?

devhawk:
> Do we need to update the RpcClient?

Yes. It would be nice to expose Iterators in RpcClient via IAsyncEnumerable. But I personally don't have an issue doing that as a separate PR

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -141,79 +150,108 @@ private static JObject ToJson(System.Collections.Generic.IEnumerable<DataCache.T
...
+
         [RpcMethod]
         protected virtual JObject InvokeFunction(JArray _params)
         {
             UInt160 script_hash = UInt160.Parse(_params[0].AsString());
             string operation = _params[1].AsString();
             ContractParameter[] args = _params.Count >= 3 ? ((JArray)_params[2]).Select(p => ContractParameter.FromJson(p)).ToArray() : System.Array.Empty<ContractParameter>();
-            Signers signers = _params.Count >= 4 ? SignersFromJson((JArray)_params[3], system.Settings) : null;
+            Signer[] signers = _params.Count >= 4 ? SignersFromJson((JArray)_params[3], system.Settings) : null;
+            Witness[] witnesses = _params.Count >= 4 ? WitnessesFromJson((JArray)_params[3]) : null;
```
shargon, line 203: Saved in the same _params as signers? (_params[3])

erikzhang: Yes.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -141,79 +150,108 @@ private static JObject ToJson(System.Collections.Generic.IEnumerable<DataCache.T
...
-                while (max > 0 && iterator.Next())
-                {
-                    array.Add(iterator.Value(null).ToJson());
-                    max--;
-                }
-                json["iterator"] = array;
-                json["truncated"] = iterator.Next();
+                Guid id = Guid.NewGuid();
+                session.Iterators.Add(id, iterator);
+                json["id"] = id.ToString();
```
AnnaShaleva, line 160: The resulting JSON contains `id` for each iterator even if sessions are disabled. Is it the desired behaviour?

AnnaShaleva, line 160: The resulting JSON contains `id` for each iterator even if sessions are disabled. Is it the desired behaviour?

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -20,97 +20,106 @@
...
-                json["stack"] = new JArray(engine.ResultStack.Select(p => ToJson(p, settings.MaxIteratorResultItems)));
+                session.Dispose();
+                throw;
             }
-            catch (InvalidOperationException)
+            if (session.Iterators.Count == 0 || !settings.SessionEnabled)
             {
-                json["stack"] = "error: invalid operation";
+                session.Dispose();
             }
```
AnnaShaleva, line 116: Could you, please, clarify: if sessions [are supposed to be disabled on the public servers](https://github.com/neo-project/neo-modules/pull/715#discussion_r887395316) (mainnet, testnet), then there's no way to get iterator result from `invokefunction` and we should use `invokescript` instead to manually traverse iterator and push its values on stack. 

That's not quite convenient IMO, maybe we still better to keep the old way of iterator dumping if sessions are disabled?

AnnaShaleva, line 116: Could you, please, clarify: if sessions [are supposed to be disabled on the public servers](https://github.com/neo-project/neo-modules/pull/715#discussion_r887395316) (mainnet, testnet), then there's no way to get iterator result from `invokefunction` and we should use `invokescript` instead to manually traverse iterator and push its values on stack. 

That's not quite convenient IMO, maybe we still better to keep the old way of iterator dumping if sessions are disabled?

erikzhang: I think public RPC nodes should only expose part of the API. And only use the private RPC server for production.

ProDog:
Tested OK with https://github.com/neo-project/neo-modules/pull/721.

superboyiii (approved):

PR #721, opened 2022-06-03
Rpc Client iterators

Ashuaidehao:
Relate https://github.com/neo-project/neo-modules/pull/715

### DIFF
--- a/src/RpcClient/Models/RpcInvokeResult.cs
+++ b/src/RpcClient/Models/RpcInvokeResult.cs
@@ -30,7 +30,7 @@ namespace Neo.Network.RPC.Models
 
         public string Exception { get; set; }
 
-        public string SessionId { get; set; }
+        public string Session { get; set; }
 
         public JObject ToJson()
         {
@@ -63,7 +63,7 @@ namespace Neo.Network.RPC.Models
             };
             if (json.ContainsProperty("exception"))
                 invokeScriptResult.Exception = json["exception"]?.AsString();
-            if (json.ContainsProperty("session")) invokeScriptResult.SessionId = json["session"]?.AsString();
+            if (json.ContainsProperty("session")) invokeScriptResult.Session = json["session"]?.AsString();
             try
             {
                 invokeScriptResult.Stack = ((JArray)json["stack"]).Select(p => Utility.StackItemFromJson(p)).ToArray();
--- a/src/RpcClient/Models/RpcInvokeResult.cs
+++ b/src/RpcClient/Models/RpcInvokeResult.cs
@@ -30,6 +30,8 @@ namespace Neo.Network.RPC.Models
 
         public string Exception { get; set; }
 
+        public string SessionId { get; set; }
+
         public JObject ToJson()
         {
             JObject json = new();
@@ -61,6 +63,7 @@ namespace Neo.Network.RPC.Models
             };
             if (json.ContainsProperty("exception"))
                 invokeScriptResult.Exception = json["exception"]?.AsString();
+            if (json.ContainsProperty("session")) invokeScriptResult.SessionId = json["session"]?.AsString();
             try
             {
                 invokeScriptResult.Stack = ((JArray)json["stack"]).Select(p => Utility.StackItemFromJson(p)).ToArray();
--- a/src/RpcClient/Models/RpcInvokeResult.cs
+++ b/src/RpcClient/Models/RpcInvokeResult.cs
@@ -61,9 +61,8 @@ namespace Neo.Network.RPC.Models
                 State = json["state"].TryGetEnum<VMState>(),
                 GasConsumed = long.Parse(json["gasconsumed"].AsString()),
             };
-            if (json.ContainsProperty("exception"))
-                invokeScriptResult.Exception = json["exception"]?.AsString();
-            if (json.ContainsProperty("session")) invokeScriptResult.Session = json["session"]?.AsString();
+            if (json.ContainsProperty("exception")) invokeScriptResult.Exception = json["exception"]?.AsString();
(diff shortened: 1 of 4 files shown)


src/RpcClient/Models/RpcInvokeResult.cs
```diff
@@ -30,6 +30,8 @@ public class RpcInvokeResult
 
         public string Exception { get; set; }
 
+        public string SessionId { get; set; }
```
erikzhang, line 33:
```suggestion
        public string Session { get; set; }
```

ProDog:
Tested OK.

erikzhang:
Why force push?

src/RpcClient/Utility.cs
```diff
@@ -253,9 +253,18 @@ public static StackItem StackItemFromJson(JObject json)
                 case StackItemType.Pointer:
                     return new Pointer(null, (int)json["value"].AsNumber());
                 case StackItemType.InteropInterface:
-                    return new InteropInterface(new object());
+                    return new InteropInterface((json.ContainsProperty("id") ? json["id"]?.AsString() : string.Empty) ?? string.Empty);
```
erikzhang, line 256: I think we shouldn't use `string` as the inner type. We should create a new type. Otherwise we cannot extend it in the future.

Ashuaidehao: Use JObject as inner type directly？

src/RpcClient/RpcClient.cs
```diff
@@ -459,6 +450,43 @@ public async Task<RpcUnclaimedGas> GetUnclaimedGasAsync(string address)
...
+        {
+            var result = await RpcSendAsync(GetRpcName(), sessionId, id, count).ConfigureAwait(false);
+            return result as JArray;
+        }
+
+        /// <summary>
+        /// Returns all results from Iterator.
+        /// This RPC call does not affect the blockchain in any way.
+        /// </summary>
+        public async IAsyncEnumerable<JObject> TraverseIteratorAllAsync(string sessionId, string id, int count)
```
erikzhang, line 467: Merge it with TraverseIteratorAsync?

src/RpcClient/Models/RpcInvokeResult.cs
```diff
@@ -59,8 +61,8 @@ public static RpcInvokeResult FromJson(JObject json)
                 State = json["state"].TryGetEnum<VMState>(),
                 GasConsumed = long.Parse(json["gasconsumed"].AsString()),
             };
-            if (json.ContainsProperty("exception"))
-                invokeScriptResult.Exception = json["exception"]?.AsString();
+            if (json.ContainsProperty("exception")) invokeScriptResult.Exception = json["exception"]?.AsString();
```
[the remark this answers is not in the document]
ProDog: Remove `?` will cause an error:
```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=RpcClient
  StackTrace:
   at Neo.Network.RPC.Models.RpcInvokeResult.FromJson(JObject json)
```

src/RpcClient/RpcClient.cs
Ashuaidehao: Which result type should the merged method use, JArray or IAsyncEnumerable?

src/RpcClient/Models/RpcInvokeResult.cs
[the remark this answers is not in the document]
erikzhang: We don't need `if (json.ContainsProperty("exception"))` check.

src/RpcClient/RpcClient.cs
erikzhang: `IAsyncEnumerable`

ProDog:
Tested OK.

superboyiii (approved):

ProDog (approved):

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -157,6 +157,7 @@ private static JObject ToJson(StackItem item, Session session)
             {
                 Guid id = Guid.NewGuid();
                 session.Iterators.Add(id, iterator);
+                json["interface"] = nameof(IIterator);
```
shargon, line 160: Use "Iterator" instead of "IIterator"?

AnnaShaleva, line 160: Could you, please, clarify, why do we need this line? We already have `JObject json = item.ToJson();` a few lines above, it creates `{"type": "InteropInterface" }`. After that, if we have an `id` inside this InteropInterface, then it's clear that it's Iterator hidden under the hood.

Are we planning to extend this API later?

erikzhang: > Are we planning to extend this API later?

Yes.

src/RpcClient/RpcClient.cs
```diff
@@ -459,6 +450,49 @@ public async Task<RpcUnclaimedGas> GetUnclaimedGasAsync(string address)
             return RpcUnclaimedGas.FromJson(result);
         }
 
+        /// <summary>
+        /// Returns all results from Iterator.
+        /// This RPC call does not affect the blockchain in any way.
+        /// </summary>
+        public async IAsyncEnumerable<JObject> TraverseIteratorAsync(string sessionId, string id, int count, bool fetchAll = false)
```
erikzhang, line 457: What about remove `fetchAll`? If `count == -1`, we fetch all items.

Ashuaidehao: How to specify the rpc parameter "count" value in this scenario?

erikzhang: `100` or something...

ProDog (approved):

PR #730, opened 2022-06-14
Don't log index check error

### LINKED ISSUE #724: Restart consensus node will cause InvalidOperationException
**Describe the bug**
Restart consensus node will cause InvalidOperationException

**To Reproduce**
1. Create a private chain, whatever single consensus or 4 consensus
2. Start consensus and stop
3. Restart

**Expected behavior**
`System.InvalidOperationException: Operation is not valid due to the current state of the object.`
Persist could still go on.

**Screenshots**
image

ZhangTao1596:
Close #724

### DIFF
--- a/src/DBFTPlugin/Consensus/ConsensusContext.cs
+++ b/src/DBFTPlugin/Consensus/ConsensusContext.cs
@@ -168,6 +168,10 @@ namespace Neo.Consensus
             {
                 Deserialize(ref reader);
             }
+            catch (InvalidOperationException)
+            {
+                return false;
+            }
             catch (Exception exception)
             {
                 Utility.Log(nameof(ConsensusContext), LogLevel.Debug, exception.ToString());
--- a/src/DBFTPlugin/Consensus/ConsensusContext.cs
+++ b/src/DBFTPlugin/Consensus/ConsensusContext.cs
@@ -264,6 +264,11 @@ namespace Neo.Consensus
             store.PutSync(ConsensusStateKey, this.ToArray());
         }
 
+        public void Delete()
+        {
+            store.Delete(ConsensusStateKey);
+        }
+
         public void Deserialize(ref MemoryReader reader)
         {
             Reset(0);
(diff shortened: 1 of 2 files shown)


superboyiii:
Test OK.

superboyiii (dismissed):

shargon (approved):

vncoelho (changes requested):
1 week for me to check.
But looks good.

src/DBFTPlugin/Consensus/ConsensusService.cs
```diff
@@ -72,6 +72,8 @@ private void OnPersistCompleted(Block block)
         {
             Log($"Persisted {nameof(Block)}: height={block.Index} hash={block.Hash} tx={block.Transactions.Length} nonce={block.Nonce}");
             knownHashes.Clear();
+            if (context.Block.Index == block.Index && context.CommitSent)
```
vncoelho, line 75: Why does it need to be Committed?
Maybe it directly received the block.

ZhangTao1596: Only committed node saves context.

vncoelho (approved):

src/DBFTPlugin/Consensus/ConsensusContext.cs
```diff
@@ -264,6 +264,11 @@ public void Save()
             store.PutSync(ConsensusStateKey, this.ToArray());
         }
 
+        public void Delete()
+        {
+            store.Delete(ConsensusStateKey);
```
erikzhang, line 269: We don't need to delete it. Because it will be replaced in the next block.

ZhangTao1596: Updated. How do you think?

superboyiii: > We don't need to delete it. Because it will be replaced in the next block.

But will show error in log.

ZhangTao1596: This log is necessary. You use a wrong Neo.dll.

erikzhang (approved):

shargon (approved):

vncoelho:
Perfect,maybe update PR title.

PR #734, opened 2022-06-23
Fix find states

ZhangTao1596:


src/StateService/StatePlugin.cs
```diff
@@ -322,11 +337,11 @@ public JObject FindStates(JArray _params)
             };
             if (0 < jarr.Count)
             {
-                json["firstProof"] = GetProof(trie, contract.Id, Convert.FromBase64String(jarr.First()["key"].AsString()));
+                json["firstProof"] = GetProof(trie, ParseStorageKey(Convert.FromBase64String(jarr.First()["key"].AsString())));
```
shargon, line 340: Why not use contract.Id?

erikzhang (dismissed):

superboyiii (approved):
Tested OK.

src/StateService/StatePlugin.cs
ZhangTao1596: Updated.

erikzhang (approved):

PR #746, opened 2022-07-26
InvokeTwice

Ashuaidehao:
Relate https://github.com/neo-project/neo-modules/pull/743


shargon:
Won't this negate the use of `GasLeft`?  Some projects might use the GasLeft method to either execute or not execute code, or simply to iterate.

Ashuaidehao:
> Won't this negate the use of `GasLeft`? Some projects might use the GasLeft method to either execute or not execute code, or simply to iterate.

It depends. It may happens even without this changes.

src/RpcServer/Session.cs
```diff
@@ -34,14 +35,17 @@ public Session(NeoSystem system, byte[] script, Signer[] signers, Witness[] witn
...
-                NetworkFee = 1_00000000,
                 Signers = signers,
                 Attributes = Array.Empty<TransactionAttribute>(),
                 Script = script,
                 Witnesses = witnesses
             };
             Engine = ApplicationEngine.Run(script, Snapshot, container: tx, settings: system.Settings, gas: gas, diagnostic: diagnostic);
+            if (Engine.State == VMState.HALT && tx != null)
+            {
+                tx.SystemFee = Engine.GasConsumed;
```
erikzhang, line 46: And `NetworkFee`?

Ashuaidehao: Use 1 Gas?

erikzhang (approved):

shargon (approved):

roman-khimov:
If both invocations end up the same way, then why do we do the second one? If the result of two invocations is different, then which is the correct one and what should be returned to the user?

BTW, network fee can probably be roughly estimated (not precisely, but still).

erikzhang:
> If the result of two invocations is different, then which is the correct one and what should be returned to the user?

The second one is the correct one.

roman-khimov:
If we suppose different execution results, then the second run can easily end up in FAULT state just because of insufficient GAS (`gas: Engine.GasConsumed` from the first invocation).

erikzhang:
> If we suppose different execution results, then the second run can easily end up in FAULT state just because of insufficient GAS

Why?

roman-khimov:
If contract's logic depends on fees it might take some longer branch on the second execution, so the GAS amount from the first execution won't be enough. I'd suggest using `gas: gas` for the second one, but even that is not perfect of course. If `Engine.GasConsumed` will be different, then we don't know what's going to happen with the contract in a transaction with this new SystemFee.

I don't see any perfect solution to this, setting some semi-random value or using statistical data to set fees for executions is wrong as well. And code can always check `Runtime.GasLeft` to see that it has more (or less) GAS there than there is in the `SystemFee`.

erikzhang:
> If contract's logic depends on fees it might take some longer branch on the second execution

The idea is that if a contract's logic depends on fees, it checks if `sys_fee` is 0 to determine if it's currently a test execution.

vncoelho (commented):
Looks go to me, I just need to test, the idea is to merge @superboyiii ?

src/RpcServer/Session.cs
```diff
@@ -41,6 +42,12 @@ public Session(NeoSystem system, byte[] script, Signer[] signers, Witness[] witn
...
+                tx.NetworkFee = 1_00000000;
+                Engine = ApplicationEngine.Run(script, Snapshot, container: tx, settings: system.Settings, gas: Engine.GasConsumed, diagnostic: diagnostic);
```
roman-khimov, line 49: So, what if it fails?

vncoelho: If it fails it will be a good change, because before it was `HALT`.
The flow looks like to be the same, @roman-khimov, nothing will change from the original flow, it just add a more precise gas limit

roman-khimov: > If it fails it will be a good change, because before it was HALT.

But what's good about it? Estimation failed? How do I send a transaction then?

I know these things are not likely to happen, but contracts checking fees shouldn't exist as well and yet they do.

> it just add a more precise gas limit

For normal contracts it changes nothing (except for additional execution overhead), for strange contracts that care about fee values it can lead to any result --- some will work, some will fail.

shargon:
I think that the project should use GasLeft, it's a very specific case that can be solved in this specific smart contract

vncoelho:
To, you mean make another change that set GasLeft as default?

vncoelho:
I think the problem here is more related to the simulation when we invoke RPC, no?

Jim8y:
close  as not planed and author agree to close

Issue #742, opened 2022-07-07
RPC server can't handle parameterless requests


roman-khimov:
**Describe the bug**
JSON-RPC defines `params` member of the Request object in the following way:
> A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted.

And we have a number of calls that take no parameters, so one could try omitting the `params`, but Neo RPC server doesn't allow that.

**To Reproduce**
```
$ curl -sd '{ "jsonrpc": "2.0", "id": 5, "method": "getversion"}' http://seed4.neo.org:10332 
{"jsonrpc":"2.0","id":5,"error":{"code":-32600,"message":"Invalid Request"}}
```
Or, a bit more provocative (`null` technically is not even a structured value, but clients can easily generate it in absence of real `params`):
```
$ curl -sd '{ "jsonrpc": "2.0", "id": 5, "method": "getversion", "params": null}' http://seed4.neo.org:10332 
{"jsonrpc":"2.0","id":5,"error":{"code":-32600,"message":"Invalid Request"}}
```
**Expected behavior**
Correct response, just as with an empty array of parameters:
```
$ curl -sd '{ "jsonrpc": "2.0", "id": 5, "method": "getversion", "params": []}' http://seed4.neo.org:10332 
{"jsonrpc":"2.0","id":5,"result":{"tcpport":10333,"wsport":10334,"nonce":1379042831,"useragent":"/Neo:3.3.1/","protocol":{"addressversion":53,"network":860833102,"validatorscount":7,"msperblock":15000,"maxtraceableblocks":2102400,"maxvaliduntilblockincrement":5760,"maxtransactionsperblock":512,"memorypoolmaxtransactions":50000,"initialgasdistribution":5200000000000000}}}
```

**Platform:**
 - OS: Any
 - Version neo-cli 3.3.1

**(Optional) Additional context**
It's not critical, but allowing this for requests that don't take any parameters can increase chances for successful communication in some cases.

roman-khimov:
Fixed in #769 and 3.5.0 release, btw.

PR #747, opened 2022-08-05
Allow parameterless requests

shargon:
Close https://github.com/neo-project/neo-modules/issues/742


shargon:
@superboyiii tested?

PR #751, opened 2022-08-24
Add Build and Sign Tx

Ashuaidehao:


src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -229,6 +229,113 @@ protected virtual JToken SendFrom(JArray _params)
...
+            return json;
+        }
+
+        /// <summary>
+        /// treat signer as single Signature
+        /// </summary>
+        /// <param name="snapshot"></param>
+        /// <param name="tx"></param>
+        /// <returns></returns>
+        private long QuickCalculateNetworkFee(DataCache snapshot, Transaction tx)
```
roman-khimov, line 321: But why? `calculatenetworkfee` implementation does it this way:
```
            long networkfee = (wallet ?? new DummyWallet(system.Settings)).CalculateNetworkFee(system.StoreView, tx.AsSerializable<Transaction>());
```
And if you're only support using one simple account you might as well create some fake `VerificationScript` for it.

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -229,6 +229,113 @@ protected virtual JToken SendFrom(JArray _params)
...
+            tx.Witnesses ??= Array.Empty<Witness>();
+            json["tx"] = tx.ToJson(system.Settings);
+            return json;
+        }
+
+        private JObject SendWithoutSign(UInt160 assetId, UInt160 from, UInt160 to, BigInteger amount, Signer[] signers, long? networkFee = null, long? sysFee = null, uint? untilBlock = null)
+        {
+            using var snapshot = system.GetSnapshot();
+            var sb = new ScriptBuilder();
+            sb.EmitDynamicCall(assetId, "transfer", from, to, amount, null);
```
roman-khimov, line 278: No `data` parameter?

src/RpcServer/RpcServer.Wallet.cs
```diff
@@ -229,6 +229,113 @@ protected virtual JToken SendFrom(JArray _params)
             return SignAndRelay(snapshot, tx);
         }
 
+
+        [RpcMethod]
+        protected virtual JToken BuildTransferTx(JArray _params)
```
roman-khimov, line 234: It's _so_ specific, that I'm not sure it's right way to go. Notice that it's NEP-17 transfer that is created, how about NEP-11? How about NEO voting? Flamingo invocations?

Clients that have all of these parameters should have zero problems creating a transaction on their side already. It effectively does four things:
 * creates a script from `transfer` parameters (but that's what `invokefunction` does already)
 * calculates system fee if needed (also done by `invokefunction`)
 * calculates network fee if needed (`calculatenetworkfee` solves that)
 * calculates VUB value (`getblockcount` + any client-side logic)

All of this can be easily hidden by some higher-level API client-side building on top of the functions provided by the RPC server now. Like https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/rpcclient/actor for example. And then something like https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/rpcclient/nep17 building on top of that.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: DummyWallet won't work here.  The result tx could be sign and relay, it's not logical to set a fake script.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: This method is base on "sendfrom"  and be used as a option of "sendfrom" without sign. So it won't support to many complicated scenarios.

src/RpcServer/RpcServer.Wallet.cs
roman-khimov: You can set it, calculate the fee and then remove again. I'd worry more about code duplication (12 case-specific lines) rather than two lines of additional code.

src/RpcServer/RpcServer.Wallet.cs
roman-khimov: Why don't you add a "dontsign" parameter to `sendfrom` then? To me it all looks like a dead end, we can't cover all potential scenarios and we should not be doing that, RPC interface provides enough basic building blocks to construct _any_ transaction on the client side. I doubt anyone is creating transactions using this interface with `curl`, they use some libraries built on top of that. And these libraries can provide any kind of more developer-friendly API, they don't need this new RPC to do that. Take [neow3j for example](https://neow3j.io/#/neo-n3/dapp_development/transactions):
```java
byte[] script = new ScriptBuilder().contractCall(NeoToken.SCRIPT_HASH, "symbol", null)
                                   .toArray();

Transaction tx = new TransactionBuilder(neow3j)
    .script(script)
    .signers(AccountSigner.calledByEntry(account))
    .wallet(wallet)
    .sign();

tx.send();
```
That's it. I don't see any `invokefunction`, `calculatenetworkfee` or `getblockcount` or anything else like that there. Yet this snippet creates and sends a correct transaction using regular RPC API.

src/RpcServer/RpcServer.Wallet.cs
Ashuaidehao: How to make [this line](https://github.com/neo-project/neo/blob/5f4c14e3801c691ceccb5705c634e9e55dd3bba1/src/Neo/Wallets/Wallet.cs#L602) work? It couldn't be just 2 lines.

src/RpcServer/RpcServer.Wallet.cs
superboyiii: Some exchanges are willing to create non-signature transactions by operation team on a public node, but need finance team to manage and sign these transactions on an offline node. It's more safe and more easy to manage.

src/RpcServer/RpcServer.Wallet.cs
roman-khimov: Use fake verification script for https://github.com/neo-project/neo/blob/5f4c14e3801c691ceccb5705c634e9e55dd3bba1/src/Neo/Wallets/Wallet.cs#L609-L612 ?

src/RpcServer/RpcServer.Wallet.cs
roman-khimov: > exchanges
> on a public node
> sign these transactions on an offline node
> It's more safe

Really? So they trust some public server they don't control to create a script (!) that they will then sign blindly (not knowing what the side-effects are)? I'd very interested to know the list of exchanges doing that, even though I know I'm not likely to get one.

But anyway, even in this scenario operation team can create a transaction already with the APIs we have. Using any RPC library we have. And even some CLI incantation like `./bin/neo-go wallet nep17 transfer --from XYZ --to ZYX --amount 100500 --token NEO -w ... -r ... --out to_sign.json` should be able to do that (I've not checked if it will do that exactly, but it can be done).

Ashuaidehao: They can check the script after build the tx, so use a public server won't be a problem. And exchanges wouldn't like use un-official apis so we need add this to our offical api.

roman-khimov: > They can check the script after build the tx, so use a public server won't be a problem

Well, then using _anything_ to create the transaction won't be a problem, since it'd be checked independently. BTW, is there any "official" tool that actually allows to do that (check the script/test-run transaction from a file)? We have that in `util txdump` command :
```
$ ./bin/neo-go contract invokefunction -w .docker/wallets/wallet1.json --out some.part.json -a NVTiAjNgagDkTr5HTzDmQP9kPwPHN5BgVq -r http://localhost:30333 0x49cf4e5378ffcd4dec034fd98a174c5491e395e2 designateAsRole 8 \[ 02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2 \] -- NVTiAjNgagDkTr5HTzDmQP9kPwPHN5BgVq:CalledByEntry
Enter account NVTiAjNgagDkTr5HTzDmQP9kPwPHN5BgVq password > 
f143059e0c03546db006608e0a0ad4b621b311a48d7fc62bb7062e405ab8e588
$ ./bin/neo-go util txdump -r http://localhost:30333 some.part.json 
Hash:                   f143059e0c03546db006608e0a0ad4b621b311a48d7fc62bb7062e405ab8e588
OnChain:                false
ValidUntil:             6004
Signer:                 NVTiAjNgagDkTr5HTzDmQP9kPwPHN5BgVq (CalledByEntry)
SystemFee:              0.0208983 GAS
NetworkFee:             0.044159 GAS
Script:                 DCECs2Ir9AF73+MXxYrtX0x1PyBrfbiWBG+n13S7xL9/jcIRwBgSwB8MD2Rlc2lnbmF0ZUFzUm9sZQwU4pXjkVRMF4rZTwPsTc3/eFNOz0lBYn1bUg==
INDEX    OPCODE       PARAMETER
0        PUSHDATA1    02b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc2    <<
35       PUSH1        
36       PACK         
37       PUSH8        
38       PUSH2        
39       PACK         
40       PUSH15       
41       PUSHDATA1    64657369676e6174654173526f6c65 ("designateAsRole")
58       PUSHDATA1    e295e391544c178ad94f03ec4dcdff78534ecf49
80       SYSCALL      System.Contract.Call (627d5b52)
{
 "state": "HALT",
 "gasconsumed": "2089830",
 "script": "DCECs2Ir9AF73+MXxYrtX0x1PyBrfbiWBG+n13S7xL9/jcIRwBgSwB8MD2Rlc2lnbmF0ZUFzUm9sZQwU4pXjkVRMF4rZTwPsTc3/eFNOz0lBYn1bUg==",
 "stack": [
  {
   "type": "Any"
  }
 ],
 "exception": null,
 "notifications": [
  {
   "contract": "0x49cf4e5378ffcd4dec034fd98a174c5491e395e2",
   "eventname": "Designation",
   "state": {
    "type": "Array",
    "value": [
     {
      "type": "Integer",
      "value": "8"
     },
     {
      "type": "Integer",
      "value": "245"
     }
    ]
   }
  }
 ]
}
```

> And exchanges wouldn't like use un-official apis so we need add this to our offical api.

It's still hard for me to imagine people `curl`ing these transactions out of the RPC server, and if they'd use some interface (CLI/web/punch cards) then it doesn't matter much how many official API calls are made (`neo-go wallet nep17 transfer` uses the same official APIs as well, all wallets use them) as long as the transaction is correct and does what it's intended to. If they don't trust any wallet implementation they might as well create some task-specific 20-line script that'd do that.

superboyiii: Simple send is just `transfer`, `from`, `to` and `amount`, they don't have different form. These exchanges have simple senerio, they have their own tool to check OP before to sign.

superboyiii: > Really? So they trust some public server they don't control to create a script (!) that they will then sign blindly (not knowing what the side-effects are)?
Their operation team has their own node, so it's not public, no risk, they trust their own.

superboyiii: Even not for exchange, isolate a transaction with two roles: `script maker` and `signer`, I think it's normal and have many senerio.

roman-khimov: > These exchanges have simple senerio

That's my primary concern, really. One exchange has one scenario (NEP-17), another one a different one (NEP-11), then some may need to send a divisible NEP-11, then someone would like to append `data`... Are we adding case-specific APIs for all of them? I doubt that. We should provide sufficient generic APIs to cover all of these scenarios and we already do that.

> Operator don't need to own wallet.

It's not a problem already, there are APIs that allow to create complete (but not yet signed) transaction client-side. Everybody is using them, signing resulting transactions in any way on the client side, it all works.

superboyiii: > > These exchanges have simple senerio
> 
> That's my primary concern, really. One exchange has one scenario (NEP-17), another one a different one (NEP-11)

Can we support NEP-17 first? Maybe can rename it as `SendUnsigned` to compare with `Send`. It's useful from my sight of view.

roman-khimov: > Can we support NEP-17 first?

To have something second, third and so on then? That's what I'm trying to avoid.

> Maybe can rename it as SendUnsigned to compare with Send

There is a big difference with `sendfrom`/`sendmany` (that NeoGo does not support at all, btw), in that they have a wallet server-side (that's why it does not support them) and they can be precise in their fee calculations. This new call can't be, it assumes that the account is a simple signature account, but serious company-level accounts can easily be multisignature which this API can't handle.

I'd do it this way (a part of NeoGo master, to be released in 0.99.3):
https://github.com/nspcc-dev/neo-go/blob/20224cb39ca372abcd4d2745c8a646e607744654/docs/cli.md#offline-signing

And this way (or any other similar way) I can also do NEP-11, divisible NEP-11, voting, pass data to the contract, do contract deployments or updates and basically _any_ invocation that I'd like to do. And I can do them from a multisignature account as well. Or a contract account. All of that using existing RPC APIs.

If we want some simpler _generic_ transaction-creating API (I'm not sure we do, but) then it's pretty easy to add `ValidUntilBlock` into `invoke*` results. Then the only thing missing there would be the network fee, but if we have witnesses added into the call (and they can be) then even that can be calculated immediately. Maybe that's the thing we want, one call that has it all (but transaction will be constructed on the client from invocation results anyway, it's just that we won't need to call `calculatenetworkfee` and `getblockcount`) .

vncoelho: I understand the discussion pointed by roman.
But, in my point of view, it is just a simple tool for double checking. Those who use it should know the risks and also the same as invokefunction.

It is useful.

superboyiii: I think maybe this topic and method name has made some misunderstanding, our basic requirement is to provide an ordinary sending transaction without signature which is compared with send method in RPC. It's a normal requirement from operator and admin to build and sign. It's better to support NEP-11 but we haven't supported NEP-11 in send method as well. We can improve this in another command in another PR.

Jim8y:
close as not planed.

PR #762, opened 2022-09-16
Just extend by factor if commit is from current view

vncoelho:


roman-khimov (approved):
Looks like we've had this behavior for some time already, seems to work fine:
https://github.com/nspcc-dev/dbft/commit/57394a124fd74bf0d3679e0f79eb1f1f8559bfee
nspcc-dev/dbft#16

vncoelho:
It good that you noticed that @roman-khimov, and also good to know that you already modified in the neo-go version.
It would had been good if you had communicated it here at the time, in order to keep nodes with same protocol.

In our vision, this change is needed to keep consistent with the mechanism that it is made for.

roman-khimov:
>  It would had been nice if you had communicated it here at the time, in order to keep nodes with same protocol.

Vitor, you're absolutely right and in general we do report on things we try/implement (like https://github.com/neo-project/neo/issues/2057 or https://github.com/neo-project/neo/issues/2058 or many other non-dBFT things), it's just this particular change was made quite a while ago, we were not entirely sure on it and we didn't track these deviations properly back then.

But now it's more than two years old, kinda tested, should be fine to have it here too.

vncoelho:
@roman-khimov, sounds good, you guys are doing a great job!

erikzhang (approved):

PR #763, opened 2022-09-16
add throwOnError param to RpcClient Send/SendAsync

devhawk:
The FindStates JSON RPC method in the state service API throws a Key Not Found exception when the caller asks for an invalid key. On the calling side, I would like to receive the RpcResponse directly, without letting `AsRpcResponse` automatically throw an exception when the RPcResponse `Error` field is not null


shargon (approved):

devhawk:
Am I supposed to merge this or am I supposed to wait for @erikzhang or @shargon to do it?

PR #765, opened 2022-09-21
Fix prepare request in recovery message

ZhangTao1596:
`ValidatorIndex` is uninitialized and unassigned in `PrepareRequestMessage` of `RecoveryMessage`. This will cause prepare request reverify failure when recovery.

### DIFF
--- a/src/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
+++ b/src/DBFTPlugin/Consensus/ConsensusContext.MakePayload.cs
@@ -139,6 +139,7 @@ namespace Neo.Consensus
                     Timestamp = Block.Timestamp,
                     Nonce = Block.Nonce,
                     BlockIndex = Block.Index,
+                    ValidatorIndex = Block.PrimaryIndex,
                     TransactionHashes = TransactionHashes
                 };
             }


roman-khimov:
It can be set on the receiver side as well, somewhere in `GetPrepareRequestPayload`: 
https://github.com/neo-project/neo-modules/blob/c6d9640548d409e94873215ffb6ed05f8273c65d/src/DBFTPlugin/Messages/RecoveryMessage/RecoveryMessage.cs#L89-L95

That's what NeoGo currently does in https://github.com/nspcc-dev/neo-go/blob/5eb4ba772f4d0e507d8a77699b3856f340c36fe8/pkg/consensus/recovery_message.go#L205

vncoelho:
I agree with @roman-khimov , this will also be better for mainnet compatibility at this moment.

vncoelho:
@ZhangTao1596, I am double checking your commit, in fact, it will not affect payload itself.
As we see, currently, it is not assigned and keeps `null`?

vncoelho (approved):
@superboyiii, if possible try an extended test.

We use to have Unit Tests for consensus, that we created sometime ago.
@erikzhang, do you know why they were not ported for `neo-modules`?

That Akka tests can cover several cases, such as this.

superboyiii:
> @superboyiii, if possible try an extended test.
> 
> We use to have Unit Tests for consensus, that we created sometime ago. @erikzhang, do you know why they were not ported for `neo-modules`?
> 
> That Akka tests can cover several cases, such as this.

I will.

shargon (approved):

superboyiii (approved):
Tested, it works well.

PR #769, opened 2022-10-04
RpcServer should support missing params request property

### LINKED ISSUE #768: RpcServer should support missing params request property
Some RPC methods (such as getbestblockhash) don't have any parameters. However, if the caller doesn't include an empty array in the request, `ProcessRequestAsync` returns an `Invalid Request` error message.

I recommend updating `ProcessRequestAsync` to validate the `params` property is an array if present and pass an empty array into the RPC method if not present

From [JSON-RPC 2.0 specification](https://www.jsonrpc.org/specification#request_object):

> **params**: A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted.

devhawk:
Fixes #768

Note, there are no tests for RpcServer currently. Should I add one for this functionality?

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -225,7 +225,8 @@ namespace Neo.Plugins
         private async Task<JObject> ProcessRequestAsync(HttpContext context, JObject request)
         {
             if (!request.ContainsProperty("id")) return null;
-            if (!request.ContainsProperty("method") || !request.ContainsProperty("params") || !(request["params"] is JArray))
+            JToken @params = request["params"] ?? new JArray();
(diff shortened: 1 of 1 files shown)


roman-khimov:
See also #747.

devhawk:
> See also #747.

I think this is a cleaner fix than #747, but I'm cool with merging either. @shargon make the call

shargon (approved):

shargon:
> > See also #747.
> 
> I think this is a cleaner fix than #747, but I'm cool with merging either. @shargon make the call

Agree

superboyiii (approved):

superboyiii:
Tested OK.

superboyiii:
@erikzhang Could you have a look?

Issue #609, opened 2021-07-06
Contract Hash Endian Mismatch between GetAppLog and GetContractState


devhawk:
I deployed a contract to test net. In the `getapplicationlog` response, the `result.executions.notifications` contains:

``` json
[
    {
        "contract": "0xfffdc93764dbaddd97c48f252a53ea4643faa3fd",
        "eventname": "Deploy",
        "state": {
            "type": "Array",
            "value": [
                {
                    "type": "ByteString",
                    "value": "tMUieFpmq0OGiYWowGJANoOfFxY="
                }
            ]
        }
    }
]
```

`tMUieFpmq0OGiYWowGJANoOfFxY=` is `0xB4C522785A66AB43868985A8C0624036839F1716` in hex. However, if I pass that hex value to GetContractState, I get a "contract not found" error. 

Via [dora](https://dora.coz.io/contract/neo3/mainnet/0x16179f83364062c0a885898643ab665a7822c5b4) I was able to determine that the *correct* contract hash is `0x16179f83364062c0a885898643ab665a7822c5b4`. the value returned from getapplicationlog needs to be byte swapped before passing it to getcontractstate.

is it possible to make these more consistent? It's bad enough that `getapplicationlog` uses base64 encoding while getcontractstate expects hex encoding. Do we have to force users to swap bytes too?

roman-khimov:
I don't think it's possible to fix this, at least not without actually fixing neo-project/neo#938. Notifications are `[name, stack item]` pairs in their essence (tied to `contract` of course) and stack items can contain whatever data can be on VM stack in exactly the way it's represented there. Contract hash is not much different from any other ByteString in VM, so it gets translated to base64 as regular ByteString and changing it doesn't seem to be correct, because in VM world that's how the hash is represented, this it what you pass to `Contract.Call` for example. Changing `getcontractstate` doesn't seem to be correct either as all other JSON communication implies reversed hash where we expect a hash to be passed (like `contract` field in this log).

devhawk:
Thanks for the link @roman-khimov. I knew that I had seen this issue somewhere before

PR #770, opened 2022-10-14
TokenId with Base64

Ashuaidehao:


roman-khimov:
Why do we need this incompatible change?

shargon (approved):

Liaojinghui:
I dont see why we need this change

shargon:
It's less amount of data to store, base64 is shorter than hex

Liaojinghui:
> incompatible

Indeed, but as @roman-khimov said,  this change is minor but incompatible. Does it worth it to have an incompatible update to save a few bytes?

Liaojinghui:
Well, actually this pr dont have to change [RpcMethod], right?

roman-khimov:
Storage-wise it's the same, we store raw id bytes, network-wise yes, there is some difference, but IDs are limited to 64 bytes, so we can save some tens (not even hundreds) of bytes, but we'll definitely break some applications. I'd rather leave it as is.

superboyiii:
I think the basic issue is all nep11 related RPC params in RpcServer return Base64 but TokensTracker still need hex input which is not uniform. For example: when someone tried to get `tokenid` list from invokefunction `tokensOf` method, he got Base64 `tokenId` but he can't push it directly into `getnep11properties` since it needs hex input. Less storage is not the point.

roman-khimov:
> someone tried to get tokenid list from invokefunction tokensOf method

Pretty similar to #609 situation.

Base64 is better in some aspects, that's true. The problem is that this change will break some existing code interacting with RPC servers. If we want to switch, we can make _an additional_ `tokenidb64` field and deprecate (but still return) the old one.

joeqian10:
Request:
```
{
  "jsonrpc": "2.0",
  "method": "getnep11properties",
  "params": ["0x9f344fe24c963d70f5dcf0cfdeb536dc9c0acb3a","ed00"],
  "id": 1
}
```

Response:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -2147024809,
        "message": "Unable to translate bytes [ED] at index 0 from specified code page to Unicode."
    }
}
```

Maybe using base64 encoded token id could avoid this kind of errors.

superboyiii:
> Request:
> 
> ```
> {
>   "jsonrpc": "2.0",
>   "method": "getnep11properties",
>   "params": ["0x9f344fe24c963d70f5dcf0cfdeb536dc9c0acb3a","ed00"],
>   "id": 1
> }
> ```
> 
> Response:
> 
> ```
> {
>     "jsonrpc": "2.0",
>     "id": 1,
>     "error": {
>         "code": -2147024809,
>         "message": "Unable to translate bytes [ED] at index 0 from specified code page to Unicode."
>     }
> }
> ```
> 
> Maybe using base64 encoded token id could avoid this kind of errors.

@roman-khimov @Liaojinghui We find this error when emit hex, it's time to change to base64.

roman-khimov:
> We find this error when emit hex, it's time to change to base64.

How is it related to hex/base64 question?
```
$ curl -d '{ "jsonrpc": "2.0", "id": 5, "method": "getnep11properties", "params": ["0x9f344fe24c963d70f5dcf0cfdeb536dc9c0acb3a","ed00"] }' https://rpc10.n3.nspcc.ru:10331
{"id":5,"jsonrpc":"2.0","result":{"name":"\ufffd\u0000","tokenURI":"ipfs.io/ipfs/bafybeies3gtnfneg5mez45em2nhwpfp3r7gzizlyz2sxtfvvxdeqyha7o4/237.json"}}
```

superboyiii:
> > We find this error when emit hex, it's time to change to base64.
> 
> How is it related to hex/base64 question?
> 
> ```
> $ curl -d '{ "jsonrpc": "2.0", "id": 5, "method": "getnep11properties", "params": ["0x9f344fe24c963d70f5dcf0cfdeb536dc9c0acb3a","ed00"] }' https://rpc10.n3.nspcc.ru:10331
> {"id":5,"jsonrpc":"2.0","result":{"name":"\ufffd\u0000","tokenURI":"ipfs.io/ipfs/bafybeies3gtnfneg5mez45em2nhwpfp3r7gzizlyz2sxtfvvxdeqyha7o4/237.json"}}
> ```
> 
> P.S. BTW, the token just seems to be broken to me. "\ufffd\u0000" is not a proper UTF-8 string and C# node barfs at it for a reason (see [neo-project/neo-vm#496](https://github.com/neo-project/neo-vm/issues/496) as well).

Yes, you're right, it's a UTF-8 issue but not hex/base64 issue.

Liaojinghui:
I will close this pr if no further claims to support it.

PR #774, opened 2022-11-29
Fix MaxBlockSystemFee

erikzhang:


src/DBFTPlugin/DBFTPlugin.cs
```diff
@@ -87,7 +88,12 @@ public void Start(Wallet wallet)
         private bool RemoteNode_MessageReceived(NeoSystem system, Message message)
         {
             if (message.Command == MessageCommand.Transaction)
-                consensus?.Tell(message.Payload);
+            {
+                Transaction tx = (Transaction)message.Payload;
+                if (tx.SystemFee > settings.MaxBlockSystemFee)
```
Liaojinghui, line 93: Only systemfee？
```suggestion
                if (tx.SystemFee+tx.NetworkFee > settings.MaxBlockSystemFee)
```

Liaojinghui, line 93: I have two questions here, 1. we can not access systemfee. 2. why not systemfee+networkfee
```suggestion
                if (tx.SystemFee + tx.NetworkFee > settings.MaxBlockSystemFee)
```

Liaojinghui (approved):

src/DBFTPlugin/DBFTPlugin.cs
roman-khimov, line 93: Does it change anything at all? Consensus process only cares about a specific list of transactions when it expects them in a particular stage of the process. `CheckPrepareResponse` does the overall check anyway for the consensus process. Then non-consensus nodes won't notice this change at all and could still have such a transaction in their mempools.

roman-khimov: I see this setting (`MaxBlockSystemFee`) is only in the dBFT plugin now in C# implementation, so ordinary nodes can't do anything wrt this limit.

roman-khimov:
The best way solve this is to move `MaxBlockSystemFee` to `ProtocolConfiguration` again (it was there way back when) and have all nodes do this check before accepting transaction into the memory pool. If we're not doing this (and keeping plugins/configs the way they are now), maybe this patch is the only way to at least make the consensus process work correctly.

dusmart:
> The best way solve this is to move `MaxBlockSystemFee` to `ProtocolConfiguration` again (it was there way back when) and have all nodes do this check before accepting transaction into the memory pool. If we're not doing this (and keeping plugins/configs the way they are now), maybe this patch is the only way to at least make the consensus process work correctly.

agree with roman-khimov

Free DOS still exists if we allow the large sysfee txs exist. Bad guys will pay nothing for it. Maybe checking it in tx's verification stage is better.

vncoelho (approved):
I see. Perhaps it should be here as you included, but also in the TCP P2P layer.

shargon (commented):
@superboyiii tested?

superboyiii:
> @superboyiii tested?

Yes, consensus throwed these over-MaxBlockSystemFee txs out of mempool. Ordinary nodes still keep them in their mempool. Anyway, it's better than nothing, at least ensured the safety of consensus. We can merge this and release v3.5.0 ASAP. Improvement can be done next time.

shargon (approved):

PR #777, opened 2022-12-10
Check NEP11 methods

Ashuaidehao:


roman-khimov:
Hi, neo-project/neo#1883. You can take a look at https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/smartcontract/manifest/standard#CheckABI as well, it has compliance checks implemented properly.

src/TokensTracker/Trackers/NEP-11/Nep11Tracker.cs
```diff
@@ -92,7 +92,18 @@ public override void OnPersist(NeoSystem system, Block block, DataCache snapshot
...
                         continue;
                     }
-
+                    if (balanceMethod != null &&
+                        !(balanceMethod.Parameters[0].Type == ContractParameterType.Hash160 && balanceMethod.ReturnType == ContractParameterType.Integer))
+                    {
+                        Log($"{state.Hash} is not nft:balance1!", LogLevel.Warning);
+                        continue;
+                    }
+                    if (balanceMethod2 != null &&
```
cschuchardt88, line 101: should be an `else if` Because the function is one or the other. Also if the token has both functions you don't check for that either. As the code sits right now, if 1st function exists look for other function. Which isn't valid to what NEP-11 says
https://github.com/neo-project/proposals/blob/master/nep-11.mediawiki#user-content-decimals

vncoelho (commented):
@Ashuaidehao, can you check @cschuchardt88 

Otherwise, @cschuchardt88, can you modify the aforementioned if statement and let's finish this PR?

cschuchardt88:
@Ashuaidehao @vncoelho

Basic Simple Check
---
```csharp
if (Decimals == 0)
{
  return BalanceOf(ownerAddress);
}
else
{
  return BalanceOf(ownerAddress, tokenId);
}
```

src/TokensTracker/Trackers/NEP-11/Nep11Tracker.cs
vncoelho: I think also balance1 and balance2 is a bad naming.
Let's say something like "balanceInteger" and "balanceDivisible"

cschuchardt88: Also there is no need to check to `abi` for these functions.

@vncoelho I think `balanceDivisible` and `balanceNonDivisible` would fit better.

Jim8y:
close as author agrees this is not necessary

Issue #736, opened 2022-06-24
Data missing for random blocks in StatesDumper


superboyiii:
**Describe the bug**
StatesDumper is useful for storage comparison espically for compatibility test. However, few data is often missing randomly when persisted. 

**To Reproduce**
1. Run two nodes, each installed StatesDumper, syncing the same network(mainnet/testnet)
2. Sync full data competely, then compare all json files in `Storage_xxxxxxx` in neo-cli

**Expected behavior**
Some data is missing.

**Screenshots**
![411e33fa8b027b5d18a26bd9a5fb424](https://user-images.githubusercontent.com/38493437/175479590-c04908c9-f767-4d00-9422-3db715955ce1.png)

**Platform:**
 - OS: [win/linux/mac]
 - Version [3.3.1]

superboyiii:
After my test, I found they will be lost when being interupted by `exit` or `ctrl + C`, see issue here: https://github.com/neo-project/neo-modules/issues/786

superboyiii:
I think the issue is because https://github.com/neo-project/neo-modules/pull/568#discussion_r1067930312

PR #789, opened 2023-01-18
Refac statedump

Ashuaidehao:
Close https://github.com/neo-project/neo-modules/issues/736

### DIFF
--- a/src/StatesDumper/Settings.cs
+++ b/src/StatesDumper/Settings.cs
@@ -1,4 +1,4 @@
-// Copyright (C) 2015-2021 The Neo Project.
+// Copyright (C) 2015-2023 The Neo Project.
 //
 // The Neo.Plugins.StatesDumper is free software distributed under the MIT software license,
 // see the accompanying file LICENSE in the main directory of the
--- a/src/StatesDumper/Settings.cs
+++ b/src/StatesDumper/Settings.cs
@@ -26,10 +26,6 @@ namespace Neo.Plugins
         /// </summary>
         public uint HeightToBegin { get; }
         /// <summary>
-        /// Height to begin real-time syncing and dumping on, consequently, dumping every block into a single files
-        /// </summary>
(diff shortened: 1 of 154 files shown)


src/StatesDumper/StatesDumper.cs
```diff
@@ -122,45 +129,63 @@ private void OnPersistStorage(uint network, DataCache snapshot)
...
         {
             //Default Parameter
             uint storagePerFolder = 100000;
-            uint folder = (((blockIndex - 1) / storagePerFolder) + 1) * storagePerFolder;
-            if (blockIndex == 0)
-                folder = 0;
-            string dirPathWithBlock = $"./Storage_{network:x8}/BlockStorage_{folder}";
-            Directory.CreateDirectory(dirPathWithBlock);
-            return dirPathWithBlock;
+            uint folder = (blockIndex / storagePerFolder) * storagePerFolder;
```
roman-khimov, line 186: This changes the naming scheme and will break scripts like https://github.com/nspcc-dev/neo-go/blob/master/scripts/compare-dumps/compare-dumps.go. Please keep the old scheme, I doubt changing it is important wrt to this PR goals.

Ashuaidehao: Sure, I'll make a new simple plugin so you won't change it.

src/StatesDumper/Settings.cs
```diff
@@ -26,10 +26,6 @@ internal class Settings
         /// </summary>
         public uint HeightToBegin { get; }
         /// <summary>
-        /// Height to begin real-time syncing and dumping on, consequently, dumping every block into a single files
-        /// </summary>
-        public int HeightToStartRealTimeSyncing { get; }
```
vncoelho, line 31: Why did you remove this configuration?
This can be useful when debugging or following the chain on real-time

Ashuaidehao:
Replaced by  https://github.com/neo-project/neo-modules/pull/790

Issue #716, opened 2022-05-22
Use a session protocol for some JSON-RPC methods 


devhawk:
Some types of remote operations would benefit from a stateful protocol such as WebSockets or [SignalR](https://docs.microsoft.com/en-us/aspnet/core/signalr/introduction). Examples:

* SendRawTransaction and wait for results
* Sever sends every new block that is minted (without needing to implement the entire P2P protocol)
* Iterate over the results InvokeScript/InvokeFunction that returns an iterator

PR #795, opened 2023-03-26
WSS: Add Web Socket Server

Liaojinghui:
Trying to add a wss plugin.

https://github.com/neo-project/neo/issues/2853 This is a good suggestion, maybe we can have `wss` supported


src/WebSocketServer/config.json
```diff
@@ -0,0 +1,17 @@
...
+      {
+        "Network": 860833102,
+        "BindAddress": "127.0.0.1",
+        "Port": 10334,
+        "SslCert": "",
+        "SslCertPassword": "",
+        "TrustedAuthorities": [],
+        "MaxConcurrentConnections": 40,
+        "SessionEnabled": false,
+        "SessionExpirationTime": 60
```
vncoelho, line 13: Allowed clients?
Maybe it will be useful to provide a list of allowed users that can connect. Or do you believe that the load will be reasonable and won't lock node?

Liaojinghui: to that i dont have a proper answer, will need well tested.

src/WebSocketServer/WebSocketServerPlugin.cs
```diff
@@ -0,0 +1,104 @@
...
+
+namespace Neo.Plugins.WebSocketServer;
+
+public class WebSocketServerPlugin : Plugin
+{
+    public override string Name => "NeoWebSocketServer";
+    public override string Description => "Enables WebSocket notifications for the node";
+
+    private Settings _settings;
+    private static readonly Dictionary<Guid, BlockWebSocketBehavior> Handlers = new();
```
cschuchardt88, line 19: I think you should be using ConcurrentDictionary, so you dont have to use Lock.

cschuchardt88:
is there a way for other plugins to add functionality to the wss like RPCMethhods works?

cschuchardt88:
This for runtime only correct? or can you go back in time. Like the place you disconnected. I didn't see anything in code and was just wondering if it was thought upon for maybe a future feature add-on. Other than that it looks good. Thanks for changing to `ConcurrentDictionary`. Locking can be a pain, I have ran into it too many times in the pass.

Liaojinghui:
@cschuchardt88 jsut made it runtime. can add other features little by little, otherwise this pr would be too complex to review.

Liaojinghui:
> is there a way for other plugins to add functionality to the wss like RPCMethhods works?

No, not going to make it work that way.

cschuchardt88:
#716 mentions  `SendRawTransaction` or `InvokeScript/InvokeFunction` for WSS. I didn't see anything in code. Thought you might be interested in adding.

Liaojinghui:
> #716 mentions `SendRawTransaction` or `InvokeScript/InvokeFunction` for WSS. I didn't see anything in code. Thought you might be interested in adding.

Well,,,,maybe i should combine rpc and wss plugins to add those rpc methods

cschuchardt88:
Why on u just import `RpcServer` plugin? Than you can just those methods. It would work the same way as `neo-express` .

Liaojinghui:
casue my first thought is  they should be seperate projects and does not rely on each other. Let's try to have your rest done first, rpc will be updated anyway.

cschuchardt88:
True, I was just thinking less code to maintain.

cschuchardt88:
@Liaojinghui 
I have a solution that only uses pure asp.net core methods just like rpc server and rest web server. It works the the exactly the same as RPC with the `RpcMethod` attributes. You just add `WebSocketMethod` attribute to the same methods. However i am use `newtonsoft` json serializers. I really think we need to switch core to use `newtonsoft` or add some kind of attribute like `newtonsoft` has for knowing whether to serialize the property or field. Check it out. If you like I can convert to `neo-cli` module and create PR and remove `newtonsoft` to use `neo.json` https://github.com/cschuchardt88/neo-express/blob/construct/v4.0.0/src/neoxp/Node/Modules/ExpressWebSocketServerPlugin.cs

Liaojinghui:
@cschuchardt88 its a good idea, we can discuss it later after the monorepo.

Liaojinghui:
use https://github.com/neo-project/neo-modules/pull/847 instead

PR #800, opened 2023-04-27
Add Oracle limit

shargon:


shargon:
@dotnet-policy-service agree

> @shargon please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @dotnet-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @dotnet-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @dotnet-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@roman-khimov

shargon:
@roman-khimov Please, could you take a look?

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -22,7 +22,11 @@ namespace Neo.Plugins
 {
     class OracleHttpsProtocol : IOracleProtocol
     {
-        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false });
+        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false })
+        {
+            Timeout = TimeSpan.FromSeconds(30),
```
roman-khimov, line 27: Already set in `Configure()`?

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -22,7 +22,11 @@ namespace Neo.Plugins
...
+            MaxResponseContentBufferSize = ushort.MaxValue
```
AnnaShaleva, line 28: Don't we want to make this value configurable?

roman-khimov: It's `OracleResponse.MaxResultSize` effectively. Maybe this constant can be used instead.

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: Done

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -22,7 +22,10 @@ namespace Neo.Plugins
 {
     class OracleHttpsProtocol : IOracleProtocol
     {
-        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false });
+        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false })
+        {
+            MaxResponseContentBufferSize = ushort.MaxValue
```
erikzhang, line 27:
```suggestion
            MaxResponseContentBufferSize = OracleResponse.MaxResultSize
```

erikzhang:
Is the response size of NeoFS also limited?

AnnaShaleva:
> Is the response size of NeoFS also limited?

It should be limited by the client, but currently it's not. So the problem exists for `OracleNeoFSProtocol` as far. The size of a single physical NeoFS object is limited by 64MB in mainnet\testnet, but there are large ("virtual") objects constructed from a set of small physical ones and these are almost unlimited.

The neofs-api-csharp Client is capable to handle such large "virtual" objects and doesn't restrict their maximum size:
https://github.com/neo-ngd/neofs-api-csharp/blob/82ca3de1d87c96a04bdf094c60c7c6039bbacf81/src/Neo.FileStorage.API/client/Client.Object.cs#L106-L123
`OracleNeoFSProtocol` uses both `client.GetObject` and `client.GetObjectPayloadRangeData`, reads object data from the stream and converts the result directly to string:
https://github.com/neo-project/neo-modules/blob/4e310685d777c4d2b550d583a640849a9eb82403/src/OracleService/Protocols/OracleNeoFSProtocol.cs#L97-L101
and
https://github.com/neo-project/neo-modules/blob/4e310685d777c4d2b550d583a640849a9eb82403/src/OracleService/Protocols/OracleNeoFSProtocol.cs#L103-L108

Ideally, to solve this problem we need a change in the [neofs-api-csharp Client's API](https://github.com/neo-ngd/neofs-api-csharp/blob/82ca3de1d87c96a04bdf094c60c7c6039bbacf81/src/Neo.FileStorage.API/client/Client.Object.cs#L94) to limit the amount of data that can be read from the stream. In neofs-sdk-go it is done by returning the `ObjectReader` structure so that it's the reader  responsibility to restrict the amount of data that can be read from the underlying transmission channel: https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.8/client#Client.ObjectGetInit.

Another way to solve this problem without neofs-api-csharp Client's API changes is to use [GetObjectPayloadRangeData](https://github.com/neo-ngd/neofs-api-csharp/blob/82ca3de1d87c96a04bdf094c60c7c6039bbacf81/src/Neo.FileStorage.API/client/Client.Object.cs#L362) combined with manually pre-checked range lenght on the `OracleNeoFSProtocol` side for both GetObjectRange and GetObject operations. However, this way requires to access and check object headers firstly and is a "cludgy" way in general.

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -22,7 +22,10 @@ namespace Neo.Plugins
 {
     class OracleHttpsProtocol : IOracleProtocol
     {
-        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false });
+        private readonly HttpClient client = new(new HttpClientHandler() { AllowAutoRedirect = false })
+        {
+            MaxResponseContentBufferSize = OracleResponse.MaxResultSize
```
AnnaShaleva, line 27: We should probably adjust the exception handler of Oracle service wrt this change. Currently if the responce content exceeds `OracleResponse.MaxResultSize` length, the exception will be thrown by `message.Content.ReadAsStringAsync(...)` method:
https://github.com/neo-project/neo-modules/blob/4e310685d777c4d2b550d583a640849a9eb82403/src/OracleService/Protocols/OracleHttpsProtocol.cs#L85
It will be catched by the following `catch` block:
https://github.com/neo-project/neo-modules/blob/4e310685d777c4d2b550d583a640849a9eb82403/src/OracleService/OracleService.cs#L361-L368
As a result, we'll have `OracleResponseCode.Error` code instead of `OracleResponseCode.ResponseTooLarge` which is more appropriate in this case. In neo-go we return `OracleResponseCode.ResponseTooLarge` in this case.

AnnaShaleva (commented):
We should also consider fixing OracleNeoFSProtocol wrt https://github.com/neo-project/neo-modules/pull/800#issuecomment-1531246072.

src/OracleService/Protocols/OracleHttpsProtocol.cs
shargon: Which exception is throw?

shargon: @superboyiii could you test it?

superboyiii: > @superboyiii could you test it?

Sure

superboyiii (changes requested):

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -22,7 +22,10 @@ namespace Neo.Plugins
...
+        };
```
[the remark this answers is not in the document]
AnnaShaleva: We can't just revert this, the whole point of this PR is to restrict the amount of data that can be read from the response stream to avoid OOM.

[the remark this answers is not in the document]
superboyiii: > We can't just revert this, the whole point of this PR is to restrict the amount of data that can be read from the response stream to avoid OOM.

OK. The problem here is that we can't achieve
```
            if (!message.Content.Headers.ContentLength.HasValue || message.Content.Headers.ContentLength > OracleResponse.MaxResultSize)
                return (OracleResponseCode.ResponseTooLarge, null);
``` 
because it will throw exception by `System` from .NET and it can't be seen since we con't `catch (Exception ex)`. Finally we will always get `OracleResponseCode.Timeout`. Any other good idea?
PS: We shouldn't return `OracleResponseCode.Timeout` as default when catch exception. It doesn't make sense, cofused me a lot. I think it should be `OracleResponseCode.rror`.

src/OracleService/Protocols/OracleHttpsProtocol.cs
superboyiii: > @superboyiii could you test it?

@shargon Due to my test result, the same as @AnnaShaleva said. We need solve it.

src/OracleService/Protocols/OracleHttpsProtocol.cs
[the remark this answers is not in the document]
superboyiii: > Oh, I missed that commit. Can we test the resulting code?

Yes, already tested it before. The result here is my test result. It works well.
https://github.com/neo-project/neo-modules/pull/800#discussion_r1193521163

superboyiii (approved):

superboyiii (approved):

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -82,6 +82,8 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
                 return (OracleResponseCode.Error, message.StatusCode.ToString());
             if (!Settings.Default.AllowedContentTypes.Contains(message.Content.Headers.ContentType.MediaType))
                 return (OracleResponseCode.ContentTypeNotSupported, null);
+            if (!message.Content.Headers.ContentLength.HasValue || message.Content.Headers.ContentLength > OracleResponse.MaxResultSize)
```
roman-khimov, line 85: The `!message.Content.Headers.ContentLength.HasValue` part of it may affect legitimate responses that fit the limit, but don't have the header for whatever reason (dynamically generated content, for example).

shargon: We can require this header

roman-khimov: That'd be an artificial limitation to me. It shouldn't be hard to handle `!message.Content.Headers.ContentLength.HasValue` case as well, NeoGo handles both cases easily.

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -82,7 +83,26 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
                 return (OracleResponseCode.Error, message.StatusCode.ToString());
             if (!Settings.Default.AllowedContentTypes.Contains(message.Content.Headers.ContentType.MediaType))
                 return (OracleResponseCode.ContentTypeNotSupported, null);
-            return (OracleResponseCode.Success, await message.Content.ReadAsStringAsync(cancellation));
+            if (message.Content.Headers.ContentLength.HasValue && message.Content.Headers.ContentLength > OracleResponse.MaxResultSize)
+                return (OracleResponseCode.ResponseTooLarge, null);
+
+            byte[] buffer = new byte[OracleResponse.MaxResultSize];
+            var stream = message.Content.ReadAsStream(cancellation);
+            var read = await stream.ReadAsync(buffer, 0, buffer.Length, cancellation);
```
AnnaShaleva, line 91: If response stream contains more than `buffer.Length` bytes, then it's an error and we need to return `OracleResponseCode.ResponseTooLarge` code. And if I'm not mistaken, the current code will just stop reading and return a part of the response with `OracleResponseCode.Success` code (which isn't the desired behaviour).

AnnaShaleva: Please, check the way how it's done in neo-go: https://github.com/nspcc-dev/neo-go/blob/7b86a54fc0dd918a2514b0108be986a6e2cc63fa/pkg/services/oracle/response.go#L71-L82

shargon: Please, check it again

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -82,7 +83,29 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
...
+
+        private static Encoding GetEncoding(HttpContentHeaders headers)
+        {
+            Encoding encoding = null;
+            if ((headers.ContentType != null) && (headers.ContentType.CharSet != null))
+            {
+                encoding = Encoding.GetEncoding(headers.ContentType.CharSet);
+            }
+
+            return encoding ?? Encoding.UTF8;
```
AnnaShaleva, line 108: Not sure, but shouldn't we *always* enforce strict UTF8 as it is done for OracleNeoFSProtocol?
https://github.com/neo-project/neo-modules/blob/b639e84c8fa7aa33c063ba764b27b9f93ef0ef04/src/OracleService/Protocols/OracleNeoFSProtocol.cs#L100

AnnaShaleva (commented):
@shargon, could you also consider fixing the same problem for OracleNeoFSProtocol as @ZhangTao1596 suggested? https://github.com/neo-project/neo-modules/pull/800#issuecomment-1545358491

src/OracleService/Protocols/OracleHttpsProtocol.cs
AnnaShaleva: This part LGTM, resolved.

src/OracleService/Protocols/OracleHttpsProtocol.cs
AnnaShaleva: @roman-khimov, what do you think? In neo-go it's always utf-8: https://github.com/nspcc-dev/neo-go/blob/aca12b58c00410d765f205b752afdc716dd9fade/pkg/services/oracle/response.go#L76

shargon: Before it worked like this, I think that it's better

roman-khimov: I don't think it ever worked with non-UTF-8:
https://github.com/neo-project/neo-modules/blob/b639e84c8fa7aa33c063ba764b27b9f93ef0ef04/src/OracleService/OracleService.cs#L508-L513
`JToken.Parse` also insists on UTF-8.

AnnaShaleva: Also, it would be nice to make it symmetrical with OracleNeoFSProtocol (it always enforses UTF-8).

shargon: What is the error that do you think that we should throw in that case?

AnnaShaleva (approved):

src/OracleService/Protocols/OracleHttpsProtocol.cs
AnnaShaleva: We should probably return just an `OracleResponseCode.Error` (0xff) in this case, because `OracleResponseCode.ContentTypeNotSupported` is returned when response content type doesn't match the configured one.

src/OracleService/Protocols/OracleHttpsProtocol.cs
```diff
@@ -94,7 +94,10 @@ public async Task<(OracleResponseCode, string)> ProcessAsync(Uri uri, Cancellati
                 return (OracleResponseCode.ResponseTooLarge, null);
 
             var encoding = GetEncoding(message.Content.Headers);
-            return (OracleResponseCode.Success, encoding.GetString(buffer, 0, read));
+            if (!encoding.Equals(Encoding.UTF8))
+                return (OracleResponseCode.ContentTypeNotSupported, null);
```
AnnaShaleva, line 98:
```suggestion
                return (OracleResponseCode.Error, null);
```

roman-khimov: Apply this one and merge? Anything else holding us on this PR?

AnnaShaleva: Everything else LGTM (except this comment).

src/OracleService/Protocols/OracleHttpsProtocol.cs
AnnaShaleva: Resolved.

AnnaShaleva (approved):
LGTM!

shargon:

Liaojinghui (approved):

superboyiii:
> @superboyiii could you test it?

I'm ing...

superboyiii (approved):
Request: https://api.github.com/users?since=800&per_page=1000

Works as expected!

PR #801, opened 2023-05-15
dbft: don't rush to send PrepareRequest when recovery

### LINKED ISSUE #799: dBFT: Recovery message triggers PrepareRequest from Primary even if the timer isn't fired yet
**Describe the bug**
See the https://github.com/nspcc-dev/dbft/issues/74, please. The same bug exists in C# implementation: https://github.com/neo-project/neo-modules/blob/4e310685d777c4d2b550d583a640849a9eb82403/src/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs#L251-L252

**Expected behavior**
Even on Recovery message receival, the primary node should wait until the timer fires before the subsequent PrepareRequest creation.

**Platform:**
 - Version: dBFT Plugin 3.5.0

ZhangTao1596:
Fix #799

### DIFF
--- a/src/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
+++ b/src/DBFTPlugin/Consensus/ConsensusService.OnMessage.cs
@@ -248,8 +248,6 @@ namespace Neo.Consensus
                             totalPrepReq = 1;
                             if (ReverifyAndProcessPayload(prepareRequestPayload)) validPrepReq++;
                         }
-                        else if (context.IsPrimary)
-                            SendPrepareRequest();
                     }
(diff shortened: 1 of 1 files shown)


AnnaShaleva (approved):

vncoelho (approved):

superboyiii (approved):
Set one CN to resend recovery message every 2 seconds, `MillisecondsPerBlock` set to 2000 as well and watch the result.
Before:

After fixing:
60dc97a1318119688b3d4765606d72d
It works effectively.

shargon (approved):

PR #805, opened 2023-07-17
rpc: add FindStorage

### LINKED ISSUE #758: Simplify getting latest state with GetState() or FindStates()
Assume I want to retrieve the latest state. The current flow to do so is
1. call `GetStateheight()` to get the root index
2. call `GetStateRoot(index)` to get the `stateroot`
3. call `GetState()` or `FindState()` with the obtained `stateroot`

That's 2 RPC calls before I can call `GetState` or `FindState`. RPC calls can be slow and it regularly happens that I get an exception from the code below while using the [contract download](https://github.com/neo-project/neo-express/pull/209) feature of neo-express simply because the `CurrentLocalRootHash` of the node has already changed before I could finally submit what I obtained as the latest root hash.

https://github.com/neo-project/neo-modules/blob/7db1c7956ac68758793a6ea30b5329cceb6ab1bc/src/StateService/StatePlugin.cs#L294-L298

One way to solve this is selecting a node with `FullState` enabled, but that severely limits the available RPC nodes to choose from while we're not even interested in the full state. 

Instead I want to propose accepting `null`  as first argument to `GetState` and `FindState` and automatically replace it with the latest root hash. e.g.

```csharp

     UInt256.TryParse(_params[0].AsString(), root_hash); 
     if (root_hash is null) {
         root_hash = StateStore.Singleton.CurrentLocalRootHash;
     } else if (!Settings.Default.FullState && StateStore.Singleton.CurrentLocalRootHash != root_hash) 
         throw new RpcException(-100, "Old state not supported");
```

Note: It's up for discussion if `GetProof()` should also support this. 

- Other: StateService plugin

ixje:
close #758 

Arguments:
1. Similar to `GetStorage()` the first argument is the `contract hash` or `contract id`
2. `search prefix`, base 64 encoded. Can be set to `""` to return all storage.
3. `start` location
    The `pageSize` is currently fixed to 50. If the find results exceed 50 results it shall return the `truncated` key set to `true` and the `next` key set to the `start` location of the next page. This way a consumer can directly use the result of `json["next"]` as 3rd parameter to continue where left off.

Up for discussion:
1. Should the `pageSize` be configurable in the RpcServer config?
2. Should the `pageSize` be configurable as parameter by the invoker?

### DIFF
--- a/src/RpcServer/RpcServer.Blockchain.cs
+++ b/src/RpcServer/RpcServer.Blockchain.cs
@@ -233,7 +233,6 @@ namespace Neo.Plugins
                         break;
                     }
 
-
                     JObject j = new();
                     j["key"] = Convert.ToBase64String(iter.Current.Key.Key.Span);
                     j["value"] = Convert.ToBase64String(iter.Current.Value.Value.Span);
@@ -241,7 +240,6 @@ namespace Neo.Plugins
                     i++;
                 }
                 json["truncated"] = hasMore;
-
             }
 
             json["next"] = start + i;
--- a/src/RpcServer/RpcServer.Blockchain.cs
+++ b/src/RpcServer/RpcServer.Blockchain.cs
@@ -198,6 +198,58 @@ namespace Neo.Plugins
         }
 
         [RpcMethod]
+        protected virtual JToken FindStorage(JArray _params)
+        {
+            using var snapshot = system.GetSnapshot();
+            if (!int.TryParse(_params[0].AsString(), out int id))
+            {
+                UInt160 hash = UInt160.Parse(_params[0].AsString());
+                ContractState contract = NativeContract.ContractManagement.GetContract(snapshot, hash);
+                if (contract is null) throw new RpcException(-100, "Unknown contract");
+                id = contract.Id;
+            }
+
+            byte[] prefix = Convert.FromBase64String(_params[1].AsString());
+            byte[] prefix_key = StorageKey.CreateSearchPrefix(id, prefix);
+
+            if (!int.TryParse(_params[2].AsString(), out int start))
+            {
+                start = 0;
+            }
+
+            JObject json = new();
+            JArray jarr = new();
+            int pageSize = 50;
+            int i = 0;
+
+            using (var iter = snapshot.Find(prefix_key).Skip(count: start).GetEnumerator())
+            {
+                var hasMore = false;
+                while (iter.MoveNext())
+                {
+                    if (i == pageSize)
+                    {
+                        hasMore = true;
+                        break;
+                    }
+
+
+                    JObject j = new();
+                    j["key"] = Convert.ToBase64String(iter.Current.Key.Key.Span);
+                    j["value"] = Convert.ToBase64String(iter.Current.Value.Value.Span);
+                    jarr.Add(j);
+                    i++;
+                }
+                json["truncated"] = hasMore;
+
+            }
+
+            json["next"] = start + i;
+            json["results"] = jarr;
+            return json;
+        }
+
+        [RpcMethod]
         protected virtual JToken GetTransactionHeight(JArray _params)
         {
             UInt256 hash = UInt256.Parse(_params[0].AsString());
--- a/src/RpcServer/RpcServer.Blockchain.cs
+++ b/src/RpcServer/RpcServer.Blockchain.cs
@@ -219,7 +219,7 @@ namespace Neo.Plugins
 
             JObject json = new();
             JArray jarr = new();
-            int pageSize = 50;
+            int pageSize = settings.FindStoragePageSize;
             int i = 0;
 
             using (var iter = snapshot.Find(prefix_key).Skip(count: start).GetEnumerator())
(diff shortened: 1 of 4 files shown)


shargon:

ixje:
> @ixje I vote for 1 option

They're not mutually exclusive. The server could set a maximum and the user could consume it at the maximum size or smaller. The question was whether we should support that or leave it hardcoded as is?

roman-khimov:
Just FYI, a paging extension to getnepXXtransfers in NeoGo:
https://github.com/nspcc-dev/neo-go/blob/master/docs/rpc.md#limits-and-paging-for-getnep11transfers-and-getnep17transfers

superboyiii:
> > @ixje I vote for 1 option
> 
> They're not mutually exclusive. The server could set a maximum and the user could consume it at the maximum size or smaller. The question was whether we should support that or leave it hardcoded as is?

I think it should be set in config.json so can be more flexible.

ixje:
> > > @ixje I vote for 1 option
> > 
> > 
> > They're not mutually exclusive. The server could set a maximum and the user could consume it at the maximum size or smaller. The question was whether we should support that or leave it hardcoded as is?
> 
> I think it should be set in config.json so can be more flexible.

@superboyiii I change it so the value is configurable in the config only. Let me know if we want to add user control up to the configured maximum as well

src/RpcServer/RpcServer.csproj
```diff
@@ -3,6 +3,8 @@
   <PropertyGroup>
     <PackageId>Neo.Plugins.RpcServer</PackageId>
     <RootNamespace>Neo.Plugins</RootNamespace>
+    <TargetFramework>net6.0</TargetFramework>
+    <LangVersion>10</LangVersion>
```
superboyiii, line 7: Why downgrade to .net6.0?

ixje: These changes should not have been included. I've reverted them.

superboyiii (approved):
OK. Now everything is OK for me.

ixje:
> OK. Now everything is OK for me.

@shargon ☝️

cschuchardt88:
@ixje I think pageSize should configurable in config.json. Or event have this feature be able to be disabled. Reason being is that with tons of Users requesting to find storage it could slow down the server or timeout requests.

ixje:
> @ixje I think pageSize should configurable in config.json. Or event have this feature be able to be disabled. Reason being is that with tons of Users requesting to find storage it could slow down the server or timeout requests.

It is now configurable in the config. Completely disabling can be done through the config by adding the method to the `DisabledMethods` list.

superboyiii:
@shargon Review again please.

src/RpcServer/RpcServer.Blockchain.cs
```diff
@@ -197,6 +197,58 @@ protected virtual JToken GetStorage(JArray _params)
...
 
+        [RpcMethod]
+        protected virtual JToken FindStorage(JArray _params)
+        {
+            using var snapshot = system.GetSnapshot();
+            if (!int.TryParse(_params[0].AsString(), out int id))
+            {
+                UInt160 hash = UInt160.Parse(_params[0].AsString());
+                ContractState contract = NativeContract.ContractManagement.GetContract(snapshot, hash);
+                if (contract is null) throw new RpcException(-100, "Unknown contract");
```
shargon, line 208: If the id is int and invalid contract, we don't get this error

ixje: @shargon this is literally a copy/paste from the existing `GetStorage()` implementation
https://github.com/neo-project/neo-modules/blob/99ffc846e4df93491d1d812cb4c6ef6ee875e056/src/RpcServer/RpcServer.Blockchain.cs#L180-L189

Ultimately it will not find any storage. What do you want it to be?

cschuchardt88: If _params[0] is "blah" the application crashes. Cause of this line 185
```csharp
UInt160 hash = UInt160.Parse(_params[0].AsString());
```
crashes the function do to no checking Use UInt160.TryParse

ixje: > If _params[0] is "blah" the application crashes. Cause of this line 185
> 
> ```cs
> UInt160 hash = UInt160.Parse(_params[0].AsString());
> ```
> 
> crashes the function do to no checking Use UInt160.TryParse

This pattern is in nearly all current RpcMethod's, just look at for example `GetBlock`, `GetBlockHash`, `GetBlockHeader` or `GetRawTransaction`. I'm not saying I disagree this could be handled better, but I'm simply sticking to the current style already in the rest of the RpcServer code unless some maintainer tells me otherwise. I prefer to have this method included in the 3.6.0 release and see more graceful handling be done in a a separate PR for all methods. Could probably be done at the same time as updating it to follow https://github.com/neo-project/proposals/pull/156

cschuchardt88: I understand completely. Not saying your wrong; you bring up a good point, but the way I look at it is that RPC Methods can be in any plugin. For the other plugins they could follow this convention or not. Just saying....

Liaojinghui (dismissed):

Issue #665, opened 2021-10-26
Rename vmstate to state in getapplicationlog result 


ixje:
**Summary or problem description**
Calling `invokefunction`, `invokescript` and `invokecontractverify` return among others these properties
```json
        "state": "HALT",
        "gasconsumed": "999972",
        "exception": null,
        "stack": [
```
`getapplicationlog` has almost the same response, but instead uses `vmstate` instead of `state` as the key name for the engine state. Now a consumer has to add special handling just for `getapplicationlog` while it can all still be considered execution results.

**Do you have any solution you want to propose?**
I'd like to propose to streamline the naming to `state`. We can first add `state` as another key, mention `vmstate` is deprecated and will be removed in the release after the one the new key is introduced in to avoid breaking existing tools like the `getversion` change did. 

**Where in the software does this update applies to?**
- RPC (HTTP) specifically [here](https://github.com/neo-project/neo-modules/blob/575efd238bf0bac87e4cef390b41c95ba6c07f3f/src/ApplicationLogs/LogReader.cs#L69)

PR #380, opened 2020-10-29
trigger type for getapplicationlog

chenzhitong:

PR #807, opened 2023-08-02
ApplicationLog: New storage implementation for expandability

cschuchardt88:
With this new implementation of "ApplicationLog" you will be easier to expand the "ApplicationLog" for use in neo-node-cli and RPC. Tested with my xunit tests locally. 

_Note that you need to resync the whole blockchain for this new version of "ApplicationLog"_

**Change Log**
- New Storage implementation
- Added new console commands to neo-cli
- Added lookups by block, transaction and contract in neo-cli
- Added Sort methods for by event name

**Storage Size comparison to blockchain**
| Sizes | BlockChain | Application |
| ----   | ------------ | ------------- |
| 2.30GB      | Testnet T5 | Blockchain |
| 2.41GB      | Testnet T5 | ApplicationLog |

### Now you can use ApplicationLog in neo-cli here are the commands

image

### **Output of Command Console for ApplicationLog**

image

image

image

### DIFF
--- a/src/RpcServer/RpcServer.csproj
+++ b/src/RpcServer/RpcServer.csproj
@@ -4,4 +4,7 @@
     <PackageId>Neo.Plugins.RpcServer</PackageId>
   </PropertyGroup>
 
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
+  </ItemGroup>
 </Project>


cschuchardt88:
@dotnet-policy-service agree

tests/Neo.Network.RPC.Tests/RpcTestCases.json
```diff
@@ -3050,166 +3050,46 @@
...
-              }
-            ]
-          },
-          {
-            "trigger": "PostPersist",
+            "trigger": "Application",
             "vmstate": "HALT",
-            "gasconsumed": "2031260",
-            "exception": null,
+            "gasconsumed": 997778,
```
roman-khimov, line 3065: See how it changes the output here.

src/RpcClient/Models/RpcApplicationLog.cs
```diff
@@ -64,10 +64,10 @@ public class Execution
         public JObject ToJson()
         {
             JObject json = new();
+            json["exception"] = ExceptionMessage;
             json["trigger"] = Trigger;
             json["vmstate"] = VMState;
-            json["gasconsumed"] = GasConsumed.ToString();
-            json["exception"] = ExceptionMessage;
+            json["gasconsumed"] = GasConsumed;
```
roman-khimov, line 70: This is an incompatible change, please don't do this.

tests/Neo.Network.RPC.Tests/RpcTestCases.json
```diff
@@ -3050,166 +3050,46 @@
...
-      "method": "getapplicationlog",
-      "params": [ "0x6ea186fe714b8168ede3b78461db8945c06d867da649852352dbe7cbf1ba3724", "OnPersist" ],
-      "id": 1
-    },
-    "Response": {
-      "jsonrpc": "2.0",
-      "id": 1,
-      "result": {
-        "blockhash": "0x6ea186fe714b8168ede3b78461db8945c06d867da649852352dbe7cbf1ba3724",
-        "executions": [
```
roman-khimov, line 3159: Ah, now I see there are even more of them. Please, do not change the API. Optimizations are OK, but we need a stable API. See #380 also.

cschuchardt88:
Ok fixed RPC schema. Do note for this to work you will have to resync node. I binary serialize it, and have it indexed, so if you were to do a search on a contract you will get newest to oldest and for transaction events you will get order executed.

cschuchardt88:
Question at
https://github.com/neo-project/neo-modules/blob/99ffc846e4df93491d1d812cb4c6ef6ee875e056/src/ApplicationLogs/LogReader.cs#L121

is implemented for storage but is not for JSON RPC. Would you like me to implemented this in the code? Another function or different parameters? would be kinda hard to determine the UInt256 if it was blockhash or txhash, unless we block index for parameter.

cschuchardt88:
@roman-khimov ignore last two comments, i change the way data is stored, that's it. implemented the new storage into the JSON RPC function. JSON RPC schema is the same as before. Works with both transactions and blocks, checked against production's schema and it matches perfectly.

cschuchardt88:
With this new implementation of "ApplicationLog" you will be easier to expand the "ApplicationLog" for use in neo-node-cli and RPC. Tested with my xunit tests locally.

**_Expandability_**
- For example "GetBlockLog" and "GetTransactionLog" is implemented. The functionally for searching "NotifyEventArgs" for all Contracts is available, just no models and/or functions exists, for example "GetContractLog". Note that all data is being captured and stored and indexed, for Block, Transaction, and Contract Logs with data only available from "Blockchain.Committing" and "Blockchain.Committed" events.
- We can now easily add functionally to neo's node-cli for querying logs right from the console now. Whether that be Block, Transaction, Contract logs.
- RPC can easily be expanded as well. For example "GetContractLog" (Note: just needs to be implemented in code.)

cschuchardt88:
### Now you can use GetApplicationLog in neo-cli here are the commands

image

### **Output of Command Console for GetApplicationLog**

image

image

image

Liaojinghui:
what `vm` in your command means?

cschuchardt88:
@Liaojinghui means execution of the vm and event and/or notifylog

Liaojinghui:
`vm` is not an intuitive  name, how about `log`

cschuchardt88:
@Liaojinghui ok made the changes you requested. I'm not good with names, so thank you.

Liaojinghui:
Change existing logic should be something to avoid as much as we can, it might cause inconsistancy and break applications that rely on it.

shargon:

cschuchardt88:
Should issue #665 be added to this PR?

roman-khimov:
It's a very different problem and problems better be solved one by one.

BTW, `getnotificationlog` worries me here. I'm OK with just about any internal storage changes for existing APIs, but a new API needs to be discussed additionally. Not that it's bad, I'm just not sure it's very useful (we've worked for a long long time without it) and it requires some additional storage.

cschuchardt88:
I figured it would be nice addition to add since we have the information. The reason for me adding it is, some people including me just want to get events on the blockchain for a certain contract and we don't know the block hash or index or even a transaction hash. I have jobs running that look for certain events on my contracts to fire than an action to do something i.e. an external application. I can remove it if need be. But instead of adding other plugins with similar storage data requirements. I added it in here. Most plugins don't have external API besides RPC which is limited. No sense in storing the same data again.

superboyiii:
Very great feature. I will test it on testnet first!

Liaojinghui (dismissed):

Liaojinghui:
> Very great feature. I will test it on testnet first!

@superboyiii Any update?

src/ApplicationLogs/LogReader.cs
```diff
@@ -51,161 +65,343 @@ protected override void OnSystemLoaded(NeoSystem system)
...
+        {
+            var txContract = string.IsNullOrEmpty(eventName) ?
+                _neostore.GetContractLog(scripthash, TriggerType.Application, page, pageSize) :
+                _neostore.GetContractLog(scripthash, TriggerType.Application, eventName, page, pageSize);
+
+            if (txContract.Count == 0)
+                ConsoleHelper.Error($"No notify logs.");
+            else
+                PrintEventModelToConsole(txContract);
         }
```
superboyiii, line 205: I like this very much since it's convenient for developer to quick scan specific event log. Most of the function works well, but still somewhere has issues, for example:

```
neo> log contract 0xf0151f528127558851b39c2cd8aa47da7418ab28 0 3
Transaction Hash: 0x16890a963a9325bcce5fa874840ee9385fb52a81b08ba87a90cc0881842b03ad

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"Llmh1poEZKFMPOkBa0p1QYFoqys="}
    arg2: {"type":"ByteString","value":"352hGohJmxzp1lmMC322MRwTfag="}
    arg3: {"type":"Integer","value":"3339600000000"}
--------------------------------
Transaction Hash: 0x5f577775c8980b9d3f17e0c241a77c19232e39ff973fad9a4e3c854074c0f657

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"BrY0XpYy7nFjXOnPMsMBAxx5HRc="}
    arg2: {"type":"ByteString","value":"YeKHeeCXGEps8sSRlxbc4v97b1U="}
    arg3: {"type":"Integer","value":"128176221000"}
--------------------------------
neo> log contract 0xf0151f528127558851b39c2cd8aa47da7418ab28 1 3
Transaction Hash: 0x16890a963a9325bcce5fa874840ee9385fb52a81b08ba87a90cc0881842b03ad

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"Llmh1poEZKFMPOkBa0p1QYFoqys="}
    arg2: {"type":"ByteString","value":"352hGohJmxzp1lmMC322MRwTfag="}
    arg3: {"type":"Integer","value":"3339600000000"}
--------------------------------
Transaction Hash: 0x5f577775c8980b9d3f17e0c241a77c19232e39ff973fad9a4e3c854074c0f657

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"BrY0XpYy7nFjXOnPMsMBAxx5HRc="}
    arg2: {"type":"ByteString","value":"YeKHeeCXGEps8sSRlxbc4v97b1U="}
    arg3: {"type":"Integer","value":"128176221000"}
--------------------------------
Transaction Hash: 0x72426a7a73e5c9389d6385b5b1c6871558025e066eeda8cab1b21aeb59193257

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"qkj3lEssPRvcRa9lNHANUgZmNbI="}
    arg2: {"type":"ByteString","value":"Llmh1poEZKFMPOkBa0p1QYFoqys="}
    arg3: {"type":"Integer","value":"3339600000000"}
--------------------------------
```
I saw the default page index is `1`, but `0` still works. Maybe start from 0 makes more sense? And here I asked for 3 results in `size` param, but it returns `2` results at the beginning and returns `3` when I try for the second time. 
I even meet a situation like:
```
neo> log contract 0xf0151f528127558851b39c2cd8aa47da7418ab28 0 0
Error: No notify logs.
neo> log contract 0xf0151f528127558851b39c2cd8aa47da7418ab28 0 1
Error: No notify logs.
neo> log contract 0xf0151f528127558851b39c2cd8aa47da7418ab28 0 2
Transaction Hash: 0xd5756860d363680da5878f4f2bc5a5e8dd0d76bf5ab5cf8e570e5daf43a3cddd

  Event Name:  Transfer
  State Parameters:
    arg1: {"type":"ByteString","value":"uYxb4QZ8TaZy6W5rOQiLkLldUp4="}
    arg2: {"type":"ByteString","value":"oyJrAPWr9bXxnMN64ykd4lk6YKk="}
    arg3: {"type":"Integer","value":"18480412641180"}
--------------------------------
```

cschuchardt88: This is because
```csharp
foreach (var (key, value) in _snapshot.Seek(prefixKey, SeekDirection.Backward)) // Get newest to oldest
{
    if (key.AsSpan().StartsWith(prefix))
    {
        var state = value.AsSerializable<ContractLogState>();
        if (state.Trigger == trigger)
        {
            if (index >= page && index < (pageSize + page)) // here this logic.
                yield return state;
            index++;
        }
    }
    else
        yield break;
}
```

But im fixing it to not allow 0 index for page and pagesize of 0

superboyiii:
@cschuchardt88 How about support `log block <blockhashOrIndex> [eventName=null]`, this could be more convenient?

cschuchardt88:
Currently the store i created doesn't not store Indexes, But i guess i could query hash than use the hash for my store. I'll look into it.

superboyiii:
> Currently the store i created doesn't not store Indexes, But i guess i could query hash than use the hash for my store. I'll look into it.

Maybe just like https://github.com/neo-project/neo-modules/blob/2d7fb2778ca7ad11815e438534916103859e36ac/src/RpcServer/RpcServer.Blockchain.cs#L76 to get BlockHash?

Liaojinghui:
@shargon @roman-khimov

roman-khimov:
@Liaojinghui, https://github.com/neo-project/neo-modules/pull/807#issuecomment-1697742736 And we've only got worse wrt this. Each logical change should be tracked and evaluated separately. Right now we have too many of them here to me. I'd start with storage schema change (that is strictly something not affecting any of external interfaces, pure optimization), then any addons (and CLI better be split from RPC ones). It'd be much easier to review, track and communicate to the outside world then (docs/changelogs).

cschuchardt88:
@superboyiii for https://github.com/neo-project/neo-modules/pull/807#issuecomment-1697742736 is commenting the code out ok? or you want it removed completely?

cschuchardt88:
This update may break `neo-express`

src/ApplicationLogs/Store/LogStorageStore.cs
```diff
@@ -0,0 +1,373 @@
...
+                .Add(id.ToByteArray())
+                .ToArray();
+            try
+            {
+                _snapshot.Put(key, BinarySerializer.Serialize(stackItem, (uint)Settings.Default.MaxStackSize));
+            }
+            catch (NotSupportedException)
+            {
+                _snapshot.Put(key, BinarySerializer.Serialize(StackItem.Null, (uint)Settings.Default.MaxStackSize));
+            }
```
superboyiii, line 146: `Now` changed to `public static byte[] Serialize(StackItem item, ExecutionEngineLimits limits)` in https://github.com/neo-project/neo/pull/2948

cschuchardt88:
LogEventArgs are now working

image

From Testnet T5
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0x49e493cee3798bd0e6f2e745ac75291534b290fa4033ac36bed9204d429ead4c",
        "executions": [
            {
                "trigger": "Application",
                "vmstate": "HALT",
                "exception": null,
                "gasconsumed": "264528",
                "stack": [
                    {
                        "type": "ByteString",
                        "value": "SGVsbG8sIGJvYg=="
                    }
                ],
                "notifications": [],
                "logs": [
                    {
                        "contract": "0xa6cd0a163167fc37b24cc3b226342474e853a30a",
                        "message": "Hello, bob"
                    }
                ]
            }
        ]
    }
}
```

cschuchardt88:
@superboyiii @Liaojinghui maybe you can help me with a problem. Trying to get all `LogEventArgs` for a transaction. But I get duplicates because the node is sending that transaction and than processing the transaction again once processed by the blockchain. Inturn emits `LogEventArgs` twice. Maybe a Issue to add `LogEventArgs` to `Blockchain.ApplicationExecuted`?

Liaojinghui:
I have considered this issue, but solution would change the core, so i suggest to keep it this way for a while. We can fix it after the monorrpo merge.

cschuchardt88:
Let me know when the new release of neo-core is here in this repo so i can update

> `Now` changed to `public static byte[] Serialize(StackItem item, ExecutionEngineLimits limits)` in [neo-project/neo#2948](https://github.com/neo-project/neo/pull/2948)

superboyiii (approved):
Tested. It's good for me now.

Liaojinghui:
@shargon

superboyiii:
Could you have a look? @shargon

shargon:
> Could you have a look? @shargon

Yes, but I need more time

cschuchardt88:
Any update @shargon on this?

src/ApplicationLogs/Store/States/ContractLogState.cs
```diff
@@ -0,0 +1,73 @@
...
+            };
+
+        #region ISerializable
+
+        public override int Size =>
+            TransactionHash.Size +
+            sizeof(byte) +
+            base.Size;
+
+        public override void Deserialize(ref MemoryReader reader)
```
shargon, line 40: `Deserialize` content **always** come from a `Serialize` output?

cschuchardt88: I'm overriding the `base` class of `NotifyLogState` in turn inherits `ISerialize`

cschuchardt88: We initially create it 1st than serialize it, Hints the `Create` `static` method, above. But to answer question yes?

cschuchardt88: @shargon any news?

shargon: I need more time sorry

PR #810, opened 2023-08-16
Add Rest Web Server `WIP`

cschuchardt88:
Adding Restful services to Neo N3 ecosystem. This uses `MVC` model without the _view_ aspect. I currently don't know how much easier I can make this plugin. When it comes to flexibility and freedom. This system was design with usability. To allow the freedom of making what you want with plugins and will have a fault tolerant model. Also this model can allow upgrades, if Neo ever decides to make `neo-cli` have a built in webserver that handles `RPC`, `Rest`, `WSS`; this would be the code to do so. The already built plugins that are using this RestServer Plugin would not need to be changed.

This plugin is `test` ready. I'm not trying to rush it and understand it takes a lot of time. But I would love some feedback on this plugin. If you decide to not use this plugin. By all means this plugin will not end here.

The goal here is to have its client be dependency free of the `Neo.dll` and all its dependencies. Of course you can still add them if need be. Like `Neo.vm.dll` for building scripts to send via `Rest` service call. I want to have this client and this plugins endpoints be available in other parts of the `dotnet` ecosystem. Just to name a few `Mono.Net`, `.NET Maui` and `Unity`. This will help the Neo user base have no limitations on whatever framework or technology they desire to use.

As for maintainability, I'm not going anywhere; anytime soon and I have tons of time on my hands to write document, answer questions and update code. You have to start somewhere; so let's come together!

Yes I know it seemed fast; for me to make this plugin, But I really started at the begin of the year working on it and learning the neo-cli. 😸 So this is 5 months worth of learning and reading source code. 3 months of programming the code. My background is architecture design of systems.

_See https://github.com/neo-project/neo/issues/1004_

### Documentation
This [RestServer Documentation](https://github.com/cschuchardt88/neo-restful-plugin/blob/master/docs/RestServer.md) explains the working of the `RestServer` plugin. _Note still a work in progress._

**RestServer Addon Plugins**
This [Example Plugin Documentation](https://github.com/cschuchardt88/neo-restful-plugin/blob/master/docs/Addons.md) explains how to make plugins for the `RestServer`.

### Downloads
- [RestServer Plugin Binary Download](https://github.com/cschuchardt88/neo-modules/releases/tag/v3.6.0-beta)
- [RestClient Source Code](https://github.com/cschuchardt88/RestClient) _(Couldn't be add do to Build Props)_
- [RestClient Binary Download](https://github.com/cschuchardt88/RestClient/releases/tag/v1.0.0-beta) - .Net Standard 2.1

Swagger UI
- http://127.0.0.1:10339/swagger/index.html

### RestClient Code Example
```csharp
using RestClient;

using var httpClient = new HttpClient();
var restClient = new RestWebClient("http://127.0.0.1:10339/", httpClient);

try
{
    var block = await restClient.GetBlockAsync(0).ConfigureAwait(false);
    Console.WriteLine("Block Hash: {0}", block.Hash);
}
catch (ApiException<ErrorModel> ex)
{
    Console.WriteLine(ex.Result.Message);
}
Console.WriteLine("Please ENTER to exit.");
Console.ReadLine();
```

**Key Features**
- Override Error Handling _(for cleaner errors as json responses, will not crash clients or give unexpected error responses)_
- Compression
- CORS
- CORS Origins
- HTTPS
- Basic Authentication
- Disable Controllers _(from other plugins)_
- Proxy Forwards _(for data center usage)_
- Wallet Sessions
- Json Converter _(for having parameter types like UInt160, UInt256, StackItem, etc)_
- Implements Contracts, Ledger, Node, Token, Wallet and Utils full functionality
- Plugin Support _(easy process with no dependencies to this library)_
- Cleaner Plugin Code _(for new plugins)_
- Parameter Type Binding
- Swagger UI
- Plugin Support

**Change Log**
- [x] Ledger Controller API _(Interaction Ability)_
- [x] Contracts Controller API _(Interaction Ability)_
- [x] Neo/Gas Controller API _(Display Accounts)_
- [x] Node Controller API _(Block Height, Peers and Other Information)_
- [x] Token Controller API _(Nep17/Nep11 Interaction)_
- [x] Wallet  Controller API _(Open, Close, Transactions, etc)_
- [x] Utils Controller API _(Conversion Tools & Helpers)_
- [x] Expandability _(Rest Routes)_
- [x] Enable Https
- [x] CORS
- [x] Response Compression _(GZip)_
- [x] Basic Authentication
- [x] The ability to disable controller routes
- [x] Any HTTP Method
- [x] Any Route
- [x] Custom HTTP Headers
- [x] Rest Client _(No Dependencies Required)_ [Get It Here](https://github.com/cschuchardt88/RestClient)
- [x] Wallet Sessions
- [x] Multi-thread
- [x] Swagger
- [x] Update Swagger Endpoint Descriptions
- [x] Check for typos & misleading end user text _(in Swagger)_
- [x] Plugin Support

**Testing Progress**
- [x] JSON Converters _(Schema)_
- [x] Routes _(Inputs/Outputs/Parameters/Headers)_
- [x] All Configurations
- [x] Authentication
- [x] Disable Controllers
- [x] CORS
- [x] Proxy Forwards
- [x] Compression
- [x] Integration _(With Other Plugins)_
- [x] Loading Process _(Load Order)_
- [x] Wallet API _(File Access, Sessions & Etc)_
- [x] Load MainNet or TestNet T5 _(Check Data Integrity & Loading on Big Data)_
- [x] Dummy Controller _(Plugin)_
- [x] Swagger

### Neo-Cli Commands
image

### You can just add Swagger OpenAPI 2.0 to any project by
**Step1**
image

**Step 2**
image

**Step 3**
image


src/RestServer/Controllers/WalletController.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (tx.Size > 1024)
+            {
+                var calFee = tx.Size * NativeContract.Policy.GetFeePerByte(snapshot) + unchecked((long)NativeContract.GAS.Factor);
+                if (tx.NetworkFee < calFee) tx.NetworkFee = calFee;
+
+            }
+            if (tx.NetworkFee > _settings.MaxTransactionFee) return BadRequest("The necessary fee is more than the MaxTransactionFee, this transaction has failed. Please increase your MaxTransactionFee value.");
+
+            // Why do this again? What if its completed already? Copied from RPC Server...
+            transContext = new ContractParametersContext(snapshot, tx, _neosystem.Settings.Network);
```
cschuchardt88, line 246: Any reason why we need to do this again? In addition, what if previous transaction was completed? Why resign it? Note this is copied from RPC Server.

src/RestServer/Controllers/WalletController.cs
```diff
@@ -0,0 +1,258 @@
...
+                    ScriptHash = scriptHash,
+                }
+            });
+            if (tx == null) return BadRequest("Tranasction");
+            var transContext = new ContractParametersContext(snapshot, tx, _neosystem.Settings.Network);
+            wallet.Sign(transContext);
+            if (transContext.Completed == false)
+                return Content(transContext.ToJson().ToString(), MediaTypeNames.Application.Json);
+            tx.Witnesses = transContext.GetWitnesses();
+            // What if this is completed? Process and sign again?
```
cschuchardt88, line 236: What if the transaction is completed? Why resign later? Copied from RPC Server.

src/RestServer/Middleware/RestServerMiddleware.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            await _next(context);
+        }
+
+        public static void SetServerInfomationHeader(HttpResponse response)
+        {
+            var neoCliAsm = Assembly.GetEntryAssembly().GetName();
+            var restServerAsm = Assembly.GetExecutingAssembly().GetName();
+
+            response.Headers.Server = $"{neoCliAsm.Name}/{neoCliAsm.Version.ToString(3)} {restServerAsm.Name}/{restServerAsm.Version.ToString(3)}";
```
cschuchardt88, line 51: Haven't tested output. But output should be for example "Server: Neo/3.5.0 RestServer/3.5.0" for server header server output or should be use something to identify else for the server name and version?

src/RestServer/RestServerSettings.cs
```diff
@@ -0,0 +1,127 @@
...
+            AllowCorsUrls = Array.Empty<string>(),
+            DisableUrlPaths = Array.Empty<string>(),
+            EnableCompression = false,
+            CompressionLevel = CompressionLevel.SmallestSize,
+            EnableForwardedHeaders = false,
+            EnableDevelopmentMode = false,
+            MaxPageSize = 50u,
+            MaxConcurrentConnections = 40L,
+            MaxTransactionFee = 0_01000000,
+            JsonSerializerSettings = new JsonSerializerSettings()
```
cschuchardt88, line 73: Any default value you want me to change?

src/RestServer/RestWebServer.cs
```diff
@@ -0,0 +1,167 @@
...
+                {
+                    // Web server configuration
+                    options.AddServerHeader = false;
+                    options.Limits.MaxConcurrentConnections = _settings.MaxConcurrentConnections;
+                    options.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(_settings.KeepAliveTimeout);
+                    options.Limits.RequestHeadersTimeout = TimeSpan.FromSeconds(15);
+                    options.Listen(_settings.BindAddress, unchecked((int)_settings.Port), listenOptions =>
+                    {
+                        if (string.IsNullOrEmpty(_settings.SslCertFile)) return;
+                        listenOptions.UseHttps(_settings.SslCertFile, _settings.SslCertPassword, httpsOptions =>
```
cschuchardt88, line 60: don't understand why we need this for https. If cert is valid their shouldn't be a problem, I thought. Copied from RPC Server.

cschuchardt88:
Before i make client can someone make sure we are going to use this?

shargon:
Swagger can be added to the list

cschuchardt88:
The RestServer implements the same things for Kestrel as RpcServer does. Plus newer features. Everything is configurable and can be turned off. For added security you can enable basic authentication and/or https, with the ability to use a forward proxy. Also you have the ability to disable controllers and by default we disable WalletController. So if anyone want to make this plugin be public on a public node. We don't want people trying to open wallets or search around on the node for one. This uses middleware to block the controllers from ever being loaded. `"DisableControllers": [ "WalletController" ]`. Trying to access disabled controller routes with end in a 404 response from the server.

This is very generic but may appear not to be. Most the code is Model/Json converters for serializing the Models to JSON and JSON to Objects. These include UInt160, UInt256, StackItem, Transactions and Blocks to just name a few. If the application breaks I have Middleware set to not leak sensitive information to the end user. Server will return generic responses like _see below_. In the event of a crash of the application from any controller will result in a generic exception json response. To ensure applications that are consuming this RestServer will not have inconsistent responses for crashes or errors. Also any exception that is thrown will also result in error response like _see below_. So instead of returning your own response in a plugin for example with your own error. You would just throw an exception and the result will be 400 response with _see below_ as example
```json
{
    "code": 1001,
    "name": "ParameterFormatException",
    "message": "Parameter hash is in an invalid format."
}
```

if you have a look at theses files `RestServerPlugin.cs`, `RestServerSettings.cs` and `RestWebServer.cs` they can tell the whole story of what's going on and give you the idea of what is happening and the working of the application.

@shargon yes I can add it.

src/RestServer/Controllers/WalletController.cs
```diff
@@ -0,0 +1,271 @@
...
+        [HttpPost("open")]
+        public IActionResult OpenWallet(
+            [FromBody]
+            WalletOpenModel model)
+        {
+            if (string.IsNullOrEmpty(model.Path))
+                throw new JsonPropertyNullOrEmptyException(nameof(model.Path));
+            if (string.IsNullOrEmpty(model.Password))
+                throw new JsonPropertyNullOrEmptyException(nameof(model.Password));
+            if (System.IO.File.Exists(model.Path) == false)
```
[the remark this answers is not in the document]
cschuchardt88: How would you like me to fix this? I don't see if being a problem.

cschuchardt88:
> Swagger can be added to the list

Is this what you were looking for? See Attachment of results. I don't use swagger, but i have implemented it for you. I have implemented it in the pass before, but just don't like using it.
[Swagger Test.pdf](https://github.com/neo-project/neo-modules/files/12413371/Swagger.Test.pdf)

cschuchardt88:
@vncoelho currently the state of plugin is sound. The burden would add to new features like something in Kestrel server Or even maybe change swagger settings (which can be an easy task). Other than that the plugins get automatically loaded and no reference needs to be referenced back to this assembly. All plugins are independent from one another. Top post i posted has a link to my repo with an example of a "dummy example plugin" for adding to this Rest server.

I designed this plugin to have no restrictions for other plugins. To allow them do what they want. For example Route, Route parameters, parameters type, erroring, throwing exceptions and more. the code uses generic error middleware for responses. So other client that program against this api won't crash or get something unexpected results. Like an developer exception page. All default settings for RestServer have been removed and i set everything manually. If you like reading code have a look at `RestWebServer.cs`. Reusability was thought of in the design pattern of this plugin. No dependency injection is required as well.

**WorkFlow**
The workflow is a simple one, it follows the premise of MVC without the views. Just simple "API Controllers". You also have the ability to disable controllers and everything is configurable. No dependency reference to this assembly for plugins or soon to come RestClient. So users can easily add RestClient into other dotnet frameworks like dotnet Maui and Unity. RestClient will not require dependency on Neo.dll. Everything is server side.

Need more information or if i wasn't clear enough let me know.

src/RestServer/Controllers/WalletController.cs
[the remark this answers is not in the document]
shargon: You must restrict it to a specific folder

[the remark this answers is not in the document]
cschuchardt88: @shargon is relative path ok? Or do you want to one folder only? if so should that path be configurable?

cschuchardt88:
Typing up all the service descriptions for each controller.
image

src/RestServer/Controllers/WalletController.cs
```diff
@@ -0,0 +1,672 @@
...
+            if (model.From == null)
+                throw new JsonPropertyNullException(nameof(model.From));
+            var session = WalletSessions[sessionId];
+            session.ResetExpiration();
+            var wallet = session.Wallet;
+            var signers = model.Signers?.Select(s => new Signer() { Scopes = WitnessScope.CalledByEntry, Account = s }).ToArray();
+            var appEngine = ScriptHelper.InvokeScript(_settings, model.Script, signers);
+            if (appEngine.State != VM.VMState.HALT)
+                throw new ApplicationEngineException(appEngine.FaultException?.InnerException?.Message ?? appEngine.FaultException?.Message ?? string.Empty);
+            var tx = wallet.MakeTransaction(_neosystem.StoreView, model.Script, model.From, signers, maxGas: _settings.MaxInvokeGas);
```
cschuchardt88, line 652: Can someone make sure that I'm doing this right? What I'm trying to do is have a script that is already built, than make a transaction of the built script. For example what's not working is that I have a script that just Invokes the Transfer function of the Neo token contract. like so _see below_ and its not popping the event. Also no Neo was transferred.
```csharp
ScriptHash.MakeScript("transfer", from, to, 1, "test")
```

src/RestServer/Controllers/WalletController.cs
```diff
@@ -0,0 +1,672 @@
...
+        /// Opens a wallet.
+        /// </summary>
+        /// <param name="model"></param>
+        /// <returns>A newly created wallet session object.</returns>
+        /// <response code="200">Returns newly create wallet session object.</response>
+        /// <response code="400">If anything is invalid or request crashes.</response>
+        [HttpPost("open", Name = "WalletOpen")]
+        [ProducesResponseType(StatusCodes.Status200OK, Type = typeof(WalletSessionModel))]
+        [ProducesResponseType(StatusCodes.Status400BadRequest, Type = typeof(ErrorModel))]
+        public IActionResult WalletOpen(
```
cschuchardt88, line 65: If the wallet file is open and another request comes in to open the same wallet file. Should we create a new session (_this is currently what it does_) or return the already opened wallet's session? I don't know if the Neo wallet API will care. For example if one was to change the password of the wallet and the wallets are already open. Does that have any effect on the wallet for making or transferring funds.

src/RestServer/RestWebServer.cs
```diff
@@ -0,0 +1,298 @@
...
+                        {
+                            options.SwaggerDoc("v1", new OpenApiInfo()
+                            {
+                                Title = "RestServer API - V1",
+                                Description = "Neo-cli's Rest API.",
+                                Version = "v1",
+                                License = new OpenApiLicense()
+                                {
+                                    Name = "MIT",
+                                    Url = new Uri("http://www.opensource.org/licenses/mit-license.php"),
```
cschuchardt88, line 178: What license should we be using for this? I currently have MIT license. This is for Swaggers, _example below_
image

cschuchardt88:
RestClient is here at my [Repo](https://github.com/cschuchardt88/RestClient). Reason being is that I tried adding the client to this pull request but the build props won't let you remove the neo.dll package.

Also you could just add to any project by _see below_ But the reason is that we want a DotNet Standard 2.1 client for support in Unity, Xamarin, Mono and all other dotnet frameworks.

### Step1
image

### Step 2
image

### Step 3
image

cschuchardt88:
This is Ready for testing. Check top post for RestServer Plugin binaries _(Only has necessary files)_.

Just need input on the Swagger endpoint descriptions. For Example when you get the exception It will show error _"If anything is invalid or request crashes."_ but you are able to get the real error see example in Top Post of RestClient Example. It's just the way Open API client works in visual studio, so we need to update that and all other output messages of any kind.

cschuchardt88:
@shargon Swagger has limitations with polymorphism classes. There is no workarounds for this problem. Here is the problem that i am having. The document that is create by swagger and the client don't know how to make classes abstract. One example of this is StackItem's and Signer's class _see below_. What the client does is generate inheritance of the classes and doesn't mark them as abstract class or abstract properties. In the case of the example below it will think that the return is ONLY `AndConditionModel` and this is no way to fix this currently due to the specification of Open API. Making a custom client defeats the purpose of using Swagger. So i removed swagger and started making another `RestServer` version that doesn't use swagger.

**Example**
```csharp
public abstract class WitnessConditionModel
{
    public abstract WitnessConditionType Type { get; }
}

public class AndConditionModel : WitnessConditionModel
{
    public WitnessConditionModel[] Expressions { get; set; }
    public override WitnessConditionType Type => WitnessConditionType.And;
}

public class OrConditionModel : WitnessConditionModel
{
    public WitnessConditionModel[] Expressions { get; set; }
    public override WitnessConditionType Type => WitnessConditionType.Or;
}

public class GroupConditionModel : WitnessConditionModel
{
    public string Group { get; set; }
    public override WitnessConditionType Type => WitnessConditionType.Group;
}
```

shargon:
You can use `options.UseOneOfForPolymorphism();`

cschuchardt88:
@shargon there is currently a bug https://github.com/domaindrivendev/Swashbuckle.AspNetCore/issues/2702 with it and i did use a workaround 
```csharp
options.UseOneOfForPolymorphism();
options.SelectSubTypesUsing(baseType =>
{
    if (baseType == typeof(WitnessConditionModel))
    {
        return new[]
        {
                typeof(AndConditionModel),
                typeof(OrConditionModel),
                typeof(GroupConditionModel),
                // ...
            };
    }

    return Enumerable.Empty<Type>();
});
```

But the problem is the Open API client won't generate the right classes as abstract or properties as abstract because it doesn't know. Because the spec doesn't say. Now we can customize the classes to abstract by that defeats the purpose.

cschuchardt88:
@shargon Open API will generates classes like this see below and only have one return type of the class, in this case `OrConditionModel` is the only one returned by the function out of all the `Conditions` see below. But the classes are partial and can be edited. we can do that with our client, but other end users importing this swagger.json file won't know. We might get spam with issues about it.

```csharp
AndConditionModel andcon = await restClient.TestAsync().ConfigureAwait(false);
OrConditionModel orcon = (OrConditionModel)andcon;
```

```charp
[System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.20.0.0 (NJsonSchema v10.9.0.0 (Newtonsoft.Json v13.0.0.0))")]
public partial class WitnessConditionModel
{
    [Newtonsoft.Json.JsonProperty("type", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
    public WitnessConditionType Type { get; set; }

}
[System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.20.0.0 (NJsonSchema v10.9.0.0 (Newtonsoft.Json v13.0.0.0))")]
public partial class OrConditionModel : WitnessConditionModel
{
    [Newtonsoft.Json.JsonProperty("expressions", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
    public System.Collections.Generic.ICollection<AndConditionModel> Expressions { get; set; }

    [Newtonsoft.Json.JsonProperty("type", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
    public WitnessConditionType Type { get; set; }

}
```

shargon:
> @shargon Open API will generates classes like this see below and only have one return type of the class, in this case `OrConditionModel` is the only one returned by the function out of all the `Conditions` see below. But the classes are partial and can be edited. we can do that with our client, but other end users importing this swagger.json file won't know. We might get spam with issues about it.
> 
> ```csharp
> WitnessConditionModel andcon = await restClient.TestAsync().ConfigureAwait(false);
> OrConditionModel orcon = (OrConditionModel)andcon;
> ```
> 
> ```csharp
> [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.20.0.0 (NJsonSchema v10.9.0.0 (Newtonsoft.Json v13.0.0.0))")]
> public partial class WitnessConditionModel
> {
>     [Newtonsoft.Json.JsonProperty("type", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
>     public WitnessConditionType Type { get; set; }
> 
> }
> [System.CodeDom.Compiler.GeneratedCode("NJsonSchema", "13.20.0.0 (NJsonSchema v10.9.0.0 (Newtonsoft.Json v13.0.0.0))")]
> public partial class OrConditionModel : WitnessConditionModel
> {
>     [Newtonsoft.Json.JsonProperty("expressions", Required = Newtonsoft.Json.Required.Default, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
>     public System.Collections.Generic.ICollection<AndConditionModel> Expressions { get; set; }
> 
>     [Newtonsoft.Json.JsonProperty("type", Required = Newtonsoft.Json.Required.DisallowNull, NullValueHandling = Newtonsoft.Json.NullValueHandling.Ignore)]
>     public WitnessConditionType Type { get; set; }
> 
> }
> ```

It's strange that only one class appears, but I think that is openApi problem, it's well defined in the json

cschuchardt88:
I did make an issue about it on their github https://github.com/RicoSuter/NSwag/issues/4506#issue-1876376051

Liaojinghui:
What a beautiful coding style.

cschuchardt88:
This is at the point where you can try it out and see if it meets your requirements. Everything is up for discussion. If you can't use it either that's fine too.

Some questions i have:
1. Is it something you can use?
1. Is it something you can maintain?
1. Missing any features?
1. What will it take to get moving forward in a release?

But i do think this is a good start to get it moving out the door for merger.

Liaojinghui (dismissed):

Liaojinghui:
too long to review every single line, but functioning ok, maybe we can have it at first, then optimize it little by little.

Liaojinghui (approved):

Liaojinghui:
@cschuchardt88 do you mind to provide a readme for this plugin? i know others dont have it, but we will add readme for every plugin in the future.

Liaojinghui (approved):

Liaojinghui:
Merge since its plugins, people dont install it if they dont like it.

roman-khimov:
@Liaojinghui, you're rushing too fast with this, IMO. People will install it, people will try building things on it. I'm not sure it's tested enough and stable enough for them to do that. And notice that it could've been a separate plugin for some time exactly for these experiments. Now it's an official plugin from the Neo project, people have higher expectations of those.

cschuchardt88:
@roman-khimov 
I use it all the time. So It pretty stable. Its maybe lacking features. But all the basic and some advanced are all there.

shargon:
> @Liaojinghui, you're rushing too fast with this, IMO. People will install it, people will try building things on it. I'm not sure it's tested enough and stable enough for them to do that. And notice that it could've been a separate plugin for some time exactly for these experiments. Now it's an official plugin from the Neo project, people have higher expectations of those.

Agree and reverted.

Liaojinghui:
> @Liaojinghui, you're rushing too fast with this, IMO. People will install it, people will try building things on it. I'm not sure it's tested enough and stable enough for them to do that. And notice that it could've been a separate plugin for some time exactly for these experiments. Now it's an official plugin from the Neo project, people have higher expectations of those.

Agree, but only if someone review it. I dont see how this can introduce more bugs than there has already be.

shargon:
> @roman-khimov I use it all the time. So It pretty stable. Its maybe lacking features. But all the basic and some advanced are all there.

> Agree, but only if someone review it. I dont see how this can introduce more bugs than there has already be. I dont understand why you still stop merge and expect some one to review it while a pr has being there for 3 months and no one to review.

I understand, but understand that you have to review the code before merging, the more changes, the more complicated and time-consuming it requires, imagine (which I don't think is the case) that you add a line to leak the private key, **all the code must be reviewed by at least two core developers**, and tested by the great @superboyiii team.

Small pr will be reviewed faster than longer ones.

Liaojinghui:
Yes, the same reason why so many pr stay there for many years without anyone care about them.

shargon:
> Yes, the same reason why so many prs stay there for many years without anyone care about them.

The solution is not merge all of them.

shargon:
@cschuchardt88 please open a new PR with these changes.

Liaojinghui:
Not all of them, but something i have reviewed, and no one else review it after months.

roman-khimov:
That's why roadmaps and plans matter. Set some target, like "this PR should be dealt with by this time" and work towards that goal. Not planned? Likely not critical for now then (sure, life is complicated and sometimes it is). Add it to some milestone in the future, but keep real concentration on the nearest goals. We can't solve all problems at once and we can't risk destabilizing the system by accepting everything without proper reviews and tests. Going one problem after another with intermediate stable releases is more viable.

PR #812, opened 2023-08-28
RcpServer: Fixed CORS and Basic Auth

### LINKED ISSUE #811: RPC Server: Basic Authentication doesn't work with CORS
**Describe the bug**
You allow CORS on RPC Server, but do not allow "Access-Control-Allow-Credentials" if there is basic auth enabled. You do not follow the [specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials). In short the browser isn't sending the credentials for the login, cause of CORS is not configured right on RPC server.

https://github.com/neo-project/neo-modules/blob/99ffc846e4df93491d1d812cb4c6ef6ee875e056/src/RpcServer/RpcServer.cs#L161

Also note The CORS specification states that setting origins to "*" (all origins) is invalid if the Access-Control-Allow-Credentials header is present.

**To Reproduce**
1. enable config.json with RpcUser and RpcPass
1. open devtools in an browser
1. type `fetch('http://127.0.0.1:20332/?jsonrpc=2.0&id=1&method=getapplicationlog&params=[%220x632158b854f3e9cf132ba20b4aacbd5066bb9c330fe1a53a51d1a38c66fd175a%22]');` in console.

image

**Expected behavior**
CORS to work with basic auth enabled.

**Platform:**
 - RPC (HTTP)

cschuchardt88:
closes #811 

**Change Log**
- Fixed basic auth to work now, and popup auth window in browser
- Fixed CORS with basic auth
- Added AllowOrigins for basic auth to work

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -52,9 +52,13 @@ namespace Neo.Plugins
         {
             if (string.IsNullOrEmpty(settings.RpcUser)) return true;
 
-            context.Response.Headers["WWW-Authenticate"] = "Basic realm=\"Restricted\"";
-
             string reqauth = context.Request.Headers["Authorization"];
+            if (string.IsNullOrEmpty(reqauth))
+            {
+                context.Response.Headers["WWW-Authenticate"] = "Basic realm=\"Restricted\"";
+                context.Response.StatusCode = 401;
+            }
+
             string authstring;
             try
             {
@@ -129,11 +133,42 @@ namespace Neo.Plugins
             }))
             .Configure(app =>
             {
+                if (settings.EnableCors)
+                    app.UseCors("All");
+
                 app.UseResponseCompression();
                 app.Run(ProcessAsync);
             })
             .ConfigureServices(services =>
             {
+                if (settings.EnableCors)
+                {
+                    if (settings.AllowOrigins.Length == 0)
+                        services.AddCors(options =>
+                        {
+                            options.AddPolicy("All", policy =>
+                            {
+                                policy.AllowAnyOrigin()
+                                .AllowAnyHeader()
+                                .WithMethods("GET", "POST");
+                                // The CORS specification states that setting origins to "*" (all origins)
+                                // is invalid if the Access-Control-Allow-Credentials header is present.
+                                //.AllowCredentials() 
+                            });
+                        });
+                    else
+                        services.AddCors(options =>
+                        {
+                            options.AddPolicy("All", policy =>
+                            {
+                                policy.WithOrigins(settings.AllowOrigins)
+                                .AllowAnyHeader()
+                                .AllowCredentials()
+                                .WithMethods("GET", "POST");
+                            });
+                        });
+                }
+
                 services.AddResponseCompression(options =>
                 {
                     // options.EnableForHttps = false;
@@ -158,10 +193,6 @@ namespace Neo.Plugins
 
         public async Task ProcessAsync(HttpContext context)
         {
-            context.Response.Headers["Access-Control-Allow-Origin"] = "*";
-            context.Response.Headers["Access-Control-Allow-Methods"] = "GET, POST";
-            context.Response.Headers["Access-Control-Allow-Headers"] = "Content-Type";
-            context.Response.Headers["Access-Control-Max-Age"] = "31536000";
             if (context.Request.Method != "GET" && context.Request.Method != "POST") return;
             JToken request = null;
             if (context.Request.Method == "GET")
--- a/src/RpcServer/Settings.cs
+++ b/src/RpcServer/Settings.cs
@@ -38,6 +38,8 @@ namespace Neo.Plugins
         public int MaxConcurrentConnections { get; init; }
         public string RpcUser { get; init; }
         public string RpcPass { get; init; }
+        public bool EnableCors { get; init; }
+        public string[] AllowOrigins { get; init; }
         public long MaxGasInvoke { get; init; }
         public long MaxFee { get; init; }
         public int MaxIteratorResultItems { get; init; }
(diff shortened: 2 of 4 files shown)


src/RpcServer/RpcServer.cs
```diff
@@ -52,9 +52,13 @@ private bool CheckAuth(HttpContext context)
         {
             if (string.IsNullOrEmpty(settings.RpcUser)) return true;
 
-            context.Response.Headers["WWW-Authenticate"] = "Basic realm=\"Restricted\"";
-
             string reqauth = context.Request.Headers["Authorization"];
+            if (string.IsNullOrEmpty(reqauth))
+            {
+                context.Response.Headers["WWW-Authenticate"] = "Basic realm=\"Restricted\"";
+                context.Response.StatusCode = 401;
```
shargon, line 59:
```suggestion
                context.Response.StatusCode = 401;
                return false;
```

cschuchardt88: if you do that then window won't popup in the browser. Its not valid spec for using WWW-Authenticate header you need to return 401 or remove www-authenticate header from the response.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate#digest_authentication_with_sha-256_and_md5
>The first time the client requests the document, no [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) header field is sent. Here the server responds with an HTTP 401 message that includes a challenge for each digest algorithm it supports, in its order of preference (SHA256 and then MD5)

Also see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
>The server responds with a [401](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) Unauthorized message that includes at least one [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) header.

shargon: It will return false later

cschuchardt88: yes, But the server with response with Status Code 200 (OK). Which is invalid. If you are using those headers and creds are wrong or not supplied.

src/RpcServer/RpcServer.cs
```diff
@@ -129,11 +133,41 @@ public void StartRpcServer()
             }))
             .Configure(app =>
             {
+                if (settings.EnableCors)
+                    app.UseCors("All");
```
vncoelho, line 137: the design looks good!

vncoelho: Coincidentally, we got this problem while using gitpod to access RPC nodes.
I will test this PR asap.

src/RpcServer/RpcServer.cs
```diff
@@ -129,11 +133,41 @@ public void StartRpcServer()
...
+                                // The CORS specification states that setting origins to "*" (all origins)
+                                // is invalid if the Access-Control-Allow-Credentials header is present.
+                            });
+                        });
+                    else
+                        services.AddCors(options =>
+                        {
+                            options.AddPolicy("All", policy =>
+                            {
+                                policy.WithOrigins(settings.AllowOrigins)
```
vncoelho, line 163: I tried to specify a random origin policy `"AllowOrigins": ["http://yourapp.com"]`,automatically, I think I should had get an error while calling it from "localhost" but it kept responding.

cschuchardt88: It doesn't care the origin only when you ask for it. Its a browser thing.

vncoelho: In fact I forgot to update the client. It is blocking right now.

vncoelho:
We are facing some CORS issue using Gitpod.

I tried to re-route C# RPC Plugin with express node, using Axios to make the `post` request.
Then it worked. The Gitpod behaved normally withour CORS problems.

I am not understanding why the browser is having issues with the request directly to the RPC plugin.

The strange situation is that this just happens when there are multiple requests. Let's say 5 requests per second.

cschuchardt88:
Some stuff to note here is the `KeepAliveTimeout` is hard coded to **1** minute and the max allow concurrent connections is configurable in the config.json aka open connections. More than likely the browser is sending `Keep-Alive` header. But note CORS specification states that setting origins to "*" (all origins) is invalid if the Access-Control-Allow-Credentials header is present aka basic authentication headers & cookies.

But if you could elaborate a little bit more. Like what is `C# RPC Plugin`? RpcClient? Also did this PR fix the issue at hand? Your having another issue with multi-requests per second? with or without the code changes?

vncoelho:
@cschuchardt88, considering your comment it was found that `MaxConcurrentConnections` was the problem for our case.
The good part is that I tested this PR in different scenarios.

For it is ready.

vncoelho:
Maybe we can also move `KeepAliveTimeout` and `RequestHeadersTimeout`  to config.
```cs
                // Default value is 1 minutes
                options.Limits.KeepAliveTimeout = TimeSpan.FromMinutes(1);
                // Default value is 15 seconds
                options.Limits.RequestHeadersTimeout = TimeSpan.FromSeconds(15);
```

cschuchardt88:
do you want that in seconds or minutes for `KeepAliveTimeout`, also should `CORS` be enabled by default?

cschuchardt88:
Also something we may want to consider in the future is "Enable forward proxies"; will help with web hosting and cloud stuff.

vncoelho:
I think the PR should keep the same behavior as nowadays, @cschuchardt88.

Parameters could be in seconds.

Regarding or issue about the `neo-cli` handling all this configurations, that is also a possibility for the future that we should consider.

vncoelho:
> Also something we may want to consider in the future is "Enable forward proxies"; will help with web hosting and cloud stuff.

sure!

vncoelho (dismissed):
@superboyiii ,can you test?

src/RpcServer/RpcServer.cs
shargon, line 59:
```suggestion
                context.Response.StatusCode = 401;
                return false;
```

src/RpcServer/Settings.cs
```diff
@@ -56,6 +60,9 @@ public record RpcServerSettings
             MaxGasInvoke = (long)new BigDecimal(10M, NativeContract.GAS.Decimals).Value,
             MaxFee = (long)new BigDecimal(0.1M, NativeContract.GAS.Decimals).Value,
             TrustedAuthorities = Array.Empty<string>(),
+            EnableCors = true,
+            KeepAliveTimeout = 1,
```
shargon, line 64: Maybe we should be able to disable it with -1

cschuchardt88: Sorry `KeepAliveTimeout` is in seconds never changed default config to reflex.

Liaojinghui (approved):

vncoelho:
ping @superboyiii

src/RpcServer/Settings.cs
shargon: I mean, if possible to disable the keep alive when -1?

cschuchardt88: yes i added https://github.com/neo-project/neo-modules/pull/812/commits/80a0dc87f2459b70a5c09a2903a54a33fe61149e

cschuchardt88:
Any update on this @superboyiii ? Just fyi, this fixes a bug in the wild.

shargon (approved):

Jim8y (approved):

PR #813, opened 2023-08-28
RpcServer:  added GetContractState by contract id support

### LINKED ISSUE #788: Support getcontractstate by ID for non-native contracts
#378 has added support for `getcontractstate` requests using native contract IDs/names (see #355 and #339 as well). Back when it had been implemented there was no way to map ID to contract hash for non-native contracts, now after neo-project/neo#2807 solving neo-project/neo#2803 it is easily possible to do that.

Add logic to allow `getcontractstate` to operate with IDs for non-native contracts. NeoGo supports this since 0.92.0 and we can finally unify behavior of this call.

- RPC (HTTP)

cschuchardt88:
closes #788 

Also per https://github.com/neo-project/neo/issues/2803#issue-1333773764

**Change Log**
- GetContractState now supports contract id. (_On RcpServer_)
- GetContractState now supports contract id. (_On RpcClient_)

### DIFF
--- a/src/RpcClient/RpcClient.cs
+++ b/src/RpcClient/RpcClient.cs
@@ -245,6 +245,15 @@ namespace Neo.Network.RPC
             return ContractStateFromJson((JObject)result);
         }
 
+        /// <summary>
+        /// Queries contract information, according to the contract id.
+        /// </summary>
+        public async Task<ContractState> GetContractStateAsync(int id)
+        {
+            var result = await RpcSendAsync(GetRpcName(), id).ConfigureAwait(false);
+            return ContractStateFromJson((JObject)result);
+        }
+
         public static ContractState ContractStateFromJson(JObject json)
         {
             return new ContractState
--- a/src/RpcServer/RpcServer.Blockchain.cs
+++ b/src/RpcServer/RpcServer.Blockchain.cs
@@ -120,9 +120,17 @@ namespace Neo.Plugins
         [RpcMethod]
         protected virtual JToken GetContractState(JArray _params)
         {
-            UInt160 script_hash = ToScriptHash(_params[0].AsString());
-            ContractState contract = NativeContract.ContractManagement.GetContract(system.StoreView, script_hash);
-            return contract?.ToJson() ?? throw new RpcException(-100, "Unknown contract");
+            if (int.TryParse(_params[0].AsString(), out int contractId))
+            {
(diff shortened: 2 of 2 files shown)


roman-khimov (approved):

Liaojinghui (approved):

src/RpcClient/RpcClient.cs
```diff
@@ -245,6 +245,15 @@ public async Task<ContractState> GetContractStateAsync(string hash)
             return ContractStateFromJson((JObject)result);
         }
 
+        /// <summary>
+        /// Queries contract information, according to the contract id.
+        /// </summary>
+        public async Task<ContractState> GetContractStateAsync(int id)
+        {
+            var result = await RpcSendAsync(GetRpcName(), id).ConfigureAwait(false);
+            return ContractStateFromJson((JObject)result);
```
cschuchardt88, line 254: Note this doesn't use the "RpcContractState" class which is in the model folder. The function is using the function below it ("ContractStateFromJson", line 257). Should that be changed and updated?

shargon (approved):

shargon:

src/RpcClient/RpcClient.cs
superboyiii: Can't use the same method name as `GetContractStateAsync(string hash)`, otherwise it will get `GetContractStateAsync(string hash)` as default when `GetRpcName()`, then id will be and invaild param.

cschuchardt88: It's just a overload for RcpClient and works fine. That's what we want to do. The method that was changed on RpcServer; `GetContractState` to allow contract id to be processed.

**Example**
```csharp
var protocolSettings = ProtocolSettings.Load("config.json", true);
RpcClient rpcClient = new(new("http://127.0.0.1:20332"), "admin", "admin", protocolSettings);
var contractstate = await rpcClient.GetContractStateAsync(742).ConfigureAwait(false);
Console.WriteLine("ScriptHash is {0}.", contractstate.Hash);
Console.WriteLine("Press ENTER to exit . .  .");
Console.ReadLine();
```

superboyiii: Sorry, just a fault on my code, it works as the expected!

superboyiii:
Test **Pass**


image

superboyiii (approved):

PR #823, opened 2023-09-13
add hardfork settings to getversion call

EdgeDLT:
Close #822

Issue #822, opened 2023-09-13
Add hardfork settings to getversion RPC call


EdgeDLT:
**Summary or problem description**
`getversion` RPC call returns important protocol configuration settings for a node, but hardfork information is not included.

**Do you have any solution you want to propose?**
Expose the hardfork settings in `getversion`.

**Where in the software does this update applies to?**
- Plugins

PR #825, opened 2023-09-15
Added hardfork to RpcClient

cschuchardt88:
Added hardfork to RpcClient and tested

This is for #823 implementation of hardfork settings in `getversion` rpc method.

### DIFF
--- a/tests/Neo.Network.RPC.Tests/RpcTestCases.json
+++ b/tests/Neo.Network.RPC.Tests/RpcTestCases.json
@@ -2491,7 +2491,13 @@
           "addressversion": 0,
           "maxtransactionsperblock": 0,
           "memorypoolmaxtransactions": 0,
-          "initialgasdistribution": 0
+          "initialgasdistribution": 0,
+          "hardforks": [
+            {
+              "name": "HF_Aspidochelone",
+              "blockheight": 0
+            }
+          ]
         }
       }
     }
--- a/src/RpcClient/Models/RpcVersion.cs
+++ b/src/RpcClient/Models/RpcVersion.cs
@@ -9,6 +9,9 @@
 // modifications are permitted.
 
 using Neo.Json;
+using System;
+using System.Collections.Generic;
+using System.Linq;
 
 namespace Neo.Network.RPC.Models
 {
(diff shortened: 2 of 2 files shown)


tests/Neo.Network.RPC.Tests/RpcTestCases.json
```diff
@@ -2491,7 +2491,13 @@
           "addressversion": 0,
           "maxtransactionsperblock": 0,
           "memorypoolmaxtransactions": 0,
-          "initialgasdistribution": 0
+          "initialgasdistribution": 0,
+          "hardforks": [
+            {
+              "name": "HF_Aspidochelone",
```
roman-khimov, line 2497: BTW, speaking of #822, can we avoid `HF_` prefixes here? It's `hardforks` section already.

cschuchardt88: but it based off Enum `Hardfork`, So its parses the name, i guess i could put value, i believe that works.

cschuchardt88: oh i see what your saying, Thats how its used in #823.

```
public enum Hardfork : byte
{
    HF_Aspidochelone,
    HF_Basilisk
}
```

roman-khimov: Yeah, it's not a client problem and not something client should be bothered with. It's just that I've only noticed it here. But server-side it'd probably be a bit better.

cschuchardt88: https://github.com/neo-project/neo-modules/blob/63e3ec3dc4092ea83f010b1fe9704a70129fbcda/src/RpcServer/RpcServer.Node.cs#L88

Code uses key of the enum which in our case is `HF_Aspidochelone` and `HF_Basilisk`. We would have to change the server to use the value instead. If you would like me to proceed in changing the server let me know.

shargon: > Yeah, it's not a client problem and not something client should be bothered with. It's just that I've only noticed it here. But server-side it'd probably be a bit better.

Why we need it in server? We can remove it from both

cschuchardt88: remember this #823 someone merged it

Liaojinghui:
Good to go? ifs its already in the server.

shargon (approved):

PR #827, opened 2023-09-18
MaxParamsDepth and MaxRequestBodySize in RpcServer

### LINKED ISSUE #820: Rpc Server: WitnessRules potential  DDOS
There is a potential problem with `rpc server` with signer rules. The neo function see below doesn't pass the max depth for json object. This will allow someone to DDOS all `Rpc nodes`. I will not post the way to do this here.

https://github.com/neo-project/neo/blob/4e9314d6b33fb8158e4f2527f889a9cbe7377a73/src/Neo/Network/P2P/Payloads/WitnessRule.cs#L63

https://github.com/neo-project/neo-modules/blob/09c2879958a916e0867fc78c64a04edfabe6935f/src/RpcServer/RpcServer.SmartContract.cs#L166-L182

Put a max depth

- RPC (HTTP)

shargon:
Close neo-project/neo#2950

### DIFF
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -109,6 +109,7 @@ namespace Neo.Plugins
             {
                 // Default value is 5Mb
                 options.Limits.MaxRequestBodySize = settings.MaxRequestBodySize;
+                options.Limits.MaxRequestLineSize = Math.Min(settings.MaxRequestBodySize, options.Limits.MaxRequestLineSize);
                 // Default value is 40
                 options.Limits.MaxConcurrentConnections = settings.MaxConcurrentConnections;
                 // Default value is 1 minutes
@@ -178,7 +179,6 @@ namespace Neo.Plugins
                 {
                     try
                     {
-                        if (_params.Length > settings.MaxRequestBodySize) return;
                         _params = Encoding.UTF8.GetString(Convert.FromBase64String(_params));
                     }
                     catch (FormatException) { }
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -178,6 +178,7 @@ namespace Neo.Plugins
                 {
                     try
                     {
+                        if (_params.Length > settings.MaxRequestBodySize) return;
                         _params = Encoding.UTF8.GetString(Convert.FromBase64String(_params));
                     }
                     catch (FormatException) { }
--- a/src/RpcServer/RpcServer.cs
+++ b/src/RpcServer/RpcServer.cs
@@ -32,6 +32,8 @@ namespace Neo.Plugins
 {
     public partial class RpcServer : IDisposable
     {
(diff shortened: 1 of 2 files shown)


src/RpcServer/RpcServer.cs
```diff
@@ -105,6 +107,9 @@ public void StartRpcServer()
         {
             host = new WebHostBuilder().UseKestrel(options => options.Listen(settings.BindAddress, settings.Port, listenOptions =>
             {
+                // Default value is 5Mb
+                options.Limits.MaxRequestBodySize = settings.MaxRequestBodySize;
```
cschuchardt88, line 111: You should remove this.

vncoelho: This is the same as default as it is, right?

vncoelho: The default it 28,6mb, @cschuchardt88.
Maybe this should be a Math.Min as well, no?

vncoelho: The default max looks like to be `30 x 1024 x 1024`

cschuchardt88: He already has that under the line i commented on. I was just saying he doesn need the extra line
```
options.Limits.MaxRequestLineSize = Math.Min(settings.MaxRequestBodySize, options.Limits.MaxRequestLineSize);
```

vncoelho: But this are different properties.

Defaults of this is  8,192 bytes (8 KB)

cschuchardt88: oh ur right, looked at it at quick glance

cschuchardt88 (approved):

src/RpcServer/RpcServer.cs
shargon: > You should remove this.

Why?

shargon: One is the request line (GET) and the other is the body, I set both just in case if the configuration is less than 8k

shargon:

src/RpcServer/RpcServer.cs
cschuchardt88: > > You should remove this.
> 
> Why?

looks good, but can you do `Math.Min` function

Liaojinghui (approved):

shargon:
Ping @superboyiii

superboyiii (approved):
Good for me.

cschuchardt88:
@shargon @Liaojinghui @superboyiii this doesnt fix the problem..... it has to do with a string in a parameter

cschuchardt88:
You need maxdepth for each parameter.

shargon:
But this 

> here is an example of what im talking about [example.txt](https://github.com/neo-project/neo-modules/files/13383150/example.txt)

But this send a lot of accounts not Rules, is a different one.

PR #828, opened 2023-09-19
DBFTPlugin: adapt new Conflicts attribute storage scheme

AnnaShaleva:
Upgrade neo-modules to use https://github.com/neo-project/neo/pull/2913.

This PR requires neo core library dependency upgrade.


shargon (changes requested):
Seems good, but wait for merge https://github.com/neo-project/neo/pull/2913

superboyiii:
I've tested this together with https://github.com/neo-project/neo/pull/2913, should be good, but need format fix.

shargon (approved):

PR #832, opened 2023-09-25
fix: selfish mining

Liaojinghui:
https://github.com/neo-project/neo/issues/2918

Prevents:

- Primary starts consensus early.
- Primary starts consensus early but with a `future` timestamp.


vncoelho (changes requested):
@Liaojinghui ,what exactly are you trying to solve and when it happens?

I see the code and saw some blockIndex and 10000 as a value of the if. Why using block index?

Liaojinghui:
The problem here is that malicious nodes can start the consensus early when they become the primary, speeding up the consensus and thus generating more blocks and mining more GAS. We did not check a thing that is similar to `mining difficulty`. As v17 has mentioned here https://github.com/neo-project/neo/issues/2918, this is not  an emergency fix, it is totally OK, it you think consensus nodes should be allowed to speed up the consensus maliciously.

shargon (commented):
Create a new method IsValidTimestamp(previous, current) with UT?

Liaojinghui:
> Create a new method IsValidTimestamp(previous, current) with UT?
would be challenging to have 10000 blocks UT, LOL, lets find out.

cschuchardt88:
One way, is to use `NeoSystem` in memory, than you could sleep  thread until height is 10k. Should download on network and should be quick.

Liaojinghui:
Funny when i set it to 10000, not anymore.

PR #833, opened 2023-09-26
RpcClient: Check for NEP17 Transfer

### LINKED ISSUE #775: Transactions for invalid transfers
I have the following issue. When we run a transaction for a certain Nep17 token, which sends tokens from one wallet to another, but the amount is more than the current balance, the transaction does not return an error. A transaction with the returned hash is successfully stored inside a block, but it actually do not transfer any tokens (because the Transfer method of the token contract returned false).
When I get the state of the transaction through RPC call, it returns same data as a successful one.

There are 2 things which I found during this research and think they are problems which should be resolved:

CreateTransferTxAsync in NEP17Api.cs creates a script for "transfer" method, which does not add ASSERT at the end (in order to check the boolean result from Transfer method of the token contract). The implementation of MakeTransaction in Wallet.cs in Neo project does it this way, so I'd trust its implementation more and I think it is just a bug in the code of the RpcClient module.

The next strange step is TransactionManagerFactory -> MakeTransactionAsync. It does a test invoke of the transaction script, but totally ignores the result from it. In my case, if I add the ASSERT from 1), the invocation result here is FAULT. I think that this method should check that before creating a transaction and returning it. Probably throw an exception or return a result of (RpcInvokeResult, Transaction).
This way the code that calls MakeTransactionAsync can decide whether to send the raw transaction or return an error or something like that. In my case if I add the ASSERT and still execute the SendRawTransaction, it also ignores the FAULT code and stores the "invalid" transfer transaction in a block successfully.

What do you think about these issues? Are they things that has to be fixed or there is a reason to create the script without ASSERT and not checking the test-invoke state before sending the raw transaction ?

cschuchardt88:
closes #775 

**Change Log**
- Add exception for invalid Nep-17 transfers

### DIFF
--- a/src/RpcClient/Nep17API.cs
+++ b/src/RpcClient/Nep17API.cs
@@ -34,6 +34,11 @@ namespace Neo.Network.RPC
         public Nep17API(RpcClient rpcClient) : base(rpcClient) { }
 
         /// <summary>
+        /// Throw an error if transfer returns false
+        /// </summary>
+        public bool ErrorOnTransferFalse { get; set; } = false;
+
+        /// <summary>
         /// Get balance of NEP17 token
         /// </summary>
         /// <param name="scriptHash">contract script hash</param>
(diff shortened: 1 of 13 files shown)


src/RpcClient/Nep17API.cs
```diff
@@ -141,9 +146,15 @@ public async Task<Transaction> CreateTransferTxAsync(UInt160 scriptHash, KeyPair
 
             TransactionManagerFactory factory = new TransactionManagerFactory(rpcClient);
             TransactionManager manager = await factory.MakeTransactionAsync(script, signers).ConfigureAwait(false);
-            return await manager
-                .AddSignature(fromKey)
-                .SignAsync().ConfigureAwait(false);
+
+            var vmResult = await rpcClient.InvokeScriptAsync(script, signers).ConfigureAwait(false);
+
+            if (ErrorOnTransferFalse == false || (vmResult.State == VMState.HALT && vmResult.Stack?[0].GetBoolean() == true))
```
roman-khimov, line 152: I think usually this is solved by adding an `ASSERT` to the end of the script (which is trivial here since the script is created in the same function`).

shargon: It's better to do it like @roman-khimov said

shargon (commented):
@cschuchardt88 Could you give me rights for modify?

cschuchardt88:
@shargon thx for update. I was originally going to do something like that with an throw exception. But wasn't sure on what you were looking for. Cause didnt want to break the interface....

shargon (approved):

PR #834, opened 2023-10-06
Added Crontab job plugin

cschuchardt88:
<p align="center" width="100%">
    Crontab-logo
</p>

```bash
*    *    *    *    *       Crontab Expression Chart
│    │    │    │    │
│    │    │    │    │
│    │    │    │    |_________ Day of Week (0 – 6) (0 is Sunday)
│    │    │    |____________ Month (1 – 12), * means every month
│    │    |______________ Day of Month (1 – 31), * means every day
│    |________________ Hour (0 – 23), * means every hour
|___________________ Minute (0 – 59), * means every minute
```
Workaround for https://github.com/neo-project/neo/issues/2903

# neo-cron-plugin
Task scheduler for sending transactions to the blockchain. Just as the
name implies `Crontab` does just that! Schedule jobs to invoke contracts
or transfer funds at certain times of the day, month, year, hour and
minute.

## Features
- Task Scheduler
- Manage jobs in `cli` console.
- Send transaction types.
  - Invoke Contract Methods
  - Send Nep-17 Transfers

image

# Example Tasks
You can find more details on how to create and configure jobs [here](https://github.com/cschuchardt88/neo-cron-plugin/blob/master/docs/JOBS.md).

**Schedule Job Examples**
- [contract](https://github.com/cschuchardt88/neo-cron-plugin/blob/master/examples/HelloInvokeMethod.job)
- [transfer](https://github.com/cschuchardt88/neo-cron-plugin/blob/master/examples/HelloTransfer.job)


cschuchardt88:
If this is something we are going to use for temporary workaround, until we implement https://github.com/neo-project/neo/issues/2903. I have problem. Github Actions not picking up new tests i added. how to add one?

I see

https://github.com/neo-project/neo-modules/blob/6ead78007ba6003eb050f7c7bffeb4e2476f348e/.github/workflows/main.yml#L24-L32

But i don't know which paths i need or if I'm in the right spot. Never used `coverlet`.

Liaojinghui (approved):

shargon:
With RPC you can already send task with crontab, why we need to replicate it?

cschuchardt88:
> With RPC you can already send task with crontab, why we need to replicate it?

You mean talking to website? I'm sure windows doesn't have anything built-in?

This has better security than using a side channels. You can think of it as `github action`s. This is just a base of what is to come of this service. Idea is to expand the cli not to limit it for non-technical users.

Liaojinghui:
> With RPC you can already send task with crontab, why we need to replicate it?

Dude, isn't this plugin? why can not we have many plugins here, users pick which one they want, right?

cschuchardt88:
It would take a year before it gets added, if interested, you can always go to my personal repos. It a collection of what `neo` should be doing or needs. https://github.com/cschuchardt88/neo-cron-plugin

PR #838, opened 2023-10-12
RpcServer, RpcClient: strip HF_ prefix from `getversion` response

AnnaShaleva:
It's nice to have clear hardfork ame without hardfork prefix so that the resulting hardfork JSON-ized name can be applicable by both C# and NeoGo nodes.

Ref. https://github.com/neo-project/neo-modules/issues/822.

### DIFF
--- a/src/RpcClient/Models/RpcVersion.cs
+++ b/src/RpcClient/Models/RpcVersion.cs
@@ -44,7 +44,8 @@ namespace Neo.Network.RPC.Models
                 json["initialgasdistribution"] = InitialGasDistribution;
                 json["hardforks"] = new JArray(Hardforks.Select(s => new JObject()
                 {
-                    ["name"] = s.Key,
+                    // Strip HF_ prefix.
+                    ["name"] = s.Key.ToString().Substring(3),
                     ["blockheight"] = s.Value,
                 }));
                 return json;
@@ -63,7 +64,7 @@ namespace Neo.Network.RPC.Models
                     MaxTransactionsPerBlock = (uint)json["maxtransactionsperblock"].AsNumber(),
                     MemoryPoolMaxTransactions = (int)json["memorypoolmaxtransactions"].AsNumber(),
                     InitialGasDistribution = (ulong)json["initialgasdistribution"].AsNumber(),
-                    Hardforks = new Dictionary<Hardfork, uint>(((JArray)json["hardforks"]).Select(s => new KeyValuePair<Hardfork, uint>(Enum.Parse<Hardfork>(s["name"].AsString()), (uint)s["blockheight"].AsNumber()))),
+                    Hardforks = new Dictionary<Hardfork, uint>(((JArray)json["hardforks"]).Select(s => new KeyValuePair<Hardfork, uint>(Enum.Parse<Hardfork>($"HF_{s["name"].AsString()}"), (uint)s["blockheight"].AsNumber()))),
                 };
             }
         }
--- a/src/RpcClient/Models/RpcVersion.cs
+++ b/src/RpcClient/Models/RpcVersion.cs
@@ -45,7 +45,7 @@ namespace Neo.Network.RPC.Models
                 json["hardforks"] = new JArray(Hardforks.Select(s => new JObject()
                 {
                     // Strip HF_ prefix.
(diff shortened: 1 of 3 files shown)


Liaojinghui:
weird to make C# aligned to neogo, instead of neogo aligned to C#. If it is not a bug or improvement proposal, i dont have any motivation to stay `aligned` to neogo.

roman-khimov:
It's obviously an improvement for the RPC protocol. These prefixes are OK for some code-level constants in a big module, but they're not really interesting for RPC reply that has `hardforks` section.

cschuchardt88:
you have to remove 

https://github.com/neo-project/neo-modules/blob/6ead78007ba6003eb050f7c7bffeb4e2476f348e/tests/Neo.Network.RPC.Tests/RpcTestCases.json#L2495-L2500

for the tests to work again.

Liaojinghui:
Changes to the interface should be very careful, we dont know if someone is already using it or not, and it is not a good experience to change the inteface from time to time. Yes, without HF prefix is much reasonable, but since it has `HF` and has being in this way for a well, then it become a historical problem. 

I wont stand my opinion if @shargon  agrees to update, but before that, i dont think it is necessary or should be done in another compatible way which does not break existing interface.

src/RpcClient/Models/RpcVersion.cs
```diff
@@ -44,7 +44,8 @@ public JObject ToJson()
                 json["initialgasdistribution"] = InitialGasDistribution;
                 json["hardforks"] = new JArray(Hardforks.Select(s => new JObject()
                 {
-                    ["name"] = s.Key,
+                    // Strip HF_ prefix.
+                    ["name"] = s.Key.ToString().Substring(3),
```
cschuchardt88, line 48: why?

src/RpcClient/Models/RpcVersion.cs
```diff
@@ -63,7 +64,7 @@ public static RpcProtocol FromJson(JObject json)
                     MaxTransactionsPerBlock = (uint)json["maxtransactionsperblock"].AsNumber(),
                     MemoryPoolMaxTransactions = (int)json["memorypoolmaxtransactions"].AsNumber(),
                     InitialGasDistribution = (ulong)json["initialgasdistribution"].AsNumber(),
-                    Hardforks = new Dictionary<Hardfork, uint>(((JArray)json["hardforks"]).Select(s => new KeyValuePair<Hardfork, uint>(Enum.Parse<Hardfork>(s["name"].AsString()), (uint)s["blockheight"].AsNumber()))),
+                    Hardforks = new Dictionary<Hardfork, uint>(((JArray)json["hardforks"]).Select(s => new KeyValuePair<Hardfork, uint>(Enum.Parse<Hardfork>($"HF_{s["name"].AsString()}"), (uint)s["blockheight"].AsNumber()))),
```
cschuchardt88, line 67: you are adding it again. why?

cschuchardt88: Test fails with `System.ArgumentException: Requested value 'HF_HF_Aspidochelone' was not found.`

shargon (commented):
I prefer to use a method to check if the string startWith, instead of replace always the first 3 chars.

AnnaShaleva:
As Roman said in https://github.com/neo-project/neo-modules/pull/838#issuecomment-1760143151, the initial idea for hardforks naming was to use "Aspidochelone", "Basilisk" and etc. for user-facing outputs (see the https://github.com/neo-project/neo/pull/2749#discussion_r878971641), and `HF_Aspidochelone` is just a code-level enum implementation that is expected to stay at the code level. So it would be nice to keep human-readable names in the JSON response output.

src/RpcClient/Models/RpcVersion.cs
AnnaShaleva: See the https://github.com/neo-project/neo-modules/pull/838#issuecomment-1761277835 and updated version.

AnnaShaleva:
> I prefer to use a method to check if the string startWith, instead of replace always the first 3 chars.

Fixed, see the updated version, please.

Liaojinghui (approved):

cschuchardt88 (approved):

PR #842, opened 2023-11-12
Add debug mode to the applicationlog to get log info

### LINKED ISSUE #841: ApplicationLog Output LogEventArgs
`ApplicationLog` needs to output the application engine logs `LogEventArgs`. Reason is for `neo-express` to give more debug information. Of course we don't have to have it. But it's a quality of life feature.

```json
{
  "transaction": {
    "hash": "0xb9203e4f28c5e07eef0bc5e225a2603cc7f4ba0bba8ba04f048750ac7f8ba590",
    "size": 205,
    "version": 0,
    "nonce": 640966548,
    "sender": "NbGJ2PMcd7DeVYCnAf3neEuEqgLdaGX6rP",
    "sysfee": "2645280",
    "netfee": "1188520",
    "validuntilblock": 5876,
    "signers": [
      {
        "account": "0xf114fa11a19cb02299010fe5a5a0518179ee61a8",
        "scopes": "CalledByEntry"
      }
    ],
    "attributes": [],
    "script": "DAVDaHJpcxHAHwwIc2F5SGVsbG8MFAIXhEkqUKHxhCwG7sF2B8byqmlBQWJ9W1I=",
    "witnesses": [
      {
        "invocation": "DEDejoY2ixnHQmejF6Yc6NsnIrHOHRMGk4l5mDvO/FYKhSMcYR/KeU51nTncemMJEPAm98UI9cVptQldeVvqikew",
        "verification": "DCECJJjt3o9js9i1pEOyfB7alrHtZ+v7Pu1Xx736tD2VS3dBVuezJw=="
      }
    ]
  },
  "application-log": {
    "txid": "0xb9203e4f28c5e07eef0bc5e225a2603cc7f4ba0bba8ba04f048750ac7f8ba590",
    "executions": [
      {
        "trigger": "Application",
        "vmstate": "HALT",
        "exception": null,
        "gasconsumed": "2645280",
        "stack": [
          {
            "type": "ByteString",
            "value": "SGVsbG8sIENocmlz"
          }
        ],
        "notifications": [],
        "logs": [ // this line here
          {
            "contract": "0x4169aaf2c60776c1ee062c84f1a1502a49841702",
            "message": "Hello, Chris"
          }
        ]
      }
    ]
  }
}
```

Add `LogEventArgs` output to `ApplicationLog` plugin and to RCPClient.

- Plugins
- RPC (HTTP)

Jim8y:
Closes #841 

This pr focus on adding a debug mode to the applicationlog plugin such that developers can get the transaction execution log via the rpc interface.

What has changed?

- The config adds a debug field.
- Added Log monitor logic.
- Logs field will be added to the json response under debug mode.
- Update the rpcclient to support logs in response.

Will this impact other tools?

No

### DIFF
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -214,9 +214,11 @@ namespace Neo.Plugins
             {
                 logList = (JArray)JToken.Parse(Neo.Utility.StrictUTF8.GetString(value));
             }
-            var logJson = new JObject();
-            logJson["contract"] = args.ScriptHash.ToString();
-            logJson["message"] = args.Message;
+            var logJson = new JObject
+            {
+                ["contract"] = args.ScriptHash.ToString(),
+                ["message"] = args.Message
+            };
             logList?.Add(logJson);
 
             _snapshot.Put(LogPrefix.Concat(tx.ToArray()).ToArray(), Neo.Utility.StrictUTF8.GetBytes(logList?.ToString()!));
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -24,8 +24,9 @@ namespace Neo.Plugins
 {
     public class LogReader : Plugin
     {
-        private IStore _db;
-        private ISnapshot _snapshot;
+        private static IStore _db;
+        private static ISnapshot _snapshot;
+        private static readonly byte[] LogPrefix = { 0x12 };
 
         public override string Name => "ApplicationLogs";
         public override string Description => "Synchronizes the smart contract log with the NativeContract log (Notify)";
@@ -34,12 +35,16 @@ namespace Neo.Plugins
         {
             Blockchain.Committing += OnCommitting;
             Blockchain.Committed += OnCommitted;
+            if (Settings.Default.Debug)
+                ApplicationEngine.Log += ApplicationEngine_Log;
         }
 
         public override void Dispose()
         {
             Blockchain.Committing -= OnCommitting;
             Blockchain.Committed -= OnCommitted;
+            if (Settings.Default.Debug)
+                ApplicationEngine.Log -= ApplicationEngine_Log;
         }
 
         protected override void Configure()
@@ -75,6 +80,12 @@ namespace Neo.Plugins
                     else
                         i++;
                 }
+            }
+            if (!Settings.Default.Debug) return raw;
+            byte[] logs = _db.TryGet(LogPrefix.Concat(hash.ToArray()).ToArray());
+            if (logs != null)
+            {
+                raw["executions"]["logs"] = (JArray)JToken.Parse(Neo.Utility.StrictUTF8.GetString(value));
             }
             return raw;
         }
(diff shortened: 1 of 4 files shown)


Liaojinghui:
@cschuchardt88 just finished coding, not tested yet, but you can take a look. Will finish it tomorrow.

cschuchardt88:
Still need to update RpcClient `RpcApplicationLog` Class. `neo-express` uses that for `FromJson`

shargon:
Why only during debug? It seems reasonable to get this information always

Liaojinghui:
> Why only during debug? It seems reasonable to get this information always

 This changes the response message size and processing time, without a solid test, i dare not add it directly in fear of DOS attacks. So, debug mode first. But we can also go straightforward to add it for production env if you think it is OK.

vncoelho:
In this case, test the potential DOS first would be good. And leave as not default for now.

Liaojinghui:
@cschuchardt88 this will not be merged until we have a monorepo.

cschuchardt88:
@Liaojinghui @vncoelho PR #807 should be able to handle any amount of logs.

@Liaojinghui also im updating PR #807 for these new changes. Can you give an example template of the json? or is it the same as what I posted in #841

Liaojinghui:
> @Liaojinghui @vncoelho PR #807 should be able to handle any amount of logs.
> 
> 
> 
> @Liaojinghui also im updating PR #807 for these new changes. Can you give an example template of the json? or is it the same as what I posted in #841 

Exactly the same as you posted

src/ApplicationLogs/LogReader.cs
```diff
@@ -186,6 +197,33 @@ private void OnCommitting(NeoSystem system, Block block, DataCache snapshot, IRe
             }
         }
 
+        private static void ApplicationEngine_Log(object sender, LogEventArgs args)
+        {
+            if (!Settings.Default.Debug) return;
+
+            var tx = ((Transaction)args.ScriptContainer).Hash;
```
AnnaShaleva, line 204: It can be not only a transaction, right? E.g. block for OnPersist and PostPersist execution results.

src/ApplicationLogs/LogReader.cs
```diff
@@ -186,6 +197,33 @@ private void OnCommitting(NeoSystem system, Block block, DataCache snapshot, IRe
...
+            else
+            {
+                logList = (JArray)JToken.Parse(Neo.Utility.StrictUTF8.GetString(value));
+            }
+            var logJson = new JObject
+            {
+                ["contract"] = args.ScriptHash.ToString(),
+                ["message"] = args.Message
+            };
+            logList?.Add(logJson);
```
AnnaShaleva, line 222: Why do you need `?` here? `logList` is always non-null.

src/ApplicationLogs/Settings.cs
```diff
@@ -18,13 +18,16 @@ internal class Settings
         public uint Network { get; }
         public int MaxStackSize { get; }
 
+        public bool Debug { get; }
```
AnnaShaleva, line 21: Are we planning to enable this setting in public networks?

src/ApplicationLogs/LogReader.cs
```diff
@@ -34,12 +35,16 @@ public LogReader()
         {
             Blockchain.Committing += OnCommitting;
             Blockchain.Committed += OnCommitted;
+            if (Settings.Default.Debug)
+                ApplicationEngine.Log += ApplicationEngine_Log;
```
AnnaShaleva, line 39: It's a nice feature. How about adding the same ability to add logs to execution result for `invokescript`, `invokefunction` and etc. in debug mode? This change is more invasive and requires changing the core library as far, but it might be useful for someone.

cschuchardt88: @AnnaShaleva Also https://github.com/neo-project/neo-modules/pull/807#issuecomment-1807590923

src/ApplicationLogs/LogReader.cs
Liaojinghui: > Why do you need `?` here? `logList` is always non-null.

It is possible since in the `else` block, the assignment might be null.

src/ApplicationLogs/LogReader.cs
Liaojinghui: > It can be not only a transaction, right? E.g. block for OnPersist and PostPersist execution results.

I think it can only be transaction, but i might be wrong. May you please confirm @shargon @roman-khimov

PR #815, opened 2023-08-29
Implement RPC errors proposal #156

Jim8y:
https://github.com/neo-project/neo-modules/issues/728
https://github.com/neo-project/neo-modules/issues/814

- [x] https://github.com/neo-project/neo-modules/pull/873

PR #845, opened 2023-11-17
Ensure max length

shargon:
Related to: https://github.com/neo-project/neo-modules/pull/827#issuecomment-1815273757

### DIFF
--- a/src/RpcServer/RpcServer.SmartContract.cs
+++ b/src/RpcServer/RpcServer.SmartContract.cs
@@ -166,6 +166,11 @@ namespace Neo.Plugins
 
         private static Signer[] SignersFromJson(JArray _params, ProtocolSettings settings)
         {
+            if (_params.Count > Transaction.MaxTransactionAttributes)
+            {
+                throw new RpcException(-100, "Max allowed witness exceeded.");
+            }
+
             var ret = _params.Select(u => new Signer
             {
                 Account = AddressToScriptHash(u["account"].AsString(), settings.AddressVersion),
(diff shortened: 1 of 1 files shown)


Liaojinghui (approved):

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -166,6 +166,11 @@ private static JObject ToJson(StackItem item, Session session)
 
         private static Signer[] SignersFromJson(JArray _params, ProtocolSettings settings)
         {
+            if (_params.Count > Transaction.MaxTransactionAttributes)
+            {
+                throw new RpcException(-100, "Max allowed witness exceeded.");
```
AnnaShaleva, line 171: > We can wait for https://github.com/neo-project/neo-modules/pull/815

Agree, some InvalidParams error can be used here. I'll review #815 one more time ASAP.

cschuchardt88:
have i max allowed witness conditions been fixed in neo-core or here? like `and` `boolean` `not` etc

shargon:
> have max allowed witness conditions been fixed in neo-core or here? like `and` `boolean` `not` etc

Must be on both, here is composed manually, not deserialized.

Liaojinghui:
@shargon , grand master, can we have a sole project where all limits are set together? the existing limitations are like here and there, hard to follow and track.

shargon:
> @shargon , grand master, can we have a sole project where all limits are set together? the existing limitations are like here and there, hard to follow and track.

Agree, but the limits are during deserialization, and here the object is composed manually

shargon:
I will merge it, #815 is delayed and it can replace this throw later

PR #846, opened 2023-11-20
Refac build configs

Ashuaidehao:
Close https://github.com/neo-project/neo-modules/issues/848

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,7 +2,7 @@
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 	<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
 	<ItemGroup>
-		<None Include="config.json" Condition="Exists('config.json') And '$(IncludeSettingsFileOutput)' != 'False'">
+		<None Include="config.json"  Condition="Exists('config.json')">
 			<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
 			<CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
 		</None>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,27 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
-  <PropertyGroup>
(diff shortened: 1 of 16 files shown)


Liaojinghui:
Why is this pr?

superboyiii:
See https://github.com/neo-project/neo-modules/issues/848 @shargon @Liaojinghui

Liaojinghui (approved):

Liaojinghui:
Behaves the same as description #848

vncoelho (commented):
Looks good change!

vncoelho:
I will compile here and see results.

Liaojinghui (dismissed):

superboyiii:
Choose #849 or this? They have the same behaviour.

shargon:
This was first, isn't it?

Liaojinghui:
> This was first, isn't it?

Yes

cschuchardt88:
however when you `dotnet build` it will build all the files. and i thought you wanted no dependencies in the output. #849 doesn't have that problem in workflow actions.

src/Directory.Build.props
```diff
@@ -1,27 +1,10 @@
...
-  </ItemGroup>
-
-  <ItemGroup>
-    <PackageReference Include="Neo" Version="3.6.2" />
-  </ItemGroup>
-
-</Project>
+	<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
+	<ItemGroup>
+		<None Include="config.json" Condition="Exists('config.json') And '$(IncludeSettingsFileOutput)' != 'False'">
```
AnnaShaleva, line 5: We've excluded `IncludeSettingsFileOutput` from every project, so I think we don't need this condition (`'$(IncludeSettingsFileOutput)' != 'False'`) to be presented here, it's just not relevant anymore.

Is `config.json` a configuration file of the module's dependent library? If so, then we probably just don't need to copy it while publishing module.

AnnaShaleva (commented):
Nice enhancement, we'll port it to neo-bench.

src/Directory.Build.props
Liaojinghui: > We've excluded `IncludeSettingsFileOutput` from every project

Not every project, just part of them.

AnnaShaleva: I mean from every neo-modules project. We don't have any mentions of `IncludeSettingsFileOutput=False` anymore in neo-modules.

AnnaShaleva (approved):

shargon (approved):

Liaojinghui:
@Ashuaidehao this breaks the publish process, please check and fix.

PR #847, opened 2023-11-23
RPC-JSON over WebSockets

### LINKED ISSUE #716: Use a session protocol for some JSON-RPC methods

cschuchardt88:
Description will be updated soon....

### Some things to know
- URL(s) - path and query can be anything. (_Note: use `OnRequest` event to get custom paths/headers and query params_)
  - `ws://127.0.0.1:10340/`
  - `ws://127.0.0.1:10340/?MyCustomParam=HelloWorld`
  - `ws://127.0.0.1:10340/MyCustomPath/FakePage.html?say=Hello%20Bob`
- Uses `Neo.Json`
- Has expandability of API in other plugins.
- Emits events that plugins can hook into for more information on clients.
- 99% uptime (for clients) (unless `neo-core` fails)
- It's up to clients to filter events
- Emits block/transaction/contract events to clients
- RPC like functionally over `WebSocket`

**Events and their IDs**
```csharp
public enum WebSocketResponseMessageEvent : byte
{
    Block = 0x02,               // Blockchain.Committed
    MemoryPool = 0x03,          // MemoryPool
    Transaction = 0x05,         // Transaction
    ContractNotify = 0x14,      // ApplicationEngine.Notify
    AppLog = 0x15,              // ApplicationEngine.Log
    DebugLog = 0x16,            // Utility.Log
    Error = 0x19,               // Error(s)
    Method = 0x20,              // Result(s) from a method call
}
```

**Request information**
```json
{
    "version": "1.0",       // Protocol version
    "requestid": 999,       // For tracking which response goto which request (-1 for system error)
    "method": "getblock",   // the name of the method to invoke
    "params": [             // parameters for the method
        "0"
    ]
}
```

**config.json**
```json
{
  "PluginConfiguration": {
    "Network": 860833102,
    "BindAddress": "127.0.0.1",
    "Port": 10340,
    "SslCertFile": "",
    "SslCertPassword": "",
    "TrustedAuthorities": [],
    "EnableBasicAuthentication": false,
    "User": "",
    "Pass": "",
    "AllowOrigins": [],
    "ConcurrentProxyTimeout": 120,
    "MessageSize": 4096,
    "MaxGasInvoke": 2000000000,
    "DebugMode": false
  }
}
```

### Events

```csharp
WebSocketServerPlugin.OnRequest(HttpContext httpContext)               // Emits after valid WebSocket connection request is made
WebSocketServerPlugin.OnServerStarted()                                // Emits after WebSocket host is up
WebSocketConnection.OnMessageSent(Guid clientId, JToken json)          // Emits after message is successfully sent to client
WebSocketConnection.OnMessageReceived(Guid clientId, JToken json)      // Emits after a message is received and serialized
WebSocketConnection.OnConnect(Guid clientId)                           // Emits after a successful connection is made
WebSocketConnection.OnDisconnect(Guid clientId)                        // Emits after a successful disconnect
```

### Methods

```csharp
WebSocketServerPlugin.RegisterMethods(object handler)                                      // Registers methods for a given class with attribute WebSocketMethod
WebSocketServerPlugin.SendAllJson(JToken json, int pluginId)                               // Sends to all clients
WebSocketServerPlugin.SendJson(Guid clientId, int requestId byte eventId, JToken json)     // Send to a specified WebSocket by client id, (note: use "OnConnect" event to get client id)
```

### Properties

```csharp
WebSocketServerPlugin.HasClients        // Whether the server has valid connections to clients
```

### Creating a method in `Code`

```csharp
[WebSocketMethod]
private JToken Echo(JArray @params) // turns into all lowercase for method name
{
    return @params;
}
// -----OR-----
[WebSocketMethod(Name = "echo")] // name is case sensitive
private async Task<JToken> EchoAsync(JArray @params) // method would be "echoasync`, if not specified by "Name" in attribute
{
    return @params;
}
```

### Video of Syncing (Blockchain)

https://github.com/neo-project/neo-modules/assets/8141309/22c5cd49-8e81-4ba9-a11b-5144cbe8b4b9

### Screenshots

**Block Event (on commit)**
Screenshot 2023-11-22 184425

**Invoke Response (Echo Method)**
Screenshot 2023-11-22 184405

**Contract Response (on syncing/new block)**
image

**Transaction Response (on syncing/new block)**
image

**Block Response (on syncing/new block)**
image

**Change Log**
- Added `WebSocketServer`

closes #716


roman-khimov:
I think community will love two absolutely different websocket interfaces.

cschuchardt88:
This would make 3? Because `neo-cli` node has one build in too _(but thats for syncing blockchain)_. But this is meant to replace #795. Now are you talking about `neo-go`'s? If that is the case maybe we can build the same way in this PR. What my goal is to allow users to have a fully functioning `API` that does everything that `neo-cli` and `neo-express` does. The reason behind this; is to be able to support other platforms and programming languages along with special user cases. The way that `neo-core` sits right now... it can't be done easily; if not at all. But with the help of this and other people expanding this `API`. We can tackle `web`, `dotnet` and other ecosystems. This also allows others to make plugins to expand the `api` for custom in house applications that developers or organizations may need for there only business needs. I think with add-on on this new sidechain; it would be time for ALL developers and users to be able to use `neo` if they choose too. `WebSockets` would be the easiest way to start. With `web3` most people want their own `dapp` to be able to get blockchain information and functionally. Currently you would have to build your own.

cschuchardt88:
What this feature offers, is **MUST** to have. My other reasoning is that with `neo3-visual-tracker` (vscode-extension) use 3rd party javascript library (neon-js) for accessing the `RPCServer` of `neo-express`. Now this is very buggy as we know. Could be the because of old version. With this websocket API we can easily get new blocks and other information that is needed for this extension. I am currently rebuilding this extension, and need this plugin, or else I'll make it just for `neo-express`, But i thought the features it offers would be a good choice for community and ecosystem.

cschuchardt88:
This can wait @shargon for now, im building out the wallet api for it.

src/WebSocketServer/WebSocketUtility.cs
```diff
@@ -16,5 +17,12 @@ public static UInt160 TryParseScriptHash(string addressOrScriptHash, byte addres
 
             return scriptHash ?? UInt160.Zero;
         }
+
+        public static BigInteger TryParseBigInteger(string value)
```
Jim8y, line 21: Why having this instead of using tryparse directly?

cschuchardt88: I dont have the `3` line of code over and over again. Code redundancy is bad.

src/WebSocketServer/WebSocketServerPlugin.cs
```diff
@@ -85,7 +85,8 @@ public override void Dispose()
         {
             _host?.Dispose();
             _connections?.Dispose();
-            _neoSystem.MemPool.TransactionAdded -= OnMemPoolTransactionAdded;
+            if (_neoSystem != null)
```
Jim8y, line 88: should not you just  stop processing if _neoSystem is null

cschuchardt88: No, if `neosystem` is not null then we need to remove the handler, or else it will crash on a new item sent to the `mempool`.

Jim8y: Ur right, I did not notice its dispose.

src/WebSocketServer/WebSocketClient.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Neo.Plugins
+{
+    internal class WebSocketClient : IDisposable, IEquatable<WebSocketClient>
+    {
+        public WebSocket Socket { get; init; }
```
shargon, line 12: required and remove null checks?

cschuchardt88: It is not null there. Hints the reason why no `?` and `init` on property. We don't want null.

cschuchardt88:
@superboyiii  If this PR can you on your list ASAP for testing. I need this PR to update the UI for `neo-express` (vscode extension). The extension is pretty bad at the moment. It unworkable. All the project and templates it generates are useless for any new developer; even those that don't know C#.

src/WebSocketServer/WebSocketClient.cs
shargon: > It is not null there. Hints the reason why no `?` and `init` on property. We don't want null.

Yep, but it's used with ? inside the class

cschuchardt88: Right and that reason, is for if client disconnect and the class becomes `disposed`.

src/WebSocketServer/WebSocketConnection.cs
```diff
@@ -13,10 +13,9 @@
 
 namespace Neo.Plugins
 {
-    internal class WebSocketConnection<TClient> : IDictionary<Guid, TClient>, IDisposable
-        where TClient : WebSocketClient, new()
```
cschuchardt88: Need to change this back. `WebSocketClient` is a `base` class. You can extend the class by inheriting it. For example if you want to add more properties/fields or methods to it. Like a tracking number or Client's login information.

cschuchardt88 (commented):
Please fix @shargon

src/WebSocketServer/WebSocketClient.cs
shargon: I marked the project as nullable and I detected it, I was going to fix it, I think that new projects should be nullable

shargon (changes requested):
Please set the project as nullable

src/WebSocketServer/WebSocketConnection.cs
shargon: > Need to change this back. `WebSocketClient` is a `base` class. You can extend the class by inheriting it. For example if you want to add more properties/fields or methods to it. Like a tracking number or Client's login information.

We can extend WebSocketClient, we don't need more types of WebSocketClients, sometimes simpler is better

cschuchardt88: Yes in theory. But this also limits the API or makes more work in the future. If the code is already there or built for reusability, than the code should stay.

shargon: > If the code is already there or built for reusability, than the code should stay.

I'm not agree, we don't need two WebSocketClients

src/WebSocketServer/WebSocketServerPlugin.cs
```diff
@@ -316,8 +316,8 @@ private void StartWebSocketServer()
                         {
                             if (err != SslPolicyErrors.None)
                                 return false;
```
shargon, line 318: If chain is null ||

src/WebSocketServer/v1/BlockchainMethods.cs
```diff
@@ -198,16 +198,16 @@ public JToken InvokeContract(JArray _params)
 
             try
             {
-                if (_params.Count >= 1 && UInt160.TryParse(_params[0].AsString(), out scriptHash))
+                if (_params.Count >= 1 && UInt160.TryParse(_params[0]?.AsString(), out scriptHash))
```
shargon, line 201: !

src/WebSocketServer/WebSocketClient.cs
```diff
@@ -0,0 +1,86 @@
...
+using System.Net.WebSockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Neo.Plugins
+{
+    internal class WebSocketClient : IDisposable, IEquatable<WebSocketClient>
+    {
+        public WebSocket? Socket { get; init; }
```
shargon, line 12: We don't need this with ?, only is needed because you want a T type that we don't need

src/WebSocketServer/WebSocketServerPlugin.cs
cschuchardt88: Specifies a callback for additional client certificate validation that will be invoked during authentication. This will be ignored if [AllowAnyClientCertificate()](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.https.httpsconnectionadapteroptions.allowanyclientcertificate?view=aspnetcore-7.0#microsoft-aspnetcore-server-kestrel-https-httpsconnectionadapteroptions-allowanyclientcertificate) is called after this callback is set.

src/WebSocketServer/WebSocketClient.cs
cschuchardt88: I don't understand. Referring to `T` type; we do need it! This a `BASE` CLASS. NOT DOING SO WILL LIMIT AND FLAW EXPANDABLE OF AND POINT OF THIS LIBRARY. If you want to keep limiting the `api` we can do that. Just look at `RpcServer` is outdated and cannot be expanded or updated easily. Which is why no one touches it and it in the state it's in now.

vncoelho (commented):
remainder to check https://github.com/neo-project/neo-modules/pull/859, remove websocker

src/WebSocketServer/WebSocketServerPlugin.cs
cschuchardt88: So no need to check `null`. `null` isnt passed.

AnnaShaleva (commented):
@shargon, @Jim8y, @vncoelho since the work on this PR is continuing, I'd like to pay your attention one more time to the problem mentioned by @roman-khimov in https://github.com/neo-project/neo-modules/pull/847#issuecomment-1823860165 and to the comment of @cschuchardt88 in https://github.com/neo-project/neo-modules/pull/847#issuecomment-1823874448.

The case is that NeoGo node already has WS server that handles all standard RPC methods *exactly the same way* as regular RPC server does (using exactly the same response structures and without handlers code duplication, see the comment below). Also, NeoGo WS server has subscriptions interface and notifications subsystem implemented for *more than 3 years* (check out the https://github.com/nspcc-dev/neo-go/blob/master/docs/notifications.md and code connected with https://github.com/nspcc-dev/neo-go/blob/d90169761595933febb8a25d1abde2e46154b96d/pkg/services/rpcsrv/server.go#L265). I know that there are community users of the existing NeoGo WS system (and, in particular, WS notification subsystem), and another WS interface implementation would be a problem for our users. We need a single unified and aligned websocket interface. What do you think?

src/WebSocketServer/v1/BlockchainMethods.cs
```diff
@@ -0,0 +1,251 @@
...
+        public BlockchainMethods(
+            NeoSystem neoSystem)
+        {
+            _neoSystem = neoSystem;
+            _localNode = _neoSystem.LocalNode.Ask<LocalNode>(new LocalNode.GetInstance()).Result;
+            WebSocketServerPlugin.RegisterMethods(this);
+        }
+
+        [WebSocketMethod]
+        public JToken GetBlockHeader(JArray _params)
```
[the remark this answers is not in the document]
cschuchardt88: Go back and read main comment. As you know PRs are slow. The plain is to just add `[WebSocketMethod]` to the `RPC` methods. However, we are not doing that because `RpcServer` is outdated and unmaintained. We are going to be build a central library to handle all the common methods. Instead of fixing `RpcServer` poor logic structure. I did so here, in this PR. Copy have to copy out `api` methods to tht central library. With this implementation of `WebSockets` you remove the need for running your own node. You just connect and that its. It has its own protocol built-in like the `nodes` P2P, but is not P2P. This plugin sends all `node` events to the client. No need for subscriptions. Could in theory use the data for syncing. All this is in real-time. Plus this is `core` is it not? That's like saying just cause `coz` added and had a `RestServer` 1st we should follow their logic. I waited a long time for `roman-khimov ` to reply and i never heard anything back.... I'll leave it as that for now.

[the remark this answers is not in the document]
roman-khimov: > RpcServer is outdated and unmaintained.

Really? Where exactly it's not maintained and how come it's outdated?

> With this implementation of WebSockets you remove the need for running your own node

Different problem not related to WS/JSON-RPC/REST/Protobuf/whatever.

> This plugin sends all node events to the client

Which is a problem right away: clients must be able to choose things they're interested in, public servers shouldn't be bothered with sending everything to everyone.

I don't see any value in this implementation, sorry. We do have well-established JSON-RPC interface. By "well-established" I mean that all of the community (wallets/SDKs/whatever from COZ/AxLabs/etc) knows how to interact with it and has a lot of tooling built around it. And then there are dApps built around that tooling. This has a lot of value.

Now for whatever reason you're trying to tell us that JSON-RPC interface is outdated/unmaintained, we should throw it away and replace with something from this PR. Does it mean that all of the tooling built around the JSON-RPC suddenly turns to dust? Should be rewrite everything, including all of the dApp backends? What's the value for users in doing so? What's gonna be the next rewrite? REST/protobuf/anything else?

https://github.com/nspcc-dev/neo-go/blob/master/docs/notifications.md uses a different approach, it's 100% compatible with the current JSON-RPC (JSON-RPC itself is transport-agnostic, so this costs nothing) and gives some new abilities. Extending existing tooling for it is trivial and there are applications already using it (the extension itself dates back to Legacy days). Which doesn't make it a standard, sure, you can create ten other variations around this theme, but please think about the value you're trying to provide to people.

See https://github.com/neo-project/neo/issues/2853 also.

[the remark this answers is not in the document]
vncoelho: I am in agreement with the idea of being 100% compatible with RPC plugin.

It is the same for Rest Plugin as well, the best scenario is that we have a common set of methods used by these interfaces.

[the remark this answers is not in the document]
cschuchardt88: @vncoelho 
Ee are merging this repo to `neo` monorepo. Me and @Jim8y where going to do one library for all common logic. But after the `RPC` proposal changes _PR_. That's why I programed the logic kind of the same with `RpcServer` `RpcMethodName` attribute. So you can just add `WebSocketMethod` and be done. Also we can update all the other `Plugins` too; just by adding `WebSocketMethod` attribute.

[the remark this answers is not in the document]
cschuchardt88: >I don't see any value in this implementation, sorry. We do have well-established JSON-RPC interface. By "well-established" I mean that all of the community (wallets/SDKs/whatever from COZ/AxLabs/etc) knows how to interact with it and has a lot of tooling built around it. And then there are dApps built around that tooling. This has a lot of value.

Limiting the community and developers isn't going to make anything better. We need a good solid reason why with `pros` and `cons`. Other than your feelings why something shouldn't or should be. This PR isn't about `RPC-JSON`; it's about `Websockets`.

>Which is a problem right away: clients must be able to choose things they're interested in, public servers shouldn't be bothered with sending everything to everyone.

Lets overload the server with `HTTP` `POST` requests or with logic that creates events and filters data, so someone can `DDOS` Or have millions of `subscription` events. That would be `x` times each client per `x` number of clients vs the `x` servers. Do the math; it's simpler and cheaper with bandwidth then `CPU` time or `RAM`. 

>Now for whatever reason you're trying to tell us that JSON-RPC interface is outdated/unmaintained, we should throw it away and replace with something from this PR

No one said that. You miss read it or misunderstand. So let me clear that up. We still need `RPC-JSON`. The code in this repo is outdated. Look how long it takes to get a `RPC-JSON` bug merged. Some of those clients or users you talk about using `RPC-JSON` had to wait months for a bug fix to merge for #812; that should of had a `hotfix` within two weeks or reasonable timeframe.

This PR isn't meant to replace `RPC-JSON` or `RestServer`. It meant for accessing or exposing anything a `Plugin` developer needs in `WebSockets` with runtime communication between you and the server. And again this isn't `RPC-JSON` over `WebSockets`. This is like `SignalR` or `WebSockets` in general; type of communication. If we want `RPC-JSON` runtime communication we can just use `HTTP/2` or `HTTP/1.1` to do that. So why use `WebSockets` for something that can already be done with `HTTP/2` streaming` or `HTTP/1.1` `KeepAlive`?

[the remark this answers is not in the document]
vncoelho: > @vncoelho 
> We are merging this repo to `neo` monorepo. Me and @Jim8y where going to do one library for all common logic. But after the `RPC` proposal changes _PR_. That's why I programed the logic kind of the same with `RpcServer` `RpcMethodName` attribute. So you can just add `WebSocketMethod` and be done. Also we can update all the other `Plugins` too; just by adding `WebSocketMethod` attribute.

Lets merge first. That is the best choice.
I am stuck with Consensus UT because of that.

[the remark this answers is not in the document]
vncoelho: Anybody that do the PR merging the modules will be good, we just need to merge that asap to move forward on other topics as well.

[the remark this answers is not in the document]
roman-khimov: > We need a good solid reason why with pros and cons

Sure, it's as easy as having the same API over different transport. If we have JSON-RPC with one API, WS with another, then whatever else with third one then we'll have to maintain a set of incompatible things and people will have to think about these differences when creating their dApps.

> it's simpler and cheaper with bandwidth

No. Bandwidth costs a lot. Filtering events is easy, sending them over some link is much more overhead both for server and client. From what I've seen what most of dApp developers want is:
 * get notifications from their contracts
 * monitor transactions they send (acceptance/execution result)

Always pushing a stream of data that's 99% irrelevant for these use cases is a resource waste. Providing just an event stream over WS and resorting to a different JSON-RPC client for other requests is a headache.

> Some of those clients or users you talk about using RPC-JSON had to wait months for a bug fix to merge for https://github.com/neo-project/neo-modules/pull/812; that should of had a hotfix within two weeks or reasonable timeframe.

I agree with that, these problems should be solved much faster, but I'm not sure how adding yet another plugin solves the problem. To me it only makes things worse, it's just more code to maintain.

> It meant for accessing or exposing anything a Plugin developer

Well, it's plugin already, so to me, it should provide some value to the end user. If I install this plugin and the only thing I get is just an event stream then how do I integrate my dApp with it? Why should it be a part of the core? If you only need an unfiltered event stream for neo-express then it might as well be a plugin maintained as a part of neo-express. Just as an example, there is a https://github.com/Hecate2/neo-fairy-test/ maintained outside of the core.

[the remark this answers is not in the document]
cschuchardt88: You do make good points. I understand completely. At 1st this PR was just `WebSocketServer` hints the name. Only acted as a `EventServer` for `node` events. See https://github.com/neo-project/neo/issues/2996 for full story.

>No. Bandwidth costs a lot. 

Here in the USA. It's almost free
image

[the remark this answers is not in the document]
vncoelho: Here in Brazil it is free. :D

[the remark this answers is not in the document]
roman-khimov: That's not the cost I'm talking about, who cares about those dollar bills... I'm more about the tx/rx overhead, latencies and channel capacity. Not transmitting even 1M of data is much much better than transmitting it, filtering events is some KBs of memory and nanoseconds of CPU time, communicating them to the client is milliseconds easily. Then there is a slow reader (client) problem which is much more likely to happen if event stream isn't filtered.

This approach is OK for some development environment where you have a node locally that's not doing a lot and just want to get everything from it. But this doesn't make for a good general-purpose plugin. So my suggestion is to either make it usable for dApps or move it out of the core.

shargon:
Repository moved to [neo](https://github.com/neo-project/neo), please re-open there

PR #849, opened 2023-11-24
fixed output of deps

### LINKED ISSUE #848: Remove unnecessary dll when build or publish a plugin
Our plugins have a complex dependency structure, for example, StateService need MPT, need RpcServer, need Neo, and Neo need Neo-VM, so when you build or publish a plugin, you may get many dll from these dependency, but what we what is just these necessary when taken them into `Plugins` path of neo-cli.

Plugins excluding `RpcServer` itself, don't need build `RpcServer` in publish,  `Neo` and `Neo.ConsoleService` as welll.
- Plugins

cschuchardt88:
The way this works.
- `dotnet build neo-modules.sln --configuration Release`
- outputs to `$(ProjectDir)/bin/dist`

### Workflow example
```yml
    - name: Zip (TokensTracker)
      working-directory: ./src/TokensTracker/bin/dist
      run: Compress-Archive -Path * -Destination TokensTracker-v3.6.3.zip

```

image

@superboyiii 

closes #848

### DIFF
--- a/src/ApplicationLogs/ApplicationLogs.csproj
+++ b/src/ApplicationLogs/ApplicationLogs.csproj
@@ -6,7 +6,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\RpcServer\RpcServer.csproj" AdditionalProperties="IncludeSettingsFileOutput=False"/>
+    <ProjectReference Include="..\RpcServer\RpcServer.csproj" AdditionalProperties="IncludeSettingsFileOutput=False">
+      <CopyLocalSatelliteAssemblies>False</CopyLocalSatelliteAssemblies>
+      <Private>False</Private>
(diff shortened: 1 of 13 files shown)


superboyiii (approved):
It works well for me.
It's the same behaviour as #846, so we need choice 1 @Liaojinghui @shargon

Liaojinghui:
> It works well for me. It's the same behaviour as #846, so we need choice 1 @Liaojinghui @shargon

I pick the earlier one if they behave the same, unless with specific reasons. My principle is first come first get.

cschuchardt88:
@superboyiii with #846 run this command `dotnet build neo-modules.sln` check output file or use `dotnet build neo-modules.sln --output ./dist` you will get all the dlls.

cschuchardt88:
nvm

Issue #850, opened 2023-11-24
[3.6.2] GetNativeContracts Exception


Devel484:
**Describe the bug**
Calling the RPC Method GetNativeContracts results in an exception on client side. RPC side is sending a json object without the field `updatehistory`, but Client side is expecting the field.

https://github.com/neo-project/neo-modules/blob/64a1ae4c27b2a8d1285976231142b39bde49cc02/src/RpcClient/Models/RpcNativeContract.cs#L26-L36

https://github.com/neo-project/neo-modules/blob/64a1ae4c27b2a8d1285976231142b39bde49cc02/src/RpcServer/Utility.cs#L35-L44

shargon:
I will remove it also in client according to https://github.com/neo-project/neo-modules/pull/828

PR #851, opened 2023-11-24
Fix RpcNativeContract

shargon:
Close https://github.com/neo-project/neo-modules/issues/850


src/RpcClient/Models/RpcNativeContract.cs
```diff
@@ -21,7 +21,6 @@ public class RpcNativeContract
         public UInt160 Hash { get; set; }
         public NefFile Nef { get; set; }
         public ContractManifest Manifest { get; set; }
-        public uint[] UpdateHistory { get; set; }
```
AnnaShaleva, line 24: Oh, we need monorepo...

AnnaShaleva (approved):

PR #853, opened 2023-11-29
v3.6.3

superboyiii:
Hotfix version for https://github.com/neo-project/neo-modules/issues/850 which breaks Ghostmarket workflow.
@shargon @roman-khimov @AnnaShaleva


AnnaShaleva (approved):
I thought we're going to include this fix right into 3.7, but if it's critical for Ghostmarket, then probably it's OK to have it in a separate release.

superboyiii:
> I thought we're going to include this fix right into 3.7, but if it's critical for Ghostmarket, then probably it's OK to have it in a separate release.

It's critical for them.

roman-khimov (approved):

Issue #854, opened 2023-11-30
Tx showed HALT in ApplicationLogs but has exception returned


superboyiii:
**Describe the bug**
`0x0819175f7cd906f2eee8fbc825496acd11f0f0afcfcaec32bfbb71a7a52b6f48` on testnet have such an applicationlog below:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0x0819175f7cd906f2eee8fbc825496acd11f0f0afcfcaec32bfbb71a7a52b6f48",
        "executions": [
            {
                "trigger": "Application",
                "vmstate": "HALT",
                "exception": "Operation is not valid due to the current state of the object.",
                "gasconsumed": "6312969",
                "notifications": []
            }
        ]
    }
}
```
The vmstate is `HALT`, but the exception is not null. This can make people confusing.

**To Reproduce**
![1701312079559](https://github.com/neo-project/neo-modules/assets/38493437/0f632bbb-9ef7-4fb8-ba76-2dd75dbab8d4)

**Expected behavior**
`exception` is not null even `vmstate` is null.

**Screenshots**
Showed above

**Platform:**
 - OS: [all]
 - Version [e.g. neo-cli 3.6.3]

roman-khimov:
```
$ curl -sd '{ "jsonrpc": "2.0", "id": 1, "method": "getapplicationlog", "params": ["0x0819175f7cd906f2eee8fbc825496acd11f0f0afcfcaec32bfbb71a7a52b6f48"] }' https://rpc.t5.n3.nspcc.ru:20331 | json_pp
{
   "id" : 1,
   "result" : {
      "txid" : "0x0819175f7cd906f2eee8fbc825496acd11f0f0afcfcaec32bfbb71a7a52b6f48",
      "executions" : [
         {
            "gasconsumed" : "6312969",
            "vmstate" : "HALT",
            "stack" : [
               "error: unserializable: nil",
               "error: unserializable: nil"
            ],
            "notifications" : [],
            "exception" : null,
            "trigger" : "Application"
         }
      ]
   },
   "jsonrpc" : "2.0"
}
```

cschuchardt88:
This is normal expected behavior. The reason why this exception is happening is because the `Stackitem` is an `InteropInterface`, that can't be serialized. Which indicates that VM is not crashing or having problems.

PR #696, opened 2022-03-04
Limit result stack

shargon:
Alternative to https://github.com/neo-project/neo-modules/pull/695
Close https://github.com/neo-project/neo/issues/2666

PR #855, opened 2023-11-30
Fix stack exception name

superboyiii:
Close https://github.com/neo-project/neo-modules/issues/854

### DIFF
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -95,7 +95,7 @@ namespace Neo.Plugins
             {
                 try
                 {
-                    stack.Add(item.ToJson());
+                    stack.Add(item.ToJson(Settings.Default.MaxStackSize));
                 }
                 catch (InvalidOperationException)
                 {
@@ -143,7 +143,7 @@ namespace Neo.Plugins
                     {
                         try
                         {
-                            stack.Add(item.ToJson());
+                            stack.Add(item.ToJson(Settings.Default.MaxStackSize));
                         }
                         catch (InvalidOperationException)
                         {
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -90,14 +90,19 @@ namespace Neo.Plugins
             trigger["vmstate"] = appExec.VMState;
             trigger["exception"] = GetExceptionMessage(appExec.Exception);
             trigger["gasconsumed"] = appExec.GasConsumed.ToString();
-            try
+            var stack = new JArray();
+            foreach (var item in appExec.Stack)
             {
-                trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
+                try
+                {
+                    stack.Add(item.ToJson());
+                }
+                catch (InvalidOperationException)
+                {
+                    stack.Add("error: invalid operation");
+                }
             }
-            catch (InvalidOperationException)
-            {
-                trigger["stack"] = "error: invalid operation";
-            }
+            trigger["stack"] = stack;
             trigger["notifications"] = appExec.Notifications.Select(q =>
             {
                 JObject notification = new JObject();
@@ -133,13 +138,17 @@ namespace Neo.Plugins
                     trigger["trigger"] = appExec.Trigger;
                     trigger["vmstate"] = appExec.VMState;
                     trigger["gasconsumed"] = appExec.GasConsumed.ToString();
-                    try
+                    var stack = new JArray();
+                    foreach (var item in appExec.Stack)
                     {
-                        trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
-                    }
-                    catch (InvalidOperationException)
-                    {
-                        trigger["stack"] = "error: invalid operation";
+                        try
+                        {
+                            stack.Add(item.ToJson());
+                        }
+                        catch (InvalidOperationException)
+                        {
+                            stack.Add("error: invalid operation");
+                        }
                     }
                     trigger["notifications"] = appExec.Notifications.Select(q =>
                     {
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -94,9 +94,9 @@ namespace Neo.Plugins
             {
                 trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
             }
-            catch (Exception ex)
+            catch (InvalidOperationException)
             {
-                trigger["stackexception"] = ex.Message;
+                trigger["stack"] = "error: invalid operation";
             }
             trigger["notifications"] = appExec.Notifications.Select(q =>
             {
@@ -137,9 +137,9 @@ namespace Neo.Plugins
                     {
                         trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
                     }
-                    catch (Exception ex)
+                    catch (InvalidOperationException)
                     {
-                        trigger["stackexception"] = ex.Message;
+                        trigger["stack"] = "error: invalid operation";
                     }
                     trigger["notifications"] = appExec.Notifications.Select(q =>
                     {
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -97,9 +97,9 @@ namespace Neo.Plugins
                 {
                     stack.Add(item.ToJson(Settings.Default.MaxStackSize));
                 }
-                catch (InvalidOperationException)
+                catch (Exception ex)
                 {
-                    stack.Add("error: invalid operation");
+                    stack.Add("error: " + ex);
                 }
             }
             trigger["stack"] = stack;
@@ -145,11 +145,12 @@ namespace Neo.Plugins
                         {
                             stack.Add(item.ToJson(Settings.Default.MaxStackSize));
                         }
-                        catch (InvalidOperationException)
+                        catch (Exception ex)
                         {
-                            stack.Add("error: invalid operation");
+                            stack.Add("error: " + ex);
                         }
                     }
+                    trigger["stack"] = stack;
                     trigger["notifications"] = appExec.Notifications.Select(q =>
                     {
                         JObject notification = new JObject();
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -96,7 +96,7 @@ namespace Neo.Plugins
             }
             catch (Exception ex)
             {
-                trigger["exception"] = ex.Message;
+                trigger["stackexception"] = ex.Message;
             }
             trigger["notifications"] = appExec.Notifications.Select(q =>
             {
@@ -139,7 +139,7 @@ namespace Neo.Plugins
                     }
                     catch (Exception ex)
                     {
-                        trigger["exception"] = ex.Message;
+                        trigger["stackexception"] = ex.Message;
                     }
                     trigger["notifications"] = appExec.Notifications.Select(q =>
                     {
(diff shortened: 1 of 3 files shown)


src/ApplicationLogs/LogReader.cs
```diff
@@ -96,7 +96,7 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
             }
             catch (Exception ex)
             {
-                trigger["exception"] = ex.Message;
+                trigger["stackexception"] = ex.Message;
```
roman-khimov, line 99: Current behavior was introduced in 81eb9b78393a080075f53afbe7ae7d915693e712, pretty much accidentally (hi, monorepo). See #854 for NeoGo behavior wrt this (which is closer to the older one). I don't like how we randomly change these things, it's a public API after all.

roman-khimov: We need to ensure application logs and test invocations are treated the same wrt this.

Liaojinghui:
a new field?

src/ApplicationLogs/LogReader.cs
superboyiii: Seems from https://github.com/neo-project/neo-modules/pull/696, how about revert to the older one?

superboyiii:
> a new field?

Two kind of exception(one from vn, one from applicationlog), but they're using the same trigger. applicationlog exception could rewrite vm exception...

src/ApplicationLogs/LogReader.cs
Liaojinghui: > Seems from #696, how about revert to the older one as `stack`?

i vote this.

superboyiii:
Now it's the same result as invocation. @roman-khimov

Liaojinghui (approved):

shargon:
If we add the stack, i prefer to add the stack item type, and the value, otherwise I prefer to use the exception field

superboyiii:
Tested. @shargon @roman-khimov 
image

src/ApplicationLogs/LogReader.cs
```diff
@@ -90,14 +90,19 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
...
+            foreach (var item in appExec.Stack)
             {
-                trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
-            }
-            catch (Exception ex)
-            {
-                trigger["exception"] = ex.Message;
+                try
+                {
+                    stack.Add(item.ToJson());
```
roman-khimov, line 98: The old one had `Settings.Default.MaxStackSize` parameter, but I'm not sure it affects anything here.

roman-khimov (commented):
How about test invocations (`GetInvokeResult`)?

OK otherwise.

shargon:
> > Why we don't use the exception field?
> 
> These two exceptions are generated from different place, but applicationlog exception could re-write vm exception which doesn't make sense and can make some Dapp backend confusing a lot for handling these exceptions. For example: Ghostmarket.

What about change the exception field to be an array, also with type?

roman-khimov:
If it'd be separated, you wouldn't know which elements are bad. In some cases it might be useful. Imagine 3 items on the stack, the second one is bad, but the other two are OK and can be serialized, if you get two stack items and one exception, what was the original index of the element that has caused this exception?

cschuchardt88:
#807 And #842 implements `AppclaitionEngine.Log` for the meantime in application log for debug mode. Until core can be update to do it properly.

superboyiii:
Done. @roman-khimov @shargon

src/ApplicationLogs/LogReader.cs
```diff
@@ -90,14 +90,19 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
...
+            foreach (var item in appExec.Stack)
             {
-                trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
-            }
-            catch (Exception ex)
-            {
-                trigger["exception"] = ex.Message;
+                try
+                {
+                    stack.Add(item.ToJson(Settings.Default.MaxStackSize));
```
AnnaShaleva, line 98: Just a side note: it's not yet implemented in NeoGo, but how about restricting the whole list of serializable stackitems by `Settings.Default.MaxStackSize` instead of restricting every single stackitem by separate `Settings.Default.MaxStackSize`?

src/ApplicationLogs/LogReader.cs
```diff
@@ -90,14 +90,19 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
...
+                }
+                catch (InvalidOperationException)
+                {
+                    stack.Add("error: invalid operation");
```
AnnaShaleva, line 102: Could we provide an exception message for more error details?

src/ApplicationLogs/LogReader.cs
```diff
@@ -90,14 +90,19 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
...
-                trigger["stack"] = appExec.Stack.Select(q => q.ToJson(Settings.Default.MaxStackSize)).ToArray();
-            }
-            catch (Exception ex)
-            {
-                trigger["exception"] = ex.Message;
+                try
+                {
+                    stack.Add(item.ToJson(Settings.Default.MaxStackSize));
+                }
+                catch (InvalidOperationException)
```
AnnaShaleva, line 100: Maybe catch every exception, not only `InvalidOperationException`?

superboyiii:
@AnnaShaleva Could you review again? Any more improvement needed?

AnnaShaleva (approved):

src/ApplicationLogs/LogReader.cs
AnnaShaleva: Exception message, it's OK now.

src/ApplicationLogs/LogReader.cs
AnnaShaleva: > Because each StackItem can be an ExecutionEngineLimits

That's the case, we can't allow this during results serilisation, because it's not allowed for regular executions. Although it's not a problem if we use the right limits for execution: we just can't get too large number of stackitems in this case.

So resolved.

superboyiii:
@Liaojinghui @shargon Is it OK for you?

src/ApplicationLogs/LogReader.cs
```diff
@@ -133,14 +138,19 @@ public static JObject BlockLogToJson(Block block, IReadOnlyList<Blockchain.Appli
...
-                    catch (Exception ex)
-                    {
-                        trigger["exception"] = ex.Message;
+                        try
+                        {
+                            stack.Add(item.ToJson(Settings.Default.MaxStackSize));
+                        }
+                        catch (Exception ex)
+                        {
+                            stack.Add("error: " + ex);
```
cschuchardt88, line 150: I dont think will output will help much. Seeing how the only reason it would fail is because its an interop and cant be serialized. Better to put an empty or null stackitem instead.

src/ApplicationLogs/LogReader.cs
```diff
@@ -90,14 +90,19 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
...
-            catch (Exception ex)
-            {
-                trigger["exception"] = ex.Message;
+                try
+                {
+                    stack.Add(item.ToJson(Settings.Default.MaxStackSize));
+                }
+                catch (Exception ex)
+                {
+                    stack.Add("error: " + ex);
```
cschuchardt88, line 102: I dont think will output will help much. Seeing how the only reason it would fail is because its an interop and cant be serialized. Better to put an empty or null stackitem instead.

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -93,14 +93,19 @@ private JObject GetInvokeResult(byte[] script, Signer[] signers = null, Witness[
...
-                catch (InvalidOperationException)
-                {
-                    json["stack"] = "error: invalid operation";
+                    try
+                    {
+                        stack.Add(ToJson(item, session));
+                    }
+                    catch (Exception ex)
+                    {
+                        stack.Add("error: " + ex);
```
cschuchardt88, line 105: same

src/ApplicationLogs/LogReader.cs
shargon, line 102:
```suggestion
                    stack.Add("error: " + ex.Message);
```

src/ApplicationLogs/LogReader.cs
shargon, line 150:
```suggestion
                            stack.Add("error: " + ex.Message);
```

src/RpcServer/RpcServer.SmartContract.cs
shargon, line 105:
```suggestion
                        stack.Add("error: " + ex.Message);
```

superboyiii:
Done @shargon

src/ApplicationLogs/LogReader.cs
```diff
@@ -147,7 +147,7 @@ public static JObject BlockLogToJson(Block block, IReadOnlyList<Blockchain.Appli
                         }
                         catch (Exception ex)
                         {
-                            stack.Add("error: " + ex);
+                            stack.Add("error: " + ex.Message);
```
cschuchardt88, line 150:
```suggestion
                            stack.Add(item.GetInterface<object>().GetType().Name);
```

src/RpcServer/RpcServer.SmartContract.cs
```diff
@@ -102,7 +102,7 @@ private JObject GetInvokeResult(byte[] script, Signer[] signers = null, Witness[
                     }
                     catch (Exception ex)
                     {
-                        stack.Add("error: " + ex);
+                        stack.Add("error: " + ex.Message);
```
cschuchardt88, line 105:
```suggestion
                        stack.Add(item.GetInterface<object>().GetType().Name);
```

src/ApplicationLogs/LogReader.cs
```diff
@@ -99,7 +99,7 @@ public static JObject TxLogToJson(Blockchain.ApplicationExecuted appExec)
                 }
                 catch (Exception ex)
                 {
-                    stack.Add("error: " + ex);
+                    stack.Add("error: " + ex.Message);
```
cschuchardt88, line 102:
```suggestion
                    stack.Add(item.GetInterface<object>().GetType().Name);
```

cschuchardt88 (changes requested):

shargon (dismissed):

shargon:
Maybe add the error description will not be deterministic between neo-go and c#?

cschuchardt88:
I think we should be doing this the proper way, indicating that there is an validate `stackitem`. We shouldnt be hiding it or saying there is something wrong with it; when there is not.

src/ApplicationLogs/LogReader.cs
superboyiii: This can break applicationlog getting the stack:
3e8148970abed45ba6ef0e98c4161c3
But you can see it's `on-chain`:
7723fcd8ecae99fdf1a629501952ada

src/RpcServer/RpcServer.SmartContract.cs
superboyiii: In this way, it will throw exception directly but can't get result stack!
725922486f0147421b02d13395e875b
Compare to current code:
302dc02b1c0fea24c7e69980f02256f

shargon:
> > error description will not be deterministic between neo-go and c#?
> 
> Yes, it will differ.

Then is better to change the message

roman-khimov:
I think we can settle on the fact that if you're receiving a string instead of a stack item, it's an error string and the original stack item couldn't be represented. And this string can be implementation-specific, just like `data` in JSON-RPC errors.

superboyiii:
@shargon @roman-khimov need approve

shargon (approved):

PR #856, opened 2023-12-06
Update workflow

Ashuaidehao:


shargon (approved):

shargon:
The workflow fails https://github.com/neo-project/neo-modules/actions/runs/7110967931

Issue #826, opened 2023-09-16
Plugin Coding Standard


cschuchardt88:
**Summary or problem description**
For example RpcClient method `GetContractStateAsync` doesn't even use its model class at all for serializing and deserializing .
![image](https://github.com/neo-project/neo-modules/assets/8141309/9dc524ea-31b3-4c00-a3b7-b2e6b9852b3f)

RpcClient
- `RpcStateRoot` ToJson method which means no test
- `RpcMethodToken` ToJson method which means no test
- `RpcFoundStates` ToJson method which means no test
- `GetContractStateAsync` 
- Possibly More

**Do you have any solution you want to propose?**
We have some kind of standard for writing plugins; code in general. For example, Folder structure layout. Maybe Interfaces for certain things like getting storage from other plugins. Making and writing unit tests. A name in convention for properties names in config.json. Just to name a couple. If we don't as you can see things can get lost in time, as new requirements come along and deadlines need to be met. Also we can update  `.editorconfig` file for most of these little things.

**Where in the software does this update applies to?**
- Plugins

PR #858, opened 2023-12-13
Fix: Code style

Jim8y:
Since discussed that the code style is a thing for the core, this pr fixs some of the code style issues that has being there for years.

### DIFF
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -56,16 +56,16 @@ namespace Neo.Plugins
         }
 
         [RpcMethod]
-        public JToken GetApplicationLog(JArray @params)
+        public JToken GetApplicationLog(JArray parameters)
         {
-            UInt256 hash = UInt256.Parse(@params[0].AsString());
+            UInt256 hash = UInt256.Parse(parameters[0].AsString());
             byte[] value = _db.TryGet(hash.ToArray());
             if (value is null)
                 throw new RpcException(-100, "Unknown transaction/blockhash");
 
             JObject raw = (JObject)JToken.Parse(Neo.Utility.StrictUTF8.GetString(value));
             //Additional optional "trigger" parameter to getapplicationlog for clients to be able to get just one execution result for a block.
-            if (@params.Count >= 2 && Enum.TryParse(@params[1].AsString(), true, out TriggerType trigger))
+            if (parameters.Count >= 2 && Enum.TryParse(parameters[1].AsString(), true, out TriggerType trigger))
             {
                 var executions = raw["executions"] as JArray;
                 for (int i = 0; i < executions.Count;)
--- a/src/ApplicationLogs/LogReader.cs
+++ b/src/ApplicationLogs/LogReader.cs
@@ -56,16 +56,16 @@ namespace Neo.Plugins
         }
 
         [RpcMethod]
-        public JToken GetApplicationLog(JArray _params)
+        public JToken GetApplicationLog(JArray @params)
         {
-            UInt256 hash = UInt256.Parse(_params[0].AsString());
+            UInt256 hash = UInt256.Parse(@params[0].AsString());
             byte[] value = _db.TryGet(hash.ToArray());
             if (value is null)
                 throw new RpcException(-100, "Unknown transaction/blockhash");
 
             JObject raw = (JObject)JToken.Parse(Neo.Utility.StrictUTF8.GetString(value));
             //Additional optional "trigger" parameter to getapplicationlog for clients to be able to get just one execution result for a block.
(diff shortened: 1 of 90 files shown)


src/LevelDBStore/IO/Data/LevelDB/Native.cs
```diff
@@ -16,8 +16,8 @@ namespace Neo.IO.Data.LevelDB
 {
     public enum CompressionType : byte
     {
-        kNoCompression = 0x0,
-        kSnappyCompression = 0x1
+        KNoCompression = 0x0,
+        KSnappyCompression = 0x1
```
shargon, line 20: We can remove the K

src/ApplicationLogs/LogReader.cs
```diff
@@ -56,16 +56,16 @@ protected override void OnSystemLoaded(NeoSystem system)
         }
 
         [RpcMethod]
-        public JToken GetApplicationLog(JArray _params)
+        public JToken GetApplicationLog(JArray @params)
```
shargon, line 59: parameters, value?

src/StateService/StatePlugin.cs
```diff
@@ -176,36 +176,36 @@ private void OnVerifyProof(UInt256 root_hash, string proof)
         }
 
         [RpcMethod]
-        public JToken GetStateRoot(JArray _params)
+        public JToken GetStateRoot(JArray @params)
```
shargon, line 179: same here and in the rest of them, I don't like the @

Liaojinghui: Using @ for reserved words to use them as a variable is a C# standard. But I'll update it. And please, we are working on the core, not some personal project, please avoid using "like" or "not like", it sounds like you are using your personal feelings instead of professional judgements (though i know you are not).

src/DBFTPlugin/DBFTPlugin.cs
```diff
@@ -34,7 +34,7 @@ public DBFTPlugin()
 
         public DBFTPlugin(Settings settings) : this()
         {
-            this.settings = settings;
+            this._settings = settings;
```
cschuchardt88, line 37: You should remove `this`. Dotnet coding standards, 99% of professional c# developers dont use that method anymore.

src/DBFTPlugin/DBFTPlugin.cs
```diff
@@ -44,55 +44,55 @@ public override void Dispose()
 
         protected override void Configure()
         {
-            settings ??= new Settings(GetConfiguration());
+            _settings ??= new Settings(GetConfiguration());
```
cschuchardt88, line 47: should remove the class. so it would be
```csharp
_settings ??= new(GetConfiguration());
```

cschuchardt88 (commented):
You know you can use the analyzers in Visual Studio to do this all for you automatically. Also we should be using the `.editorconfig` so this doesn't happen again. See #826

src/StateService/StatePlugin.cs
shargon: > Using @ for reserved words to use them as a variable is a C# standard. But I'll update it. And please, we are working on the core, not some personal project, please avoid using "like" or "not like", it sounds like you are using your personal feelings (though i know you are not) instead of professional judgements.

Could you show me where said that use a reserved word is a good practice?

Liaojinghui: I am not saying using reserved word is a good practice, cause if i were the one who wrote the code, i will not use `params` as the parameter. But it is params and is used as _params, which is definately not a good practice. You said we should have a good code style, yet you allow someone to use `_params` as the name of the parameter.

Liaojinghui:
> You know you can use the analyzers in Visual Studio to do this all for you automatically. Also we should be using the `.editorconfig` so this doesn't happen again. See #826

Only if someone does it. Anything in neo core can take years.

src/DBFTPlugin/DBFTPlugin.cs
Liaojinghui: This does not matter. Many places keep it, many remove it.

src/StateService/StatePlugin.cs
roman-khimov: @Liaojinghui, you're trying to change some established code. `_params` was there since like forever, so for good or bad, but everyone just got used to it being this way. If all methods consistently use this name, no one can really object adding another one using the same style, that's just consistency. Is it a bad name? Sure it is. But just making it `@params` is not an improvement to me, at least not worth the trouble.

`parameters` is a different story, nice to have it instead and have it used consistently for every method as well. :+1:

Jim8y:
close as not planed, may work with editorconfig

PR #859, opened 2024-01-03
Fix response from getversion due to WS removal

vncoelho:


cschuchardt88:
Can you post reminders on `RestServer` and `WebSocketServer` PRs?

Jim8y (dismissed):

Jim8y:
We still have other places that contain wsport, mind to remove them as well?

shargon:
> We still have other places that contain wsport, mind to remove them as well?

If there are more we should remove them together

vncoelho:
> We still have other places that contain wsport, mind to remove them as well?

I am not sure

vncoelho:
Maybe rpcclient, lets double check

cschuchardt88:
Try compiling the whole solution, and check for errors. After you update all the packages.

vncoelho:
This was not compilation error,@cschuchardt88

cschuchardt88:
But if it errors, it will let you know if anymore exist.

vncoelho:
I was still setting up our tools that creates neo-cli client and add plugins.

I will soon compile again, but since I was able to add the plugin I think that the library does not provide compilation error, just when called.
I also added RPCClient to it.

However, thinking about the logic there must be a similar call there. Later I will check.

Jim8y:
Just do a wsport search you will find them.

cschuchardt88:
You want to be searching for `ws` and `websocket`

superboyiii:
62bb544de4aeedba1e708995542dd5a
All these should be removed.

shargon (approved):

Issue #860, opened 2024-01-03
Change nuget for ConsoleService for a modified local client 


vncoelho:
Directory.Build.props was moved to root.
From there we can modify it and change the reference for Neo.

However, some plugins have their on reference such as:
`<PackageReference Include="Neo.ConsoleService" Version="1.2.0" ExcludeAssets="runtime" />`

What is the best way to change those to your local development?

cschuchardt88:
### Add change `nuget.json` file to:
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <clear />
    <add key="NuGet.org" value="https://api.nuget.org/v3/index.json" />
    <add key="local" value="./deps/" />
    <add key="neo-myget" value="https://www.myget.org/F/neo/api/v3/index.json" />
  </packageSources>
</configuration>
```

### Create directory `deps` in root of the repo.

### Put your nuget packages in `deps` directory that you built.

### Use nuget to update the packages you want to change; to `local`.

![image](https://github.com/neo-project/neo-modules/assets/8141309/5ee62e63-b3fd-42ad-84c9-ca7614bd3945)

### Update the packages from `local` nuget source like so:
![image](https://github.com/neo-project/neo-modules/assets/8141309/1449ca50-3ea8-4c80-ab30-757dc8d5e66d)

**NOTE**: Install or update, or reinstall

PR #861, opened 2024-01-03
Local dev setup

cschuchardt88:
Setup for local environment

See https://github.com/neo-project/neo-modules/issues/860#issuecomment-1876124370


NuGet.Config
```diff
@@ -1,8 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
```
Jim8y, line 2: why your pr changes so many lines? different encoding?

cschuchardt88: using the `LF` encoding. that is set in `.editorconfig`

Jim8y (approved):

shargon (approved):

Issue #887, opened 2024-03-18
AI Code Reviewer problems when checking PRs


vncoelho:
Recently reported on #680  and currently occurring on master.

```js
Run freeedcom/ai-codereviewer@a9a064dfa1db8c83f40ef63f6e247fa09c935ed6
Error: RequestError [HttpError]: Not Found
    at /home/runner/work/_actions/freeedcom/ai-codereviewer/a9a064dfa1db8c83f40ef63f6e247fa09c935ed6/webpack:/open-ai-reviewer/node_modules/@octokit/request/dist-node/index.js:78:1
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  status: 404,
  response: {
    url: 'https://api.github.com/repos/neo-project/neo-modules/pulls/',
    status: 404,
    headers: {
      'access-control-allow-origin': '*',
      'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
      connection: 'close',
      'content-encoding': 'gzip',
      'content-security-policy': "default-src 'none'",
      'content-type': 'application/json; charset=utf-8',
      date: 'Mon, 18 Mar 2024 15:25:44 GMT',
      'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
      server: 'GitHub.com',
      'strict-transport-security': 'max-age=31536000; includeSubdomains; preload',
      'transfer-encoding': 'chunked',
      vary: 'Accept-Encoding, Accept, X-Requested-With',
      'x-content-type-options': 'nosniff',
      'x-frame-options': 'deny',
      'x-github-api-version-selected': '2022-11-28',
      'x-github-media-type': 'github.v3; format=json',
      'x-github-request-id': '5059:0E48:A1B8E:130D85:65F85CF8',
      'x-ratelimit-limit': '5000',
      'x-ratelimit-remaining': '4997',
      'x-ratelimit-reset': '1710779043',
      'x-ratelimit-resource': 'core',
      'x-ratelimit-used': '3',
      'x-xss-protection': '0'
    },
    data: {
[...]

PR #888, opened 2024-03-18
Remove IA

shargon:
Close https://github.com/neo-project/neo-modules/issues/887


vncoelho:
We can try this again soon I think
The problem looks like to be the same as here

https://github.com/freeedcom/ai-codereviewer/issues/34

vncoelho (approved):

Jim8y (approved):

PR #898, opened 2024-04-11
Fix storage snapshot sync

cschuchardt88:
_Change Log_
---
- Fixed sync order

_**Note**_:  DO NOT need to resync

Closes #897

PR #902, opened 2024-05-07
Avoid re-init properties

shargon:
Related to https://github.com/neo-project/neo-modules/pull/898/files#r1584389953


cschuchardt88 (commented):
You need these or else you have to `initialize` each one, because they will be `null` or no `instance` then you do `new()`; `ISerializable.Serialize` or `ISerializable.Deserialize` will get `exception` `object not set to an instance of an object`. Also on some of them `Size` won't work because of that reason.

shargon:
> you have to `initialize` each one

Are always initialized, isn't it?

cschuchardt88:
No, you would have to do something like this:

```csharp
var obj = new AClass()
{
  AProperty = [],
}
```

src/ApplicationLogs/Store/States/TransactionLogState.cs
```diff
@@ -33,16 +33,16 @@ public class TransactionLogState : ISerializable, IEquatable<TransactionLogState
         public virtual void Deserialize(ref MemoryReader reader)
         {
             // It should be safe because it filled from a transaction's notifications.
-            uint aLen = reader.ReadUInt32();
+            var aLen = reader.ReadUInt32();
```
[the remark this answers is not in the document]
cschuchardt88: No, it's just using dynamic type variable now; instead of being bound to a type specific variable.

shargon:
> No, you would have to do something like this:
> 
> ```cs
> var obj = new AClass()
> {
>   AProperty = new(),
> }
> ```

It's currently done, otherwise you will have a default unused value, that will be changed in `Create` method

cschuchardt88:
Yes, But if someone was do `AClass.AsSerializable()` it would `throw` that `exception`. This library was importable for other plugins to access the store.

shargon:
> Yes, But if someone was do `bytes.AsSerializable<AClass>()` it would `throw` that `exception`. This library was importable for other plugins to access the store.
> 
> Lets have @superboyiii test 1st. before merge.

Good point, but it call `Deserialize`

src/ApplicationLogs/Store/States/NotifyLogState.cs
```diff
@@ -17,11 +17,11 @@ namespace ApplicationLogs.Store.States
 {
     public class NotifyLogState : ISerializable, IEquatable<NotifyLogState>
     {
-        public UInt160 ScriptHash { get; protected set; } = new();
-        public string EventName { get; protected set; } = string.Empty;
-        public Guid[] StackItemIds { get; protected set; } = [];
+        public UInt160 ScriptHash { get; protected set; }
+        public string EventName { get; protected set; }
+        public Guid[] StackItemIds { get; protected set; }
```
cschuchardt88, line 22: We should make this `required` as well.

src/ApplicationLogs/Store/States/TransactionEngineLogState.cs
```diff
@@ -15,9 +15,9 @@ namespace Neo.Plugins.Store.States
 {
     public class TransactionEngineLogState : ISerializable, IEquatable<TransactionEngineLogState>
     {
-        public Guid[] LogIds { get; private set; } = Array.Empty<Guid>();
+        public Guid[] LogIds { get; private set; }
```
cschuchardt88, line 18: We should make this `required` as well.

src/ApplicationLogs/Store/States/TransactionLogState.cs
```diff
@@ -16,9 +16,9 @@ namespace ApplicationLogs.Store.States
 {
     public class TransactionLogState : ISerializable, IEquatable<TransactionLogState>
     {
-        public Guid[] NotifyLogIds { get; private set; } = Array.Empty<Guid>();
+        public Guid[] NotifyLogIds { get; private set; }
```
cschuchardt88, line 19: We should make this `required` as well.

src/ApplicationLogs/Store/States/ExecutionLogState.cs
```diff
@@ -17,12 +17,12 @@ namespace ApplicationLogs.Store.States
...
     public class ExecutionLogState : ISerializable, IEquatable<ExecutionLogState>
     {
-        public VMState VmState { get; private set; } = VMState.NONE;
-        public string Exception { get; private set; } = string.Empty;
-        public long GasConsumed { get; private set; } = 0L;
-        public Guid[] StackItemIds { get; private set; } = [];
+        public VMState VmState { get; private set; }
+        public string Exception { get; private set; }
+        public long GasConsumed { get; private set; }
+        public Guid[] StackItemIds { get; private set; }
```
cschuchardt88, line 23: We should make this `required` as well.

src/ApplicationLogs/Store/States/EngineLogState.cs
```diff
@@ -15,8 +15,8 @@ namespace Neo.Plugins.Store.States
 {
     public class EngineLogState : ISerializable, IEquatable<EngineLogState>
     {
-        public UInt160 ScriptHash { get; private set; } = new();
-        public string Message { get; private set; } = string.Empty;
+        public UInt160 ScriptHash { get; private set; }
+        public string Message { get; private set; }
```
cschuchardt88, line 19: We should make this `required` as well.

src/ApplicationLogs/Store/States/ContractLogState.cs
```diff
@@ -18,10 +18,10 @@ namespace ApplicationLogs.Store.States
 {
     public class ContractLogState : NotifyLogState, IEquatable<ContractLogState>
     {
-        public UInt256 TransactionHash { get; private set; } = new();
-        public TriggerType Trigger { get; private set; } = TriggerType.All;
+        public UInt256 TransactionHash { get; private set; }
+        public TriggerType Trigger { get; private set; }
```
cschuchardt88, line 22: We should make this `required` as well.

src/ApplicationLogs/Store/States/BlockLogState.cs
```diff
@@ -16,9 +16,9 @@ namespace ApplicationLogs.Store.States
 {
     public class BlockLogState : ISerializable, IEquatable<BlockLogState>
     {
-        public Guid[] NotifyLogIds { get; private set; } = [];
+        public Guid[] NotifyLogIds { get; private set; }
```
cschuchardt88, line 19: We should make this `required` as well.

cschuchardt88 (changes requested):

src/ApplicationLogs/Store/States/TransactionEngineLogState.cs
shargon: Then set can't be private

src/ApplicationLogs/Store/States/BlockLogState.cs
Jim8y: like this? Kinda ugly:

```C#
public virtual void Deserialize(ref MemoryReader reader)
{
    // It should be safe because it filled from a block's notifications.
    var aLen = reader.ReadUInt32();
    var notifyLogIds = new Guid[aLen];
    for (var i = 0; i < aLen; i++)
        notifyLogIds[i] = new Guid(reader.ReadVarMemory().Span);

    // Create a new instance of BlockLogState using the Create factory method
    var blockLogState = Create(notifyLogIds);

    // Assign the created instance to the current object
    NotifyLogIds = blockLogState.NotifyLogIds;
}

```

shargon:
Repository moved to [neo](https://github.com/neo-project/neo), please re-open there

PR #906, opened 2024-05-11
fix CalculateNetworkFee

Jim8y:
This pr is to fix the null object error in CalculateNetworkFee


superboyiii (approved):

cschuchardt88 (commented):
why you have to discard it? why not `default!`

Jim8y:
> why you have to discard it? why not `default!`

What?

shargon (commented):
LGTM

